PR #10, opened 2016-04-19
Second design prototype

loosebazooka:


loosebazooka:
oops, this is goign the wrong way.

PR #13, opened 2016-04-19
Refactoring code

meltsufin:
Replaces the Actions abstraction with service interfaces and removes the AppExecutor abstraction.


meltsufin:
@patflynn, @loosebazooka PTAL

src/main/java/com/google/cloud/tools/app/impl/config/DefaultStageFlexibleConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+package com.google.cloud.tools.app.impl.config;
+
+import com.google.cloud.tools.app.api.deploy.StageFlexibleConfiguration;
+
+import java.io.File;
+
+/**
+ * Plain Java bean implementation of {@link StageFlexibleConfiguration}
+ */
+public class DefaultStageFlexibleConfiguration implements StageFlexibleConfiguration {
```
loosebazooka: why not just use the builders we had before?

meltsufin: Just that this is simpler to generate. What's the advantage of the builders? To make these immutable?

src/main/java/com/google/cloud/tools/app/api/genconfig/GenConfigParams.java
```diff
@@ -0,0 +1,30 @@
+/**
```
loosebazooka: are we getting rid of genconfig?

patflynn:
I'm never going to be able to review this in a timely manner. Merge it and I'll do a post merge review.

PR #14, opened 2016-04-22
Fix deploy command and improve build speed

meltsufin:
During refactoring I broke the "depoy" command by changing it to "cloud" accidentally. This fixes it.
Also, there is a performance issue with maven-jar-plugin when used with Java 7 that this fixes.

@loosebazooka  PTAL

### DIFF
--- a/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/sdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/sdk/CloudSdk.java
@@ -55,6 +55,8 @@ public class CloudSdk {
     command.add("app");
     command.addAll(args);
 
+    sysOutCommand(command);
+
     return processRunner.run(command.toArray(new String[command.size()]));
   }
 
@@ -64,7 +66,18 @@ public class CloudSdk {
     command.add(getDevAppServerPath().toString());
     command.addAll(args);
 
+    sysOutCommand(command);
+
     return processRunner.run(command.toArray(new String[command.size()]));
+  }
+
+  private void sysOutCommand(List<String> command) {
+    StringBuilder commandString = new StringBuilder();
+    for (String argument : command) {
+      commandString.append(" ");
+      commandString.append(argument);
+    }
+    System.out.println("submitting command:" + commandString);
   }
 
   private Path getSdkPath() {
--- a/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/sdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/sdk/CloudSdk.java
@@ -22,11 +22,14 @@ import com.google.common.base.Joiner;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.logging.Logger;
 
 /**
  * Cloud SDK CLI wrapper.
  */
 public class CloudSdk {
+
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/sdk/CloudSdk.java
```diff
@@ -64,9 +66,20 @@ public int runDevAppServerCommand(List<String> args) throws ProcessRunnerExcepti
     command.add(getDevAppServerPath().toString());
     command.addAll(args);
 
+    sysOutCommand(command);
+
     return processRunner.run(command.toArray(new String[command.size()]));
   }
 
+  private void sysOutCommand(List<String> command) {
+    StringBuilder commandString = new StringBuilder();
```
loosebazooka: Possible to use Guava Joiner here? http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Joiner.html

loosebazooka: Also not sure I'd want this printed out by default. Would be related to log level wouldn't it?

loosebazooka: test should have caught this?

src/test/java/com/google/cloud/tools/app/CloudSdkAppEngineDeploymentTest.java
```diff
@@ -78,7 +78,7 @@ public void testNewDeployAction_allFlags() throws AppEngineException, ProcessRun
     deployment.deploy(configuration);
 
     List<String> expectedCommand = ImmutableList
-        .of("cloud", appYaml1.toString(), "--bucket", "gs://a-bucket", "--docker-build",
+        .of("deploy", appYaml1.toString(), "--bucket", "gs://a-bucket", "--docker-build",
```
loosebazooka: oh I see why tests didn't catch this. lulz

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/sdk/CloudSdk.java
meltsufin: Joiner is a good idea. I was looking for something like that.

meltsufin: We're not passing a logger in at the moment. Maybe we should consider something like that.

loosebazooka:
lgtm, (explore logging as future work?)

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/sdk/CloudSdk.java
```diff
@@ -64,9 +70,16 @@ public int runDevAppServerCommand(List<String> args) throws ProcessRunnerExcepti
     command.add(getDevAppServerPath().toString());
     command.addAll(args);
 
+    sysOutCommand(command);
+
     return processRunner.run(command.toArray(new String[command.size()]));
   }
 
+  private void sysOutCommand(List<String> command) {
```
loosebazooka: This is just nitpicky, but this functions utility doesn't seem as obvious as before, feel like inline is easier to read?

loosebazooka:
lgtm

PR #15, opened 2016-04-22
Fix App Engine Sdk system property

loosebazooka:


loosebazooka:
@meltsufin

meltsufin:
Oops, that was another error in refactoring.
:ship:

PR #19, opened 2016-04-26
Use File instead of Path everywhere

### LINKED ISSUE #18: Use File instead of Path everywhere
Maven and Gradle do not support automatic parameter binding to Path.

meltsufin:
Fixes #18.

### DIFF
--- a/src/main/java/com/google/cloud/tools/app/impl/appcfg/AppCfgAppEngineStandardStaging.java
+++ b/src/main/java/com/google/cloud/tools/app/impl/appcfg/AppCfgAppEngineStandardStaging.java
@@ -90,7 +90,7 @@ public class AppCfgAppEngineStandardStaging implements AppEngineStandardStaging 
       appEngineSdk.runCommand(arguments);
 
       if (dockerfile != null && dockerfileDestination != null) {
-        Files.copy(dockerfile, dockerfileDestination);
+        Files.copy(dockerfile, new File(dockerfileDestination, dockerfile.getName()));
       }
 
     } catch (IOException e) {
--- a/src/main/java/com/google/cloud/tools/app/impl/appcfg/AppCfgAppEngineStandardStaging.java
+++ b/src/main/java/com/google/cloud/tools/app/impl/appcfg/AppCfgAppEngineStandardStaging.java
@@ -19,10 +19,11 @@ import com.google.cloud.tools.app.api.AppEngineException;
 import com.google.cloud.tools.app.api.deploy.AppEngineStandardStaging;
 import com.google.cloud.tools.app.api.deploy.StageStandardConfiguration;
 import com.google.common.base.Preconditions;
-import com.google.common.io.Files;
 
-import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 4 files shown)


loosebazooka:
Yeah I guess if we're changing the configurations to be File, might as well be consistent, would have ideally liked if gradle handled this correctly :\

src/main/java/com/google/cloud/tools/app/impl/appcfg/AppCfgAppEngineStandardStaging.java
```diff
@@ -82,16 +82,17 @@ public void stageStandard(StageStandardConfiguration configuration) throws AppEn
...
+        configuration.getDockerfile() == null ? null : configuration.getDockerfile();
+    File dockerfileDestination = configuration.getStagingDirectory();
 
     try {
 
       appEngineSdk.runCommand(arguments);
 
       if (dockerfile != null && dockerfileDestination != null) {
-        Files.copy(dockerfile, dockerfileDestination, StandardCopyOption.REPLACE_EXISTING);
+        Files.copy(dockerfile.toPath(), dockerfileDestination.toPath(),
```
loosebazooka: should we use the copy that works on Files here?

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/sdk/CloudSdk.java
```diff
@@ -97,24 +97,24 @@ private void outputCommand(List<String> command) {
...
   }
 
-  private Path getSdkPath() {
+  private File getSdkPath() {
     return sdkPath;
   }
 
-  private Path getGCloudPath() {
-    return sdkPath.resolve(GCLOUD);
+  private File getGCloudPath() {
```
loosebazooka: Would it make sense to have a pure file implementation here for these methods

src/main/java/com/google/cloud/tools/app/impl/appcfg/AppCfgAppEngineStandardStaging.java
meltsufin: It would be nice, but I'm not aware of a simple API in the JDK for copying files without using Paths.

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/sdk/CloudSdk.java
meltsufin: Yes, but how? Also, I want to address the questions of whether we need all of these get...Path methods in a separate PR.

src/main/java/com/google/cloud/tools/app/impl/appcfg/AppCfgAppEngineStandardStaging.java
meltsufin: I guess we can use Apache FileUtils. What do you think?

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/sdk/CloudSdk.java
```diff
@@ -97,24 +98,24 @@ private void outputCommand(List<String> command) {
...
+    return new File(sdkPath, GCLOUD);
   }
 
-  private Path getDevAppServerPath() {
-    return sdkPath.resolve(DEV_APPSERVER_PY);
+  private File getDevAppServerPath() {
```
loosebazooka: Still using path here?

meltsufin: You mean in the method name?

meltsufin: Oh I see!

loosebazooka:
lgtm

PR #21, opened 2016-04-29
Correct stage task parameter order and update stage and deploy configurations

loosebazooka:
change config param applicationID to application
change -A to --application

### DIFF
--- a/src/main/java/com/google/cloud/tools/app/impl/config/DefaultStageStandardConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/app/impl/config/DefaultStageStandardConfiguration.java
@@ -29,7 +29,7 @@ public class DefaultStageStandardConfiguration implements StageStandardConfigura
   private boolean enableQuickstart;
   private boolean disableUpdateCheck;
   private String version;
-  private String applicationId;
+  private String application;
   private boolean enableJarSplitting;
   private String jarSplittingExcludes;
   private String compileEncoding;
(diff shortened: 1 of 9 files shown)


loosebazooka:
@meltsufin

loosebazooka:
It actually looks like application doesn't even do anything here? Should we just not expose it?

meltsufin:
@loosebazooka How are you determining the command syntax and whether parameters are used?

loosebazooka:
Stage syntax is just appcfg [options] stage source destination

loosebazooka:
@meltsufin ptal

src/main/java/com/google/cloud/tools/app/impl/config/DefaultStageStandardConfiguration.java
```diff
@@ -29,7 +29,7 @@
   private boolean enableQuickstart;
   private boolean disableUpdateCheck;
   private String version;
-  private String applicationId;
+  private String application;
```
meltsufin: Did you mean to remove application altogether?

loosebazooka: oops yes

loosebazooka: updated

meltsufin:
:ship:

PR #25, opened 2016-05-03
Process runner that can run sync or async and with listeners

meltsufin:
Changes the signature of ProcessRunner.run to return void. Replaces SimpleProcessRunner with DefaultProcessRunner that can run sync or async and can also forward subprocess output to listeners. The process exit code can be read using ProcessExitListener.


.idea/codeStyleSettings.xml
```diff
@@ -297,8 +297,15 @@
         <ADDITIONAL_INDENT_OPTIONS fileType="yml">
```
loosebazooka: maybe take this code style thing outta this pr

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
```diff
@@ -0,0 +1,193 @@
...
+    this.processBuilder = processBuilder;
+  }
+
+  /**
+   * Executes a shell command.
+   *
+   * @param command The shell command to execute
+   */
+  public void run(String[] command) throws ProcessRunnerException {
+
```
loosebazooka: don't you have to set the command on the process builder first?

meltsufin: good catch! otherwise it's pretty useless!

.idea/codeStyleSettings.xml
loosebazooka: what about this?

loosebazooka:
:rocket: after tests pass

PR #29, opened 2016-05-03
update app engine tools version to 1.9.36

loosebazooka:


loosebazooka:
@meltsufin also see #30

meltsufin:
:rocket:

pom.xml
```diff
@@ -9,7 +9,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <appengine.target.version>1.9.34</appengine.target.version>
+    <appengine.target.version>1.9.36</appengine.target.version>
```
patflynn: I'm wondering if we need this.  It seems like it's conceivable that we could always use the latest for v1 App Engine APIs and tooling.

meltsufin: Do you mean using "LATEST" as the version?

patflynn: maybe?  Why not?

loosebazooka: I would rather we just use whatever is in the user's cloud sdk

patflynn: Yeah that's what I'm saying!

loosebazooka:
dropping this.

PR #34, opened 2016-05-04
Use cloudsdk appcfg

### LINKED ISSUE #30: Use appcfg directly from the sdk reference instead of using a maven dependency
see https://github.com/GoogleCloudPlatform/gradle-appengine-plugin/blob/master/src/main/groovy/com/google/appengine/task/appcfg/AppConfigTaskTemplate.groovy#L73

This way we don't have to update the maven dependency like we are in #29

loosebazooka:
Fixes #30

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,6 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <appengine.target.version>1.9.36</appengine.target.version>
   </properties>
 
   <dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -14,11 +14,6 @@
 
   <dependencies>
     <dependency>
-      <groupId>com.google.appengine</groupId>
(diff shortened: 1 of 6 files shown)


loosebazooka:
@meltsufin @patflynn

pom.xml
```diff
@@ -9,16 +9,11 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <appengine.target.version>1.9.34</appengine.target.version>
+    <appengine.target.version>1.9.36</appengine.target.version>
```
meltsufin: Do you still need this property?

meltsufin:
The test needs to be fixed to compile.

pom.xml
loosebazooka: nah, you're right

loosebazooka:
weird that mine local test didn't catch this :\

meltsufin:
:ship: it

PR #40, opened 2016-05-05
Use maven-bundle-plugin to add OSGi metadata to consume within Eclipse

briandealwis:
I've put `.java` into the bundle name.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,6 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <bundle.symbolicname>com.google.cloud.tools.app.java</bundle.symbolicname>
   </properties>
 
   <distributionManagement>
@@ -104,7 +103,7 @@
         <version>3.0.1</version>
         <configuration>
           <instructions>
-            <Bundle-SymbolicName>${bundle.symbolicname}</Bundle-SymbolicName>
+            <Bundle-SymbolicName>com.google.cloud.tools.app.lib</Bundle-SymbolicName>
             <Bundle-RequiredExecutionEnvironment>JavaSE-1.7</Bundle-RequiredExecutionEnvironment>
           </instructions>
         </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -9,6 +9,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <bundle.symbolicname>com.google.cloud.tools.app.java</bundle.symbolicname>
   </properties>
 
   <distributionManagement>
@@ -70,6 +71,12 @@
             <version>2.0.5</version>
           </dependency>
         </dependencies>
+        <configuration>
+          <archive>
+            <!-- include manifest produced by maven-bundle-plugin -->
+            <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+          </archive>
+        </configuration>
       </plugin>
 
       <plugin>
@@ -86,6 +93,27 @@
           <execution>
             <goals>
               <goal>check</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <version>3.0.1</version>
+        <configuration>
+          <instructions>
+            <Bundle-SymbolicName>${bundle.symbolicname}</Bundle-SymbolicName>
+            <Bundle-RequiredExecutionEnvironment>JavaSE-1.7</Bundle-RequiredExecutionEnvironment>
+          </instructions>
+        </configuration>
+        <executions>
+          <execution>
+            <id>bundle-manifest</id>
+            <phase>process-classes</phase>
+            <goals>
+              <goal>manifest</goal>
             </goals>
           </execution>
         </executions>


loosebazooka:
Can you elaborate on what this does?

briandealwis:
Eclipse is based on OSGi, and this metadata is necessary to be consumed by the [GoogleCloudPlatform/gcloud-tools-eclipse project](https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools).  The `maven-bundle-plugin` computes OSGi metadata and writes it into the `META-INF/MANIFEST.MF`:
- provides a unique name and version
- documents the packages it imports and exports, and their implementation dependencies
- documents the compatible java runtime
OSGi provides strong classpath isolation between bundles.

The jar can be used outside of OSGi — this will have no effect on other consumers.

The `MANIFEST.MF` will look like the following:
```
Manifest-Version: 1.0
Bnd-LastModified: 1462395584686
Build-Jdk: 1.8.0_77
Built-By: bsd
Bundle-ManifestVersion: 2
Bundle-Name: Google Cloud Java management client library
Bundle-RequiredExecutionEnvironment: JavaSE-1.7
Bundle-SymbolicName: com.google.cloud.tools.app.java
Bundle-Version: 0.1.0.SNAPSHOT
Created-By: Apache Maven Bundle Plugin
Export-Package: com.google.cloud.tools.app.api;version="0.1.0",com.googl
 e.cloud.tools.app.api.deploy;version="0.1.0";uses:="com.google.cloud.to
 ols.app.api",com.google.cloud.tools.app.api.devserver;version="0.1.0";u
 ses:="com.google.cloud.tools.app.api",com.google.cloud.tools.app.api.ge
 nconfig;version="0.1.0";uses:="com.google.cloud.tools.app.api",com.goog
 le.cloud.tools.app.api.module;version="0.1.0";uses:="com.google.cloud.t
 ools.app.api"
Import-Package: com.google.cloud.tools.app.api;version="[0.1,1)",com.goo
 gle.cloud.tools.app.api.deploy;version="[0.1,1)",com.google.cloud.tools
 .app.api.devserver;version="[0.1,1)",com.google.cloud.tools.app.api.gen
 config;version="[0.1,1)",com.google.cloud.tools.app.api.module;version=
 "[0.1,1)",com.google.common.base;version="[17.0,18)"
Originally-Created-By: Apache Maven Bundle Plugin
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.7))"
Tool: Bnd-3.0.0.201509101326
```

loosebazooka:
That's interesting. A couple of question for things I don't understand
1. Does this affect our bundling strategy (This wont make fat jars, right)?
2. Can eclipse resolve the dependencies directly for these?
3. How exactly will we (as IDE agnostic developers) be restricted in our library's use of dependencies that aren't OSGI spec? Can it slow us down as we add new non-osgi compatible dependencies for example (com.bob.strings:best-string-parser-ever:1.2.3)?

briandealwis:
This patch only changes the MANIFEST.MF (no fat jars).  Eclipse / OSGi uses the `Import-Package` and `Export-Package` entries to wire up bundles to each other; if you look carefully, this bundle now only exports what it imports except for `com.google.common.base`, which comes from Guava, which also carries OSGi metadata.

New dependencies will be cause new packages to be added to the `Import-Package` header, and will only affect us; we'll need to find providers of those packages and otherwise wrap them ourselves.  I held off submitting the patch until I saw you'd removed the the app-engine-sdk for partially that reason.

That said, if you think you'll be pulling in all sorts of dependencies, then it may be better to hold off on this patch and we'll just wrap `app-tools-lib-for-java` and its dependencies ourselves.

loosebazooka:
I think this is actually okay, just wanna hold off till after Friday, perhaps Monday (May 9th) when we're relatively stable. I'd like to play around with it a little first before incorporating it in.

pom.xml
```diff
@@ -9,6 +9,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <bundle.symbolicname>com.google.cloud.tools.app.java</bundle.symbolicname>
```
meltsufin: Maybe `.lib` instead of `.java`? Also, can you just use this directly in the plugin section instead of defining a property? I think it would be more clear as to what it's for.

briandealwis:
> Maybe .lib instead of .java?

Makes sense — I originally put `.java` since the project seemed focused on Java support, but that can be changed of course.

meltsufin:
LGTM, pending @loosebazooka's thumbs up

loosebazooka:
:+1:

meltsufin:
@briandealwis Can you rebase? And then we can merge it into master.

briandealwis:
Bah, _Update branch_ doesn't rebase.

briandealwis:
Rebased @meltsufin

meltsufin:
Thanks!

elharo:
Why/how is this only exporting com.google.cloud.tools.app.api? I think this may have broke with the package rename.

patflynn:
for me the 'api' package is a signal that only classes in that package are
part of the public api,  however it seems like that's not the case since
CloudSdk at least is needed to actually do anything with this library.

Basically if we have an 'api' package then everything else should be under
internal and not for public consumption.

meltsufin:
Currently we're using 'api' package to designate interfaces and classes that are not Cloud SDK specific. So, if we ever move away from using gcloud, we won't have to touch this package. The 'cloudsdk' package itself has an 'internal' subpackage that is not for consumption by clients.

patflynn:
I don't think that's meaningful to clients.  In the end the contract is the
set of public classes I need to use to do stuff with this library.  Perhaps
a solution is to just bring all of the .api package's contents up one level
and removing .api as a concept all together.

I do agree that we don't want to expose through our service methods the
fact that they're implementing their actions through the gcloud CLI, and
that we want nice Java-esque APIs to wrap those interactions,  but it's OK
to expose the bits of the public API that require Cloud SDK configuration
as part of the 'public API'.

meltsufin:
Created an issue to fix this: #110

PR #42, opened 2016-05-05
Change boolean to Boolean

loosebazooka:
- Allowing null option to delegate to the underlying tool to
respect its defaults (false OR true)
- Explicitly pass true and false values to boolean flags

### DIFF
--- a/src/main/java/com/google/cloud/tools/app/api/deploy/DeployConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/app/api/deploy/DeployConfiguration.java
@@ -28,17 +28,17 @@ public interface DeployConfiguration {
 
   String getDockerBuild();
 
-  boolean isForce();
+  Boolean getForce();
 
   String getImageUrl();
 
   String getProject();
 
-  boolean isPromote();
+  Boolean getPromote();
 
   String getServer();
 
-  boolean isStopPreviousVersion();
+  Boolean getStopPreviousVersion();
 
   String getVersion();
 
--- a/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineDeployment.java
+++ b/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineDeployment.java
@@ -19,8 +19,8 @@ import com.google.cloud.tools.app.api.deploy.AppEngineDeployment;
 import com.google.cloud.tools.app.api.deploy.DeployConfiguration;
 import com.google.cloud.tools.app.impl.cloudsdk.internal.process.ProcessRunnerException;
 import com.google.cloud.tools.app.impl.cloudsdk.internal.sdk.CloudSdk;
+import com.google.cloud.tools.app.impl.cloudsdk.util.Args;
 import com.google.common.base.Preconditions;
-import com.google.common.base.Strings;
 
 import java.io.File;
 import java.util.ArrayList;
@@ -39,15 +39,15 @@ public class CloudSdkAppEngineDeployment implements AppEngineDeployment {
   }
 
   @Override
-  public void deploy(DeployConfiguration configuration) throws AppEngineException {
-    Preconditions.checkNotNull(configuration);
-    Preconditions.checkNotNull(configuration.getDeployables());
-    Preconditions.checkArgument(configuration.getDeployables().size() > 0);
+  public void deploy(DeployConfiguration config) throws AppEngineException {
+    Preconditions.checkNotNull(config);
+    Preconditions.checkNotNull(config.getDeployables());
+    Preconditions.checkArgument(config.getDeployables().size() > 0);
     Preconditions.checkNotNull(sdk);
 
     List<String> arguments = new ArrayList<>();
     arguments.add("deploy");
-    for (File deployable : configuration.getDeployables()) {
+    for (File deployable : config.getDeployables()) {
       if (!deployable.exists()) {
         throw new IllegalArgumentException(
             "Deployable " + deployable.toPath().toString() + " does not exist.");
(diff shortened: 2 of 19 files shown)


loosebazooka:
@meltsufin @etanshaul

etanshaul:
lgtm

.idea/codeStyleSettings.xml
```diff
@@ -299,6 +299,6 @@
         </ADDITIONAL_INDENT_OPTIONS>
       </value>
     </option>
-    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Google Style" />
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Google Configuration Checker Style" />
```
meltsufin: Are you trying to sneak in the style stuff?

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineDeployment.java
```diff
@@ -79,17 +80,19 @@ public void deploy(DeployConfiguration configuration) throws AppEngineException
...
 
     if (!Strings.isNullOrEmpty(configuration.getServer())) {
       arguments.add("--server");
       arguments.add(configuration.getServer());
     }
 
-    if (configuration.isStopPreviousVersion()) {
+    if (configuration.getStopPreviousVersion() != null) {
       arguments.add("--stop-previous-version");
+      arguments.add(configuration.getStopPreviousVersion().toString());
```
meltsufin: If it's false, you should add " --no-stop-previous-version"

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineDeployment.java
```diff
@@ -79,17 +80,19 @@ public void deploy(DeployConfiguration configuration) throws AppEngineException
       arguments.add(configuration.getProject());
     }
 
-    if (configuration.isPromote()) {
+    if (configuration.getPromote() != null) {
       arguments.add("--promote");
+      arguments.add(configuration.getPromote().toString());
```
meltsufin: If false, add --no-promote

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineDeployment.java
etanshaul: Hmm. Are you sure about that? To me it seems like they should be separate Booleans (since they are distinct flags). Assuming we intend the Booleans to have a 1-to-1 mapping to gcloud flags.

src/main/java/com/google/cloud/tools/app/api/deploy/DeployConfiguration.java
```diff
@@ -28,17 +28,17 @@
 
   String getDockerBuild();
 
-  boolean isForce();
```
meltsufin: I think force should remain a boolean because null is identical to false.

etanshaul: I am also unsure about this comment. I guess it depends on the philosophy of how we map to gcloud.

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineDeployment.java
meltsufin: Boolean technically has 3 states: true, false, null for which map nicely to: --stop, --no-stop, and <blank>.

src/main/java/com/google/cloud/tools/app/api/deploy/DeployConfiguration.java
meltsufin: This is a case where you have only two states and boolean seems to map better.

etanshaul: Agreed that there are 2 states. But you can omit that flag altogether from the gcloud command right? Just because it defaults to false now when you omit it doesn't mean that the default couldn't change in the future.

etanshaul: On second thought since its really just a flag (it doesn't take in a value like -force true / false). Maybe what I'm saying doesn't make sense.

meltsufin:
@loosebazooka As you can see from my code comments, I'm now thinking we shouldn't really be so aggressive with switching everything to Boolean. Only options that map to --flag, --no-flag, and blank [some default] should be of type Boolean. What do you think?

.idea/codeStyleSettings.xml
loosebazooka: GODDAMNIT, I usually only add src, I've made a huge mistake

src/main/java/com/google/cloud/tools/app/api/deploy/DeployConfiguration.java
loosebazooka: It's actually possible this PR is completely unnecesary, we don't set any flags unnecessarily

loosebazooka: and gcloud binary flags don't take values, if we want to add the no-stop-previous-version flag we can

loosebazooka:
should we just expose no-promote and promote? @meltsufin or we do special work on the promote parsing?

loosebazooka:
looks like every boolean gcloud flag has a --no equivalent so i think what I'm going to do is:
1. use Boolean everywhere for consistency
2. create a (ex:force) --force and --no-force for each Boolean flag

meltsufin:
LGTM :ship:
Good stuff with the Args utility, and it get us closer to the auto-generation of args with reflection.

Issue #45, opened 2016-05-05
Add --format flag support


etanshaul:
For some clients its useful to be able to control the output format of cammand via the `--format` flag. For example `--format=json` so that the output is easily parseable.

PR #46, opened 2016-05-05
Add common configuration

etanshaul:
See #45 

@loosebazooka @meltsufin what do you think about something like this? I only added usage of the common configuration to CloudSdkAppEngineDeployment for now, but you get the idea.

### DIFF
--- a/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/sdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/sdk/CloudSdk.java
@@ -18,6 +18,7 @@ import com.google.cloud.tools.app.api.AppEngineException;
 import com.google.cloud.tools.app.impl.cloudsdk.internal.process.DefaultProcessRunner;
 import com.google.cloud.tools.app.impl.cloudsdk.internal.process.ProcessRunner;
 import com.google.cloud.tools.app.impl.cloudsdk.internal.process.ProcessRunnerException;
+import com.google.cloud.tools.app.impl.cloudsdk.util.Args;
 import com.google.common.base.Joiner;
 
 import java.io.File;
@@ -43,6 +44,8 @@ public class CloudSdk {
 
   private Path sdkPath = null;
   private ProcessRunner processRunner = null;
+
+  private String appCommandOutputFormat;
 
   /**
    * Initializes an instance using the default location of Cloud SDK and default ProcessRunner.
@@ -96,6 +99,8 @@ public class CloudSdk {
     command.add("preview");
     command.add("app");
     command.addAll(args);
+
+    command.addAll(Args.string("format", appCommandOutputFormat));
 
     outputCommand(command);
 
@@ -194,4 +199,7 @@ public class CloudSdk {
     }
   }
 
+  public void setAppCommandOutputFormat(String appCommandOutputFormat) {
+    this.appCommandOutputFormat = appCommandOutputFormat;
+  }
 }


src/main/java/com/google/cloud/tools/app/api/Configuration.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.app.api;
+
+/**
+ * Base Cloud SDK configuration for configuration spanning multiple commands.
+ */
+public interface Configuration {
+    String getFormat();
```
loosebazooka: The problem is that the configurations are agnostic to the implementation.

Format may not be relevant to non cloud sdk implementations. You could potentially drop this kind of configuration in the cloudSdk implementations of the actions though.

etanshaul: Yea good point.

Regarding your suggestion: the way its designed the cloud sdk implementations only take in an implementation-agnostic configuration. I don't think we want to hard code this config in the action. Unless I misunderstood what you are suggesting?

etanshaul:
Changed the approach. Setting it on the CloudSdk - this actually doesn't feel _too_ bad (see the test to get a feel for how it would be consumed).

loosebazooka:
This looks good, I think you need to fix up the checkstyle violations though.

etanshaul:
checkstyle is happy now

meltsufin:
It could be confusing to the clients that they're setting configuration on the whole class, but it only applies to runAppCommand, and not to runDevAppServerCommand and runAppCfgCommand. They can already use runAppCommand(List<String> args) to pass in any additional flags. It seems that this logic would sit better with the CloudSdkAppEngine* classes because they construct the arguments.

etanshaul:
Yea, valid point. Do you think this is something that naming could help with? 

I originally went down the path of adding it to CloudSdkAppEngine* but ran into usability issues. For example:

AppEngineDeployment deployment = new CloudSdkAppEngineDeployment();

This is a valid (and I'd say proper) way to instantiate a deployment action. @loosebazooka correctly pointed out at that it wouldn't make sense to add this configuration to CloudSdk agnostic interfaces. So we have an API usability issue here. I'm open to any other design options here.

meltsufin:
It's true that it doesn't make sense to add this to the interfaces, but what's wrong with this?
```java
CloudSdkAppEngineDeployment deployment = new CloudSdkAppEngineDeployment(..);
deployment.setFormat("json");
deployment.deploy();
```

etanshaul:
Nothing, I guess.

meltsufin:
On a second thought, scoping this "format" concern to the CloudSdk class seems less invasive to the API.
I would just suggest creating a setAppCommandOutputFormat setter directly on the CloudSdk class, rather than passing a Configuration object. This would be more in line with the rest of the CloudSdk configuration, like the path and the process runner, that are not set using a Configuration object.

meltsufin:
This can also be rolled into the CloudSdk Builder when that's ready. See: #51

etanshaul:
Sure. But just to point out: a change I plan on making soon is adding in support for the `credential-file-override` field which is another cross-cutting flag. This will mean more and more setters on CloudSdk (or its builder) as we add more common config - I think this is fine, especially once there's a builder.

etanshaul:
@meltsufin updated based on your comment

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/sdk/CloudSdk.java
```diff
@@ -72,6 +75,11 @@ public void runAppCommand(List<String> args) throws ProcessRunnerException {
     command.add("app");
     command.addAll(args);
 
+    if (!Strings.isNullOrEmpty(appCommandOutputFormat)) {
```
meltsufin: Can you use Args util here?

meltsufin:
Other than my comment about using Args, LGTM

src/test/java/com/google/cloud/tools/app/CloudSdkAppEngineDeploymentTest.java
```diff
@@ -59,6 +58,9 @@
   public void setUp() throws IOException {
     appYaml1 = tmpDir.newFile("app1.yaml");
     appYaml2 = tmpDir.newFile("app2.yaml");
+
+    sdk.setAppCommandOutputFormat("json");
```
loosebazooka: Why are you setting this on this test?

src/test/java/com/google/cloud/tools/app/CloudSdkAppEngineDeploymentTest.java
```diff
@@ -38,6 +33,10 @@
 import java.util.Arrays;
 import java.util.List;
 
+import static org.mockito.Matchers.eq;
```
loosebazooka: are these being used?

etanshaul: yea, they are being used in every test it looks like

src/test/java/com/google/cloud/tools/app/CloudSdkAppEngineDeploymentTest.java
etanshaul: I originally added it just to show what consuming it would look like. It doesn't do anything.

loosebazooka:
lgtm.

PR #56, opened 2016-05-10
Credential file override flag support

### LINKED ISSUE #41: Allow passing in auth context to supported gcloud operations
For consumers of this API that can't rely on having to do `gcloud auth login` there needs to be a way to pass in the auth context.

An initial solution could be to support the `--credential-file-override` flag and require consumers to do the staging of the credential file themselves.

etanshaul:
fixes #41


meltsufin:
LGTM

etanshaul:
Want to merge in the builder first and I'll update this?

meltsufin:
Sure, let me do that.

etanshaul:
Updated to use builder.

meltsufin:
:ship:

PR #58, opened 2016-05-10
Expose the current running process in DefaultProcessRunner

### LINKED ISSUE #57: Allow clients to destroy the running process
Both for sync or async.

Need to expose ability to kill a process that was started for consumers using a custom ProcessRunner.

meltsufin:
Fixes #57 allowing clients to destroy an executing process.

### DIFF
--- a/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
+++ b/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
@@ -31,6 +31,8 @@ public class DefaultProcessRunner implements ProcessRunner {
   private ProcessOutputLineListener stdErrLineListener;
   private ProcessExitListener exitListener;
 
+  private Process process;
+
   /**
    * Create the process runner with a default process builder, with inheritIO enabled.
    */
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
```diff
@@ -51,12 +53,18 @@ public DefaultProcessRunner(ProcessBuilder processBuilder) {
    *
    * @param command The shell command to execute
    */
-  public void run(String[] command) throws ProcessRunnerException {
+  public synchronized void run(String[] command) throws ProcessRunnerException {
```
etanshaul: Can you explain why you synchronized here?

meltsufin: There is a race condition between checking if the process is null and starting one. I should probably scope the synchronization to the affected part of the method only. Doing it on the whole method also has the effect of blocking other threads when they call run and there is a synchronous execution in progress. It's probably not the behavior we want.

etanshaul:
lgtm

PR #64, opened 2016-05-11
Add environment variable setters to cloudsdk builder

### LINKED ISSUE #39: make sure we allow setting the environment variable to identifying the deployment source

etanshaul:
fixes #39 

@loosebazooka @meltsufin, thoughts?

### DIFF
--- a/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/ProcessRunner.java
+++ b/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/ProcessRunner.java
@@ -14,11 +14,15 @@
 
 package com.google.cloud.tools.app.impl.cloudsdk.internal.process;
 
+import java.util.Map;
+
 /**
  * Interface for running arbitrary commands.
  */
 public interface ProcessRunner {
 
   void run(String[] command) throws ProcessRunnerException;
+
+  void run(String[] command, Map<String, String> environment) throws ProcessRunnerException;
 
 }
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/ProcessRunner.java
```diff
@@ -14,11 +14,15 @@
...
+import java.util.Map;
+
 /**
  * Interface for running arbitrary commands.
  */
 public interface ProcessRunner {
 
   void run(String[] command) throws ProcessRunnerException;
 
+  void run(String[] command, Map<String, String> environment) throws ProcessRunnerException;
```
meltsufin: I'm not sure about overloading the run method. Maybe we can just add a `setEnvironment(Map<String, String> environment)`?

etanshaul: Yea, I think thats ok too. I'll update it

etanshaul:
@meltsufin removed overloading and added a setter

meltsufin:
Looks good.
...All these members with appCommand* prefix are starting to make CloudSdk look like its trying to do too much, but that's a thing we should deal with separately.

etanshaul:
I agree. I actually have one more to add based on @chanseokoh on the issue. It might be worthwhile (in a different PR) to extract those out into a separate builder object that's passed in to the sdk.

etanshaul:
Added support for the metrics environment version env var.

meltsufin:
LGTM

PR #66, opened 2016-05-12
Adding source plugin and changing jar plugin version for performance

### LINKED ISSUE #65: Add sources jar to build
Sources jar is needed for IDEs to provide documentation and debugging support.

meltsufin:
Fixes #65.

I downgraded the maven-jar-plugin to 2.3 and not using the latest version of maven-source-plugin because these versions run much faster. 'mvn clean install' runs in less than 8 seconds now for me.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,27 +55,30 @@
         </configuration>
       </plugin>
 
-      <!-- Work around a performance issue with Maven and Java 7:
-        Maven has a dependency that does a getgrgid for each file.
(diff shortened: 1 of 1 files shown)


loosebazooka:
weird, why is this happening?

meltsufin:
It looks like there is an issue with plexus-io versions 2.0.0+.

loosebazooka:
:ship:

PR #67, opened 2016-05-12
Allow waiting for successful start of async commands

### LINKED ISSUE #60: Detect when Dev App Server has started

### LINKED ISSUE #63: When staging for standard environment, do not attempt to copy Dockerfile if it doesn't exist

meltsufin:
This fixes #60 which is about detecting when Dev App Server has started.
This also adds a builder to the DefaultProcesRunner.
Also fixes #63 

Notice that this adds a general mechanism for detecting when an asynchronously started operation has succeeded.

The clients are also no longer allowed to configure and provide the ProcessBuilder because if they set inheritIO to true, capturing output doesn't work. Instead, clients that need to consume the process output should use the ProcessOutputLineListener.

@loosebazooka @etanshaul

### DIFF
--- a/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineStandardStaging.java
+++ b/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineStandardStaging.java
@@ -74,7 +74,7 @@ public class CloudSdkAppEngineStandardStaging implements AppEngineStandardStagin
 
       cloudSdk.runAppCfgCommand(arguments);
 
-      if (dockerfile != null) {
+      if (dockerfile != null && dockerfile.toFile().exists()) {
         Files.copy(dockerfile, dockerfileDestination, StandardCopyOption.REPLACE_EXISTING);
       }
 
--- a/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
+++ b/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
@@ -15,6 +15,7 @@
 package com.google.cloud.tools.app.impl.cloudsdk.internal.process;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
@@ -26,38 +27,37 @@ import java.util.Scanner;
  */
 public class DefaultProcessRunner implements ProcessRunner {
   private boolean async;
-  private ProcessOutputLineListener stdOutLineListener;
-  private ProcessOutputLineListener stdErrLineListener;
+  private List<ProcessOutputLineListener> stdOutLineListeners;
+  private List<ProcessOutputLineListener> stdErrLineListeners;
   private ProcessExitListener exitListener;
-  private AsyncProcessStartWaiter asyncProcessStartWaiter;
 
+  private ProcessBuilder processBuilder = new ProcessBuilder();
   private Process process;
 
   private Map<String, String> environment;
 
   private DefaultProcessRunner(Builder builder) {
     this.async = builder.async;
-    this.stdOutLineListener = builder.stdOutLineListener;
-    this.stdErrLineListener = builder.stdErrLineListener;
+    this.stdOutLineListeners = builder.stdOutLineListeners;
+    this.stdErrLineListeners = builder.stdErrLineListeners;
     this.exitListener = builder.exitListener;
-    if (async) {
-      this.asyncProcessStartWaiter = builder.asyncProcessStartWaiter;
-    }
   }
 
   /**
    * Executes a shell command.
    *
+   * <p>If any output listeners were configured, output will go to them only. Otherwise, process
+   * output will be redirected to the caller via inheritIO.
+   *
    * @param command The shell command to execute
    */
   public void run(String[] command) throws ProcessRunnerException {
     try {
-
-      if (asyncProcessStartWaiter != null) {
-        asyncProcessStartWaiter.reset();
-      }
-
+      // configure process builder
       final ProcessBuilder processBuilder = new ProcessBuilder();
+      if (stdErrLineListeners.size() == 0 && stdOutLineListeners.size() == 0) {
+        processBuilder.inheritIO();
+      }
       if (environment != null) {
         processBuilder.environment().putAll(environment);
       }
@@ -83,10 +83,6 @@ public class DefaultProcessRunner implements ProcessRunner {
         syncRun(process);
       }
 
-      if (asyncProcessStartWaiter != null) {
-        asyncProcessStartWaiter.await();
-      }
-
     } catch (IOException | InterruptedException | IllegalThreadStateException e) {
       throw new ProcessRunnerException(e);
     }
@@ -97,44 +93,6 @@ public class DefaultProcessRunner implements ProcessRunner {
    */
   public void setEnvironment(Map<String, String> environment) {
     this.environment = environment;
-  }
-
-  /**
-   * Sets the process execution to be asynchronous
-   *
-   * @param async False by default.
-   */
-  public void setAsync(boolean async) {
-    this.async = async;
-  }
-
-  /**
-   * Set the listener for standard output of the subprocess. Note that this will not work if you set
-   * inheritIO.
-   *
-   * @param stdOutLineListener Can be null.
-   */
-  public void setStdOutLineListener(ProcessOutputLineListener stdOutLineListener) {
-    this.stdOutLineListener = stdOutLineListener;
-  }
-
-  /**
-   * Set the listener for standard error output of the subprocess. Note that this will not work if
-   * you set inheritIO.
-   *
-   * @param stdErrLineListener Can be null.
-   */
-  public void setStdErrLineListener(ProcessOutputLineListener stdErrLineListener) {
-    this.stdErrLineListener = stdErrLineListener;
-  }
-
-  /**
-   * Sets the subprocess exit listener for collecting the exit code of the subprocess.
-   *
-   * @param exitListener Can be null.
-   */
-  public void setExitListener(ProcessExitListener exitListener) {
-    this.exitListener = exitListener;
   }
 
   /**
@@ -176,17 +134,13 @@ public class DefaultProcessRunner implements ProcessRunner {
 
   private void consumeLine(String line, boolean errorStream) {
     if (errorStream) {
-      if (stdErrLineListener != null) {
+      for (ProcessOutputLineListener stdErrLineListener : stdErrLineListeners) {
         stdErrLineListener.outputLine(line);
       }
     } else {
-      if (stdOutLineListener != null) {
+      for (ProcessOutputLineListener stdOutLineListener : stdOutLineListeners) {
         stdOutLineListener.outputLine(line);
       }
-    }
-
-    if (asyncProcessStartWaiter != null) {
-      asyncProcessStartWaiter.inputLine(line);
     }
   }
 
@@ -241,10 +195,10 @@ public class DefaultProcessRunner implements ProcessRunner {
 
   public static class Builder {
     private boolean async = false;
-    private ProcessOutputLineListener stdOutLineListener;
-    private ProcessOutputLineListener stdErrLineListener;
+    private List<ProcessOutputLineListener> stdOutLineListeners = new ArrayList<>();
+    private List<ProcessOutputLineListener> stdErrLineListeners = new ArrayList<>();
     private ProcessExitListener exitListener;
-    private AsyncProcessStartWaiter asyncProcessStartWaiter;
+    private WaitingProcessOutputLineListener waitingProcessOutputLineListener;
 
     /**
      * Whether to run commands asynchronously.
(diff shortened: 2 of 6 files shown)


src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
```diff
@@ -124,49 +133,79 @@ public void setStdErrLineListener(ProcessOutputLineListener stdErrLineListener)
...
+    stdOutThread.setDaemon(true);
+    stdOutThread.start();
+  }
+
+  private void handleWaitSuccess() throws InterruptedException, ProcessRunnerException {
+    try {
+      if (async && waitSuccessMessage != null && waitSuccessTimeoutSeconds != 0) {
+        log.info("Waiting " + waitSuccessTimeoutSeconds
+            + " seconds for the operation to succeed...");
+        if (!waitSuccessLatch.await(waitSuccessTimeoutSeconds, TimeUnit.SECONDS)) {
```
etanshaul: won't this cause synchronous runs to block?

etanshaul: oh wait never mind, you're only doing this for async

meltsufin: even for sync runs this shouldn't block because this would be called after the process exits

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
```diff
@@ -219,4 +258,72 @@ public void run() {
...
+      this.exitListener = exitListener;
+      return this;
+    }
+
+    /**
+     * The message to look for in the standard or error output of the process to consider it to be
+     * successfully started. If the message is not seen within the specified timeout {@link
+     * #waitSuccessTimeoutSeconds(int)}, a {@link ProcessRunnerException} will be thrown.
+     */
+    public Builder waitSuccessMessage(String waitSuccessMessage) {
```
etanshaul: What is an example use-case of this?

etanshaul: Also, if the waitSuccessMessage isn't set, then the timeout mechanism is ignored?

meltsufin: Yes. It's used when dev app server is started asynchronously: https://github.com/GoogleCloudPlatform/app-maven-plugin/blob/sync_with_lib/src/main/java/com/google/cloud/tools/maven/RunAsyncMojo.java

etanshaul: ah. makes sense now.

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineStandardStaging.java
```diff
@@ -74,7 +74,7 @@ public void stageStandard(StageStandardConfiguration config) throws AppEngineExc
 
       cloudSdk.runAppCfgCommand(arguments);
 
-      if (dockerfile != null) {
+      if (dockerfile != null && dockerfile.toFile().exists()) {
```
loosebazooka: Should we error if the dockerfile doesn't exist?

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
```diff
@@ -19,34 +19,36 @@
...
-   * Create the process runner with the provided process builder.
-   */
-  public DefaultProcessRunner(ProcessBuilder processBuilder) {
-    this.processBuilder = processBuilder;
+  private DefaultProcessRunner(Builder builder) {
+    this.async = builder.async;
+    this.stdOutLineListener = builder.stdOutLineListener;
+    this.stdErrLineListener = builder.stdErrLineListener;
+    this.exitListener = builder.exitListener;
+    this.waitSuccessMessage = builder.waitSuccessMessage;
```
loosebazooka: I feel like we should do something like "exitListener" here. If wait success message and wait success timeout are tied together, perhaps they should be one object? WaitDirective or something?

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineStandardStaging.java
meltsufin: Possibly, but for now, this makes it consistent with flex staging.

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
meltsufin: Can you elaborate on the listener idea?
Good idea about combining wait message and timeout. I'll change that.

loosebazooka: I really only meant the one thing - combining them into a single input. Similar to how exitListener covered the exit cases. It still feels a little funny to me to define the Async behavior here. I feel like we're messing with the Process Execution too much and our implementation of async waiting is too specific to define within the process runner.

meltsufin: I think, ultimately, we might want to support regular expressions for successMessage, but otherwise, I think it's pretty generic for any process execution.

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
```diff
@@ -124,49 +133,79 @@ public void setStdErrLineListener(ProcessOutputLineListener stdErrLineListener)
...
+      if (stdErrLineListener != null) {
+        stdErrLineListener.outputLine(line);
+      }
+    } else {
+      if (stdOutLineListener != null) {
+        stdOutLineListener.outputLine(line);
+      }
+    }
+
+    if (waitSuccessMessage != null && line.contains(waitSuccessMessage)) {
```
loosebazooka: I think it makes sense to check if the countdown latch has been counted down to zero before doing a contains here?

meltsufin: Yes! Thanks.

loosebazooka:
I realized that this only works for single servers. If a user starts two modules for a server, they can't actually configure the asynchronous wait mechanism to detect both starts. I don't think it's unreasonable for the developer to configure their asynchronous run in their tooling as we were previously doing in gradle/maven. A fully configured line reading solution doesn't seem generic enough for all cases.

Removing redirectIO also makes it a little tedious for command line tool developers to write the output to screen, having to provide output processors for every invocation. I feel like if we want to provide tooling for a developer to run asychronous dev app server with waits, we can provide an implementation of the outputline listener that does it.

meltsufin:
This should work for multiple concurrent processes, as long as each uses a separate CloudSdk instance, which should be the case for Maven and Gradle. Are you thinking of a scenario where the CloudSdk instance is reused within one invocation of Maven or Gradle?

The problem with redirectIO is that if you expose it, then you can't support output listeners. We can make it easier to configure the output listeners for sysout by providing a SysOutProcessOutputLineListener, which will achieve the same thing as redirectIO.

Regarding an output listener that does the waiting, I'm open to it, but I don't think it's that straightforward. For example, where would you call the await() on the latch? Take a look at the AsyncProcessStartWaiter that I just added. I think it's pretty close to what you have in mind.

loosebazooka:
Oh, no you can actually start multiple modules with the same dev_appserver run.
```
dev_appserver.py app1.yaml app2.yaml
```

We don't necessarily need to expose redirectIO, but I think we should redirect IO by default.

Our current implementation doesn't differentiate between stderr and stdout. If the user wants to specifically read the stdout stream, then they should. But to me it makes sense that they would be doing this on the Stream listener itself.

I'm trying to figure out if it is possible to provide a good waiter as part of the outputLineListener class, and given that the waiter is waiting on the output (and not some other kind of wait), it would make sense to tie to the outputLineListener.

loosebazooka:
Something like this?

```java
public class WaitingProcessOutputLineListener implements ProcessOutputLineListener {

  private final CountDownLatch semaphore;
  private final String matchingString;

  public WaitingProcessOutputLineListener(String matchingString, int count) {
    semaphore = new CountDownLatch(count);
    this.matchingString = matchingString;
  }

  @Override
  public void outputLine(String line) {
    if (line.contains(matchingString)) {
      semaphore.countDown();
    }
  }

  public void wait() throws InterruptedException {
    semaphore.await(asdf,asdf);// or something);
  }
}
```

meltsufin:
Maybe if output listeners are not configured at all and there is no wait directive, we can redirectIO by default. That should be fine.

For the WaitingProcessOutputLineListener, we need to allow multiple listeners to be registered on CloudSdk.

I think this should be good. I'll try to integrate both ideas.

loosebazooka:
Yeah any stream without listeners should redirect IO. I don't mean this specifically but something like it? : https://github.com/GoogleCloudPlatform/app-tools-lib-for-java/blob/5c2368893e96a1fa7e36e8922006b1ca3df98b2e/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkCli.java#L163

meltsufin:
@loosebazooka See my latest change. I tried to address the inheritIO and waiting using the output line listener.

There is a little more burden on the clients now for handling start listening, but that should be OK.

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
```diff
@@ -25,38 +26,41 @@
...
-    processBuilder.inheritIO();
-  }
+  private Map<String, String> environment;
 
-  /**
-   * Create the process runner with the provided process builder.
-   */
-  public DefaultProcessRunner(ProcessBuilder processBuilder) {
-    this.processBuilder = processBuilder;
+  private DefaultProcessRunner(Builder builder) {
```
loosebazooka: This is a builder style I'm not familiar with. Wouldn't the builder call the DefaultProcessRunner's private constructor with all paramters?

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
```diff
@@ -25,38 +26,41 @@
...
+   * output will be redirected to the caller via inheritIO.
+   *
    * @param command The shell command to execute
    */
   public void run(String[] command) throws ProcessRunnerException {
     try {
-
+      // configure process builder
+      final ProcessBuilder processBuilder = new ProcessBuilder();
+      if (stdErrLineListeners.size() == 0 && stdOutLineListeners.size() == 0) {
```
loosebazooka: I feel like inheriting should be per stream?
processBuilder.redirectError(Redirect.INHERIT) ?

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
meltsufin: Effectively, it's just another style that's doing the same thing. I'll switch to the style you're suggesting though because it looks like it's more common.

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
meltsufin: Yes, good point.

meltsufin:
@loosebazooka Addressed your comments. PTAL.

meltsufin:
Actually, hold off for now, I'm making a few more small changes.

meltsufin:
OK, take a look now. I've made it easier for clients to configure waiting for the dev app server to start. Now all they have to do is set CloudSdk.Builder.runDevAppServerWaitSeconds and the listeners are automatically configured and the waiting is performed.

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
```diff
@@ -79,94 +97,64 @@ public void run(String[] command) throws ProcessRunnerException {
...
-      stdOutThread.start();
+      }
+    };
+    stdOutThread.setDaemon(true);
+    stdOutThread.start();
+  }
+
+
+  private void consumeLine(String line, boolean errorStream) {
+    if (errorStream) {
```
loosebazooka: This method seems to have lost its value, not doing much anymore after switching on the value of `errorstream`. Can we consume the line in the handleX methods now?

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/sdk/CloudSdk.java
```diff
@@ -51,15 +54,41 @@
...
+  private final WaitingProcessOutputLineListener runDevAppServerWaitListener;
+
+
+  private CloudSdk(Path sdkPath, String appCommandMetricsEnvironment,
+                   String appCommandMetricsEnvironmentVersion, Integer appCommandGsUtil,
+                   File appCommandCredentialFile, String appCommandOutputFormat, boolean async,
+                   List<ProcessOutputLineListener> stdOutLineListeners,
+                   List<ProcessOutputLineListener> stdErrLineListeners,
+                   ProcessExitListener exitListener,
+                   int runDevAppServerWaitSeconds) {
```
loosebazooka: I guess this is the best we can do combing all methods into one.  I remember you talking about separating everything out (by command), we should think about how to do that next.

loosebazooka: in another pr...

meltsufin: Yep, sounds good.

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
meltsufin: You're right.

loosebazooka:
waiting on tests... but :ship:

PR #68, opened 2016-05-12
Hide process runner behind cloud sdk cli

loosebazooka:
Expose everything a user needs directly though the CloudSdkCli object.


loosebazooka:
@meltsufin @etanshaul what do you guys think. We really don't want the user to have to deal with any of our process runner abstractions. I think I need to move a few more things out of internal pkg

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkCli.java
```diff
@@ -0,0 +1,261 @@
...
+     */
+    public Builder gcloudOutputFormat(String format) {
+      this.gcloudOutputFormat = format;
+      return this;
+    }
+
+    /**
+     * An environment variable for gcloud invocations.
+     */
+    public Builder environmentVariable(Environment key, String value) {
```
etanshaul: I wonder if there's any value in hiding the environment variables from the client too. I.e. pass them in like any other gcloud parameter so the client doesn't need to know if its a flag or env var.

meltsufin:
I think hiding the process runner stuff behind the CloudSdk will indeed make the library usage simpler.  What's the benefit of splitting CloudSdk into CloudSdk that just takes that sdk path, and the CloudSdkCli?

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkCli.java
```diff
@@ -0,0 +1,261 @@
...
+     */
+    public Builder async() {
+      this.async = true;
+      return this;
+    }
+
+    /**
+     * Client listener on the process output stream.
+     */
+    public Builder stdOutListener(ProcessOutputLineListener listener) {
```
etanshaul: +1 on hiding the process running and exposing the listeners like this.

meltsufin:
@loosebazooka  Do you want to close this one, since it's taken care of by #67?

PR #69, opened 2016-05-13
Proxy ProcessRunnner configuration in CloudSdk.Builder

meltsufin:
Attempt to address the goals of #68 building on top of #67.

@loosebazooka @etanshaul PTAL


etanshaul:
So looks like you removed inheritIO altogether? 

This looks good.

meltsufin:
inheritIO interferes with the output listeners. So, if the client needs process output, they should just use the listeners.

etanshaul:
lgtm

PR #71, opened 2016-05-16
Make args translation command specific

loosebazooka:


loosebazooka:
@meltsufin @etanshaul

meltsufin:
There is some code duplication between the three classes where some methods are identical. Can we eliminate that somehow? Maybe Args class with three static inner classes where they delegate to methods in the parent class?

loosebazooka:
Hrmm... interesting lemme try that.

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/args/GcloudArgs.java
```diff
@@ -0,0 +1,77 @@
...
+   * @return [--name, file.getAbsolutePath()] or [] if file=null.
+   */
+  public static List<String> get(String name, File file) {
+    return Args.filePath(name, file);
+  }
+
+  /**
+   * @return [key1=value1,key2=value2,...], [] if keyValueMapping=empty/null
+   */
+  public static List<String> keyValues(Map<?, ?> keyValueMapping) {
```
meltsufin: Should this be in Args as well?

loosebazooka:
rebased, PTAL

meltsufin:
:ship: :+1:

PR #72, opened 2016-05-16
Update codestyle with correct Java indent settings

### LINKED ISSUE #61: Fix IntelliJ codestyle settings

meltsufin:
Fixes #61.

Imported IntelliJ codestyle settings from here: https://github.com/google/styleguide/.
IntelliJ modified the original source after import, probably to fit the schema of the current version of the IDE.


loosebazooka:
I think you just need to use the right file here. This works fine for me.

loosebazooka:
Or update the styleguide file.

meltsufin:
Hmmmm, I'm not sure what's going on with my codestyle settings then.

loosebazooka:
It's possible I'm using a file from another source

loosebazooka:
so lets :ship: for now and figure out what we want to do with .idea

PR #73, opened 2016-05-16
search path variable for Cloud Sdk

### LINKED ISSUE #62: Improve Cloud SDK location detection
Use something like a `which` command to find the gcloud installation.

meltsufin:
Fixes #62


src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/sdk/PathResolver.java
```diff
@@ -25,39 +29,70 @@
...
-        sdkDir = System.getProperty("user.home") + "/google-cloud-sdk";
-        if (!new File(sdkDir).exists()) {
-          // try devshell VM:
-          sdkDir = "/google/google-cloud-sdk";
-          if (!new File(sdkDir).exists()) {
-            // try bitnami Jenkins VM:
-            sdkDir = "/usr/local/share/google/google-cloud-sdk";
-          }
+        if (path.endsWith("google-cloud-sdk" + File.separator + "bin")) {
+          possiblePaths.add(path);
```
loosebazooka: is this right? We need the base directory don't we?

path - bin?

meltsufin: Good catch!

meltsufin:
@loosebazooka PTAL again

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/sdk/PathResolver.java
```diff
@@ -25,39 +29,70 @@
...
-        sdkDir = System.getProperty("user.home") + "/google-cloud-sdk";
-        if (!new File(sdkDir).exists()) {
-          // try devshell VM:
-          sdkDir = "/google/google-cloud-sdk";
-          if (!new File(sdkDir).exists()) {
-            // try bitnami Jenkins VM:
-            sdkDir = "/usr/local/share/google/google-cloud-sdk";
-          }
+        if (path.endsWith("google-cloud-sdk" + File.separator + "bin")) {
+          possiblePaths.add(path.substring(0, path.length() - 4));
```
loosebazooka: is File.separator a string or character? would -4 actually work no matter what? should we do -3?

meltsufin: According to javadocs, it's a single characters. We don't need the trailing /. So, this should work.

loosebazooka:
:ship:

PR #78, opened 2016-05-17
Add a listener to give access to process

### LINKED ISSUE #77: Expose underlying Process to allow destroying it

meltsufin:
Fixes #77 

@loosebazooka @etanshaul

### DIFF
--- a/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
+++ b/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
@@ -32,8 +32,7 @@ public class DefaultProcessRunner implements ProcessRunner {
   private final List<ProcessOutputLineListener> stdOutLineListeners;
   private final List<ProcessOutputLineListener> stdErrLineListeners;
   private final ProcessExitListener exitListener;
-
-  private Process process;
+  private final ProcessStartListener startListener;
 
   private Map<String, String> environment;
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
```diff
@@ -77,19 +78,15 @@ public void run(String[] command) throws ProcessRunnerException {
       }
       processBuilder.command(makeOsSpecific(command));
 
-      synchronized (this) {
```
meltsufin: It's nice that we can get rid of this messiness.

etanshaul: I like this solution. It felt a bit restrictive before to only allow a single process at a time per sdk.

etanshaul:
lgtm

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
loosebazooka: Yeah maybe we should separate out the cloudsdk configuration and the cloud sdk execution. It seems a little dangerous to re-use an execution environment, no?

loosebazooka:
sure :ship: but we need to revisit this architecture at some point

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessRunner.java
meltsufin: In the build plugins, there isn't really any re-use, right? We create a new instance of CloudSdk every time. This could be a little less straight-forward in the IDE plugins, but still doable.

PR #81, opened 2016-05-23
Deploy restructuring idea, request for feedback

loosebazooka:
@meltsufin @patflynn @etanshaul 

Looking at a review of the concept more than the actual code for this.

Take a look at DeployExample.java to start

the structure of the composition is basically :

- GcloudAppDeploy (Type: Deploy)
  - [required] CloudSdk
  - [optional] CliProcessStarter (Type: ProcesStarter)
  - [optional] DeployConfigurationTranslator (Type: ConfigurationTranslator)

1. CloudSdk right now if from the existing code, not much will change, execution is taken out of it though.
1. CliProcessStarter can configure two things, the jvm shutdown hook (most only disabled when running the devappserver in background mode) and the process environment
1. DeployConfigurationTranslator translates a deploy configuration to a cli

Theres some notes in the source files too. You can ignore build errors, some methods I just made up.

Also some of this stuff might conflict with things I preferred in earlier code reviews (like: inheritted stdout/stderr in process builder), so, yeah, sorry.

### DIFF
--- a/src/main/java/com/google/cloud/tools/app/deploy/Deploy.java
+++ b/src/main/java/com/google/cloud/tools/app/deploy/Deploy.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2016 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.app.deploy;
+
+import com.google.cloud.tools.app.api.deploy.DeployConfiguration;
+import com.google.cloud.tools.app.deploy.process.OutputHandler;
+
+import java.util.concurrent.Future;
+
+// this is the API to deploy, it's assumed to be asynchronous and returns
+// a future to manipulate the process
+
+public interface Deploy {
+
+  Future<String> deploy(DeployConfiguration config);
+
+  // should this be part of the API? Getting non-result related output?
+  void setOutputHandler(OutputHandler handler);
+
+}
--- a/src/main/java/com/google/cloud/tools/app/deploy/Deploy.java
+++ b/src/main/java/com/google/cloud/tools/app/deploy/Deploy.java
@@ -17,18 +17,32 @@
 package com.google.cloud.tools.app.deploy;
 
 import com.google.cloud.tools.app.api.deploy.DeployConfiguration;
+import com.google.cloud.tools.app.deploy.gcloud.GcloudAppDeploy;
 import com.google.cloud.tools.app.deploy.process.OutputHandler;
+import com.google.cloud.tools.app.impl.cloudsdk.internal.sdk.CloudSdk;
 
 import java.util.concurrent.Future;
 
-// this is the API to deploy, it's assumed to be asynchronous and returns
-// a future to manipulate the process
+/**
+ * Created by appu on 5/24/16.
+ */
+public class Deploy {
 
-public interface Deploy {
+  // If there's a good way to configure the cloud SDK here instead of
+  // providing an object, we can explore that.
+  public static DeployRequestFactory newRequestFactory(CloudSdk sdk) {
+    return new GcloudAppDeploy.Builder(sdk);
+  }
 
-  Future<String> deploy(DeployConfiguration config);
+  private Deploy() {
+  }
 
-  // should this be part of the API? Getting non-result related output?
-  void setOutputHandler(OutputHandler handler);
+  public interface DeployRequestFactory {
+    DeployRequest newDeploymentRequest(DeployConfiguration config);
+  }
 
+  public interface DeployRequest {
+    DeployRequest setStatusUpdater(OutputHandler handler);
+    Future<DeployResult> deploy();
+  }
 }
--- a/src/main/java/com/google/cloud/tools/app/deploy/DeployExample.java
+++ b/src/main/java/com/google/cloud/tools/app/deploy/DeployExample.java
@@ -16,8 +16,11 @@
 
 package com.google.cloud.tools.app.deploy;
 
+import com.google.cloud.tools.app.api.deploy.DeployConfiguration;
 import com.google.cloud.tools.app.deploy.gcloud.GcloudAppDeploy;
 import com.google.cloud.tools.app.deploy.process.LoggingOutputHandler;
+import com.google.cloud.tools.app.deploy.process.NullOutputHandler;
+import com.google.cloud.tools.app.deploy.process.PrintStreamOutputHandler;
 import com.google.cloud.tools.app.impl.cloudsdk.internal.sdk.CloudSdk;
 import com.google.cloud.tools.app.impl.config.DefaultDeployConfiguration;
 
@@ -30,23 +33,29 @@ public class DeployExample {
 
   public static void howToUse() {
     // presumably the sdk object comes from somewhere else, but lets build it here
-    // you can also set enviroment on it
+    // you can also set some other things on it
     CloudSdk sdk = new CloudSdk.Builder().sdkPath(new File("some/path")).build();
 
-    // create the deploy object
-    Deploy deploy = GcloudAppDeploy.newBuilder(sdk).build();
+    // this thing doesn't exist, but should return a DefaultDeployConfiguration
+    DeployConfiguration config = new DeployConfigurationBuilder().whatever("asdf").build();
 
-    // set the output handler for non-result output
-    deploy.setOutputHandler(new LoggingOutputHandler("test", Level.INFO));
 
-    // start the deployment
-    Future<String> deployFuture = deploy.deploy(new DefaultDeployConfiguration());
+    // do the deploy
+    Future<DeployResult> deployFuture = Deploy
+        .newRequestFactory(sdk) // [required]
+        .newDeploymentRequest(config) // [required]
+        .setStatusUpdater(new LoggingOutputHandler("test", Level.INFO)) // [optional]
+        .setStatusUpdater(new NullOutputHandler()) // [optional]
+        .setStatusUpdater(new PrintStreamOutputHandler(System.err)) // [optional]
+        .deploy();
 
     try {
-      // get the result -- a waiting call
-      deployFuture.get();
+      // get the result -- a blocking call
+      DeployResult result = deployFuture.get();
+
       // or kill it!
       deployFuture.cancel(true);
+
       // or whatever, I don't know, anything a future can do
 
     } catch (InterruptedException | ExecutionException e) {
--- a/src/main/java/com/google/cloud/tools/app/deploy/DeployExample.java
+++ b/src/main/java/com/google/cloud/tools/app/deploy/DeployExample.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2016 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.app.deploy;
+
+import com.google.cloud.tools.app.deploy.gcloud.GcloudAppDeploy;
+import com.google.cloud.tools.app.deploy.process.LoggingOutputHandler;
+import com.google.cloud.tools.app.impl.cloudsdk.internal.sdk.CloudSdk;
+import com.google.cloud.tools.app.impl.config.DefaultDeployConfiguration;
+
+import java.io.File;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.logging.Level;
+
+public class DeployExample {
+
+  public static void howToUse() {
+    // presumably the sdk object comes from somewhere else, but lets build it here
+    // you can also set enviroment on it
+    CloudSdk sdk = new CloudSdk.Builder().sdkPath(new File("some/path")).build();
+
+    // create the deploy object
+    Deploy deploy = GcloudAppDeploy.newBuilder(sdk).build();
+
+    // set the output handler for non-result output
+    deploy.setOutputHandler(new LoggingOutputHandler("test", Level.INFO));
+
+    // start the deployment
+    Future<String> deployFuture = deploy.deploy(new DefaultDeployConfiguration());
+
+    try {
+      // get the result -- a waiting call
+      deployFuture.get();
+      // or kill it!
+      deployFuture.cancel(true);
+      // or whatever, I don't know, anything a future can do
+
+    } catch (InterruptedException | ExecutionException e) {
+      e.printStackTrace();
+    }
+  }
+}
--- a/src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudAppDeploy.java
+++ b/src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudAppDeploy.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2016 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.app.deploy.gcloud;
+
+import com.google.cloud.tools.app.api.AppEngineException;
+import com.google.cloud.tools.app.api.deploy.DeployConfiguration;
+import com.google.cloud.tools.app.deploy.Deploy;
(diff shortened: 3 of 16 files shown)


src/main/java/com/google/cloud/tools/app/deploy/DeployExample.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.logging.Level;
+
+public class DeployExample {
+
+  public static void howToUse() {
+    // presumably the sdk object comes from somewhere else, but lets build it here
+    // you can also set enviroment on it
+    CloudSdk sdk = new CloudSdk.Builder().sdkPath(new File("some/path")).build();
```
elharo: don't we have code to find this?

src/main/java/com/google/cloud/tools/app/deploy/Deploy.java
```diff
@@ -0,0 +1,34 @@
...
+
+import com.google.cloud.tools.app.api.deploy.DeployConfiguration;
+import com.google.cloud.tools.app.deploy.process.OutputHandler;
+
+import java.util.concurrent.Future;
+
+// this is the API to deploy, it's assumed to be asynchronous and returns
+// a future to manipulate the process
+
+public interface Deploy {
```
elharo: Type names usually aren't verbs,

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudAppDeploy.java
```diff
@@ -0,0 +1,108 @@
...
+import com.google.cloud.tools.app.deploy.process.ProcessStarter;
+import com.google.cloud.tools.app.deploy.process.StreamConsumer;
+import com.google.cloud.tools.app.impl.cloudsdk.internal.sdk.CloudSdk;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.concurrent.Future;
+import java.util.logging.Level;
+
+public class GcloudAppDeploy implements Deploy {
```
elharo: What does the interface buy us? Will there be a second implementation?

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudAppDeploy.java
```diff
@@ -0,0 +1,108 @@
...
+    this.configurationTranslator = configurationTranslator;
+  }
+
+  @Override
+  public Future<String> deploy(DeployConfiguration config) {
+    try {
+      String[] command = sdk.createAppCommand("deploy", configurationTranslator.translate(config));
+      Process process = processStarter.startProcess(command);
+
+      if (outputHandler == null) {
```
elharo: Why is this lazily initialized?

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudAppDeploy.java
```diff
@@ -0,0 +1,108 @@
...
+        outputHandler = new LoggingOutputHandler(GcloudAppDeploy.class.getName(), Level.INFO);
+      }
+      StreamConsumer.startNewConsumer(process.getErrorStream(), outputHandler);
+
+      return new GcloudFuture(process);
+
+    } catch(IOException e) {
```
elharo: Checked exceptions should not be converted to runtime exceptions. This operation cries out for a checked exception..

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudAppDeployTranslator.java
```diff
@@ -0,0 +1,57 @@
...
+    Preconditions.checkNotNull(config);
+    Preconditions.checkNotNull(config.getDeployables());
+    Preconditions.checkArgument(config.getDeployables().size() > 0);
+
+    List<String> arguments = Lists.newArrayList();
+    arguments.add("deploy");
+    for (File deployable : config.getDeployables()) {
+      if (!deployable.exists()) {
+        throw new IllegalArgumentException(
+            "Deployable " + deployable.toPath().toString() + " does not exist.");
```
elharo: toString() not needed

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudFuture.java
```diff
@@ -0,0 +1,93 @@
...
+      }
+    });
+  }
+
+  @Override
+  public boolean cancel(boolean mayInterruptIfRunning) {
+    if (!mayInterruptIfRunning) {
+      try {
+        process.exitValue();
+        return cancelled = false;
```
elharo: BUG: you want == here

Better yet, just !cancelled

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudFuture.java
```diff
@@ -0,0 +1,93 @@
...
+      } catch(IllegalThreadStateException e) {
+        // we can continue to interrupt now
+      }
+    }
+    process.destroy();
+    return cancelled = true;
```
elharo: same BUG

return cancelled

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudFuture.java
```diff
@@ -0,0 +1,93 @@
...
+    } catch(IllegalThreadStateException e) {
+      return false;
+    }
+  }
+
+  @Override
+  public String get() throws InterruptedException, ExecutionException {
+    int exitCode = process.waitFor();
+    if (exitCode != 0) {
+      throw new ExecutionException("Process exited with code : " + exitCode, new Throwable());
```
elharo: you shouldn't need a Throwable here

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudFuture.java
loosebazooka: The intention was to do a assignment + return. But it looks like that's confusing and a bad idea.

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudFuture.java
loosebazooka: Execution Exception is kinda weird, doesn't have just a (message) constructor.

src/main/java/com/google/cloud/tools/app/deploy/DeployExample.java
loosebazooka: Yeah it's in the sdk. Anyway, looking for a higher level concept review for now.

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudFuture.java
elharo: Try using null and if that doesn't work, maybe an Exception but *not* a Throwable. Per Effective Java, treat Throwable as an abstract class.

isusanin:
It looks like you're trying out several independent ideas here, like the use of futures and configuration translators. Do you want to try to introduce them separately by changing a slice of the existing codebase? I think it would be easier to discuss them in that context. Also, can you point out the issues with the current API that you're trying to address?
When you say, "less inheritancy", which classes are you referring to?

meltsufin:
It looks like you're trying out several independent ideas here, like the use of futures and configuration translators. Do you want to try to introduce them separately by changing a slice of the existing codebase? I think it would be easier to discuss them in that context. Also, can you point out the issues with the current API that you're trying to address?
When you say, "less inheritancy", which classes are you referring to?

etanshaul:
I am also a little confused about the goal. Looking through the current design I don't really find much "bad" inheritance. What I see is a basic strategy pattern with interfaces for the actions, and concrete CloudSdk implementations for them which compose config classes. 

I think the future idea could be good though - i.e. to solve things like @patflynn comment here: https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/674#discussion_r64275222

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudFuture.java
loosebazooka: Ah, makes sense.

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudAppDeploy.java
loosebazooka: Yeah maybe we need a checked and a runtime exception for the library.

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudAppDeploy.java
loosebazooka: Oh yeah, I don't know, perhaps I should just set it as the default.

loosebazooka:
So let me rename this, inheritency was an old commit message that I let pass through. 

1. I want the api to provide everything that the user needs... Previously it seemed like the user had to configure the ProcessRunner, etc and provide output processing and everything from the implementation class. Here deploy provides access to two things the user should worry about: 
  1. the deploy result
  2. any output of the deployer.

2. I also want to separate out the glcoud/appcfg/devappserver processes from the SDK configuration itself. It allows for better reuse of the SDK without having to complicate process management in the sdk object. It also seems to make more sense when passing the SDK object around.

3. ConfigurationTranslator is a little extra here, probably didn't need to be in this PR. We could just do translation in the Deployer implementation. The idea was that we could have the annotation based translator at some point, and we could just swap it out.

src/main/java/com/google/cloud/tools/app/deploy/DeployExample.java
```diff
@@ -0,0 +1,56 @@
...
+
+    // create the deploy object
+    Deploy deploy = GcloudAppDeploy.newBuilder(sdk).build();
+
+    // set the output handler for non-result output
+    deploy.setOutputHandler(new LoggingOutputHandler("test", Level.INFO));
```
elharo: What if we don't set this? Does the output just vanish? That may well be what we want. If there's critical information here, we need a better return type that encapsulates it.

src/main/java/com/google/cloud/tools/app/deploy/DeployExample.java
```diff
@@ -0,0 +1,56 @@
...
+
+    // start the deployment
+    Future<String> deployFuture = deploy.deploy(new DefaultDeployConfiguration());
+
+    try {
+      // get the result -- a waiting call
```
elharo: waiting --> blocking

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudAppDeploy.java
```diff
@@ -0,0 +1,108 @@
...
+
+public class GcloudAppDeploy implements Deploy {
+  private final CloudSdk sdk;
+  private final ConfigurationTranslator<DeployConfiguration> configurationTranslator;
+  private OutputHandler outputHandler;
+  private ProcessStarter processStarter;
+
+  private GcloudAppDeploy(CloudSdk sdk,
+      ConfigurationTranslator<DeployConfiguration> configurationTranslator,
+      ProcessStarter processStarter) {
```
elharo: why is ProcessStarter injected? it seems like an implementation detail.

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudAppDeploy.java
```diff
@@ -0,0 +1,108 @@
...
+    private ConfigurationTranslator<DeployConfiguration> configurationTranslator;
+
+    private Builder(CloudSdk sdk) {
+      this.sdk = sdk;
+      // getEnvironment doesn't exist
+      this.processStarter = CliProcessStarter.newBuilder().environment(sdk.getEnvironment()).build();
+      this.configurationTranslator = new GcloudAppDeployTranslator();
+    }
+
+    public Builder processStarter(ProcessStarter processStarter) {
```
elharo: This value doesn't seem to be used.

elharo: why inject this? Does the client need to configure it?

src/main/java/com/google/cloud/tools/app/deploy/Deploy.java
```diff
@@ -0,0 +1,34 @@
...
+
+  Future<String> deploy(DeployConfiguration config);
```
meltsufin: Why is the output of deploy a String?

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudAppDeploy.java
```diff
@@ -0,0 +1,108 @@
...
+
+  public static Builder newBuilder(CloudSdk sdk) {
+    // there's probably some work to get the sdk refactored as storage for path/environment and stuff
+    return new Builder(sdk);
+  }
+
+  // this will probably be similar for all commands, but I don't know how to
+  // abstract it out yet, or if I even should?
+  public static class Builder {
+    private CloudSdk sdk;
```
meltsufin: I think Builders can work well here. Maybe it can extend the CloudSdk.Builder instead?

src/main/java/com/google/cloud/tools/app/deploy/DeployExample.java
loosebazooka: I think the default is to log it as well. What do you mean critical information? Like if the ouput has different levels of output?

I was wondering if outputHandler might need to have a logLevel-esqe writer.

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudAppDeploy.java
loosebazooka: The could? For instance, the devappserver running in async mode would want to disable the shutdown hook. Although I think I did it wrong, I actually want to reverse that shutdown hook thing to be on by default and off otherwise.

src/main/java/com/google/cloud/tools/app/deploy/Deploy.java
loosebazooka: It was DeployResult but Deploy result was just a string so I took it out. But yeah it could a more meaningful thing

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudAppDeploy.java
loosebazooka: Or maybe I don't need this at all?

src/main/java/com/google/cloud/tools/app/deploy/DeployExample.java
```diff
@@ -0,0 +1,56 @@
...
+
+public class DeployExample {
+
+  public static void howToUse() {
+    // presumably the sdk object comes from somewhere else, but lets build it here
+    // you can also set enviroment on it
+    CloudSdk sdk = new CloudSdk.Builder().sdkPath(new File("some/path")).build();
+
+    // create the deploy object
+    Deploy deploy = GcloudAppDeploy.newBuilder(sdk).build();
```
patflynn: it could be nice to have a global builder factory so that they only need to override the SDK path once for all commands used. 

Conceivably all of the commands could implement something like StatusMessageEmitter or some such nonsense and the client could configure the global builder factory with a default StatusMessageHandler.  Not even sure if that's useful though.

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudFuture.java
```diff
@@ -0,0 +1,93 @@
...
+
+public class GcloudFuture implements Future<String> {
+
+  // does this need to be synchronized some how when storing the result?
+
+  private List<String> result = Lists.newArrayList();
+  private final Process process;
+  private boolean cancelled = false;
+
+  public GcloudFuture(Process process) {
```
patflynn: I haven't really thought about this much, but isn't the standard way to go about doing something like this is put the process calling code into a Callable and passing that into an executor service?

src/main/java/com/google/cloud/tools/app/deploy/process/LoggingOutputHandler.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class LoggingOutputHandler implements OutputHandler {
+
+  private final Logger log;
+  private final Level level;
+
+  public LoggingOutputHandler(String logName, Level logLevel) {
```
patflynn: I find having logging as an API concept a bit surprising.

src/main/java/com/google/cloud/tools/app/deploy/DeployExample.java
loosebazooka: OutputHandler is the StatusMessageHandler you're looking for?

I mean the sdk potentially only needs to be configured and built once and can be passed around through an SdkManager/SdkService (or something). Since this implementation expects it to be a configuration only class.

```java
public class SdkManager/Service {
  CloudSdk instance
  setSdk()
  getSdk()
}
````

A global factory like this?
````java
class SdkBuilderFactory {
  void setSdkPath();
  void setEnvironment();
  Builder getBuilder() {
    return instance/new builder...
  }
````
 I think in the SDK case specifically, factory on the builder won't provide us much as I don't expect different SDKs to be built for the same application (unless the user starts changing path, perhaps though we trigger an update to the manager there).

src/main/java/com/google/cloud/tools/app/deploy/process/LoggingOutputHandler.java
loosebazooka: Its not in the API specifically, but it's just a default implementation. Unless I'm using as a default somewhere?

src/main/java/com/google/cloud/tools/app/deploy/gcloud/GcloudFuture.java
loosebazooka: I guess we could potentially drop off the management of the Process to the ExecutorService instead of managing it ourselves with shutdown hooks, etc.

loosebazooka:
Tried to make things more discoverable here. Still have the cloudSdk object though.

See DeployExample.java

User Visible classes now : Deploy, DeployResult, DeployConfiguration and... CloudSdk...

Api needs CloudSdk if using the CloudSdk implementation (the only one for now)
We potentially have a DeployConfigurationBuilder that provides a DefaultDeployConfiguration to the user.

There's still a little bit of building. But at least a user can start from Deploy and move forward from there instead of having to guess the Implementation class and find it's builder.

meltsufin:
@loosebazooka see #87 which attempts to hide the CloudSdk object as well.

meltsufin:
@loosebazooka should we close this one?

PR #87, opened 2016-05-24
Experiment with using a Builder for deployment [EXPERIMENTAL]

meltsufin:
@loosebazooka @patflynn


meltsufin:
I added a usage example to parallel #81.

There is no direct interaction with the CloudSdk class at all here. The client only needs to know about DeploymentBuilder, Deployment, and DeploymentResult.

PR #89, opened 2016-05-25
Ensure docker can read artifacts in flexible deploys

### LINKED ISSUE #90: Ensure artifacts have the correct file permissions
After staging a flexible style project, we need to ensure the artifact is readable by everyone or the user launching it inside the docker wont be able to run it. Deploys fail to verify the vm is running and an 502 is presented when trying to hit the url.

loosebazooka:
Fixes #90

### DIFF
--- a/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
+++ b/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
@@ -20,9 +20,13 @@ import com.google.cloud.tools.app.api.AppEngineException;
 import com.google.cloud.tools.app.api.deploy.AppEngineFlexibleStaging;
 import com.google.cloud.tools.app.api.deploy.StageFlexibleConfiguration;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.Sets;
 
 import java.io.IOException;
 import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.attribute.PosixFilePermission;
+import java.util.Set;
 
 /**
  * Cloud SDK based implementation of {@link AppEngineFlexibleStaging}.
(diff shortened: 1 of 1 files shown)


loosebazooka:
@meltsufin

meltsufin:
LGTM

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
```diff
@@ -73,10 +77,20 @@ public void stageFlexible(StageFlexibleConfiguration config) throws AppEngineExc
...
-        Files.copy(config.getArtifact().toPath(),
-            config.getStagingDirectory().toPath()
-                .resolve(config.getArtifact().toPath().getFileName()),
-            REPLACE_EXISTING);
+        Path destination = config.getStagingDirectory().toPath()
+            .resolve(config.getArtifact().toPath().getFileName());
+        Files.copy(config.getArtifact().toPath(), destination, REPLACE_EXISTING);
+
+        // Update artifact permissions so docker can read it when deployed
+        if (!System.getProperty("os.name").contains("Windows")) {
```
patflynn: We're doing this differently in IJ: https://github.com/GoogleCloudPlatform/gcloud-intellij/blob/master/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineDeployAction.java#L116

If this is the 'right' way we should probably use it there too.

loosebazooka: I don't think there's a right way, but there's definitely a much cleaner way. Will update and test.

loosebazooka: Oh I didn't see it used File and not Path. Let me look into this a little.

loosebazooka: So it appears Path is recommended going forward. I'm not sure how this is supposed to happen in Windows... added #91 for a set of Windows related checks we need to do.

loosebazooka: Going to submit this for now, File looks way more convenient though... :\

PR #98, opened 2016-05-26
Add project to common configuration

### LINKED ISSUE #97: Project flag needs to be available for all commands
This parameter spans multiple commands. While it probably doesn't apply to _all_ commands, I think it should be available to all of them in the same way that `format` is.

etanshaul:
fixes #97

@loosebazooka, @meltsufin 

I only added it to a couple commands for now (stop, start).

### DIFF
--- a/src/main/java/com/google/cloud/tools/app/api/DefaultConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/app/api/DefaultConfiguration.java
@@ -14,18 +14,21 @@
  * limitations under the License.
  */
 
-package com.google.cloud.tools.app.api.versions;
-
-import com.google.cloud.tools.app.api.Configuration;
-
-import java.util.Collection;
+package com.google.cloud.tools.app.api;
 
 /**
- * Identifies a version or a set of versions for use in {@link AppEngineVersions}.
+ * Base implementation for common configuration elements.
  */
-public interface VersionsSelectionConfiguration extends Configuration {
+public abstract class DefaultConfiguration implements Configuration {
 
-  Collection<String> getVersions();
+  private String project;
 
-  String getService();
+  @Override
+  public String getProject() {
+    return project;
+  }
+
+  public void setProject(String project) {
+    this.project = project;
+  }
 }
--- a/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineVersions.java
+++ b/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineVersions.java
@@ -63,6 +63,7 @@ public class CloudSdkAppEngineVersions implements AppEngineVersions {
     arguments.add("start");
     arguments.addAll(configuration.getVersions());
     arguments.addAll(GcloudArgs.get("service", configuration.getService()));
+    arguments.addAll(GcloudArgs.get(configuration));
 
     execute(arguments);
   }
@@ -82,6 +83,7 @@ public class CloudSdkAppEngineVersions implements AppEngineVersions {
     arguments.add("stop");
     arguments.addAll(configuration.getVersions());
     arguments.addAll(GcloudArgs.get("service", configuration.getService()));
+    arguments.addAll(GcloudArgs.get(configuration));
 
     execute(arguments);
   }
--- a/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineVersions.java
+++ b/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineVersions.java
@@ -61,9 +61,7 @@ public class CloudSdkAppEngineVersions implements AppEngineVersions {
     List<String> arguments = new ArrayList<>();
     arguments.add("versions");
     arguments.add("start");
-    arguments.addAll(configuration.getVersions());
-    arguments.addAll(GcloudArgs.get("service", configuration.getService()));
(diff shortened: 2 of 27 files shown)


src/main/java/com/google/cloud/tools/app/impl/config/DefaultVersionsSelectionConfiguration.java
```diff
@@ -26,10 +26,20 @@
  */
 public class DefaultVersionsSelectionConfiguration implements VersionsSelectionConfiguration {
 
+  private String project;
   private Collection<String> versions;
   private String service;
 
   @Override
+  public String getProject() {
```
meltsufin: I think we should create an abstract class called DefaultConfiguration that will have the getProject/setProject, and then just extend it here. Otherwise we'll be duplicating these accessor methods a lot.

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineVersions.java
```diff
@@ -63,6 +63,7 @@ public void start(VersionsSelectionConfiguration configuration) throws AppEngine
     arguments.add("start");
     arguments.addAll(configuration.getVersions());
     arguments.addAll(GcloudArgs.get("service", configuration.getService()));
+    arguments.addAll(GcloudArgs.get("project", configuration.getProject()));
```
meltsufin: Maybe we can add a method GcloudArgs.get(Configuration config) that will translate all the common configuration? This would eliminate the duplication of this line, and possibly more in the future.

src/main/java/com/google/cloud/tools/app/impl/config/DefaultVersionsSelectionConfiguration.java
etanshaul: If we do this, then we can no longer code the CloudSdk action classes against configuration interfaces

etanshaul:
@meltsufin took your suggestions. Let me know if that's what you had in mind.

src/main/java/com/google/cloud/tools/app/api/DefaultConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.app.api;
+
+/**
+ * Base implementation for common configuration elements.
+ */
+public abstract class DefaultConfiguration implements Configuration {
```
meltsufin: I think this should go into the `com.google.cloud.tools.app.impl.config` package.

etanshaul: Oh right. I forgot to move it.

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/args/GcloudArgs.java
```diff
@@ -65,4 +62,13 @@
   public static List<String> get(Map<?, ?> keyValueMapping) {
     return Args.keyValues(keyValueMapping);
   }
+
+  /**
+   * @return list of args for the common arguments in {@link Configuration}.
+   */
+  public static List<String> get(Configuration configuration) {
```
meltsufin: Can you add a test for this method?

meltsufin:
This is pretty much what I had in mind. Thanks!

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/args/GcloudArgs.java
etanshaul: Added a test. Also swapped around the order for the assert params in some of the other tests in the class since I noticed the expected / actual values were in the wrong order.

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineVersions.java
```diff
@@ -82,6 +83,7 @@ public void stop(VersionsSelectionConfiguration configuration) throws AppEngineE
     arguments.add("stop");
     arguments.addAll(configuration.getVersions());
     arguments.addAll(GcloudArgs.get("service", configuration.getService()));
+    arguments.addAll(GcloudArgs.get(configuration));
```
meltsufin: What about `delete` and `list`?
I would actually factor out these 3 lines into a private method for reuse:
```Java
private addVersionsArguments(AppEngineVersions configuration, List<String> arguments) {
  arguments.addAll(configuration.getVersions());
  arguments.addAll(GcloudArgs.get("service", configuration.getService()));
  arguments.addAll(GcloudArgs.get(configuration));
}
```

etanshaul: done. will add it to remaining commands where it makes sense.

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/args/GcloudArgs.java
meltsufin: Great! Thanks!

etanshaul:
I added the common config to the remaining commands where it makes sense (for example, I left it out of dev-server, staging etc.). I also left it out of gen-config until we hear otherwise.

meltsufin:
LGTM

PR #100, opened 2016-06-03
Renames listener methods to use the onSomething naming convention

### LINKED ISSUE #99: listener interfaces should use on... naming convention for listening methods
i.e. 

class BlaListener {
   void bla(Strinb bla);
}

should be changed to
class BlaLIstener {
  void onBla(String bla);
}

patflynn:
Closes #99.

### DIFF
--- a/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessExitListener.java
+++ b/src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessExitListener.java
@@ -19,8 +19,10 @@ package com.google.cloud.tools.app.impl.cloudsdk.internal.process;
 /**
  * Default process exit listener that simply captures the exit code and makes it available with a
  * getter. Before the process exit code is captured, it's initialized to -1.
+ *
+ * <p>When used with {@link com.google.cloud.tools.app.impl.cloudsdk.internal.sdk.CloudSdk.Builder},
+ * this allows checking of the success or failure of an action execution.
  */
-// TODO: Justify existence. What is this class for?
 public class DefaultProcessExitListener implements ProcessExitListener {
   private int exitCode = -1;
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/ProcessOutputLineListener.java
```diff
@@ -17,9 +17,13 @@
 package com.google.cloud.tools.app.impl.cloudsdk.internal.process;
 
 /**
- * Process listener that can be used with {@link DefaultProcessRunner} to read the process std or
- * error output line by line.
+ * Process listener that can be used with {@link DefaultProcessRunner} to read the merged stdout
```
meltsufin: They're not merged. You have to register a separate listener for stdout and stderr.

patflynn: Ah I get it.  This is generic.  Will change docs.

BTW this looks like it's used as part of the public api, why is under impl....internal?

meltsufin: It really only belongs to the CloudSdk, which is "internal".

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/process/DefaultProcessExitListener.java
```diff
@@ -19,8 +19,10 @@
 /**
  * Default process exit listener that simply captures the exit code and makes it available with a
  * getter. Before the process exit code is captured, it's initialized to -1.
+ *
+ * <p>When used with {@link com.google.cloud.tools.app.impl.cloudsdk.internal.sdk.CloudSdk.Builder},
+ * this allows checking of the success or failure of an action execution.
```
patflynn: Do you really save anything from using this implementation vs using an anonymous implementation that reads out the value?

meltsufin: I'm not really sure, but you save a few lines of code and the cost of the anonymous classes.

patflynn: I'd get rid of this until you feel real pain from the lack of it.  With
Java 8 it will be even less necessary.

patflynn:
Is this good to go?

meltsufin:
:ship:

Issue #101, opened 2016-06-03
revisit package structure so that public API is clearer


patflynn:
In gcloud-intellij we are using some classes under impl/.../internal/.. which seems wrong. All classes that should be used by clients should be in a package that makes that clear.

PR #103, opened 2016-06-03
[OSGi] Export .impl.* packages but marked as internal

briandealwis:
I'm now starting to integrate `app-tools-lib-for-java` for the `GoogleCloudPlatform/gcloud-eclipse-tools` tooling.  It looks like need to access classes in `.impl`, particularly the `CloudSdk` class at the very least.  I've changed up the OSGi metadata to export the `.impl` packages but in such a way that they're flagged as being internal classes, which produces a warning in Eclipse, using the same setup used by Eclipse-provided bundles.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -108,6 +108,10 @@
           <instructions>
             <Bundle-SymbolicName>com.google.cloud.tools.app.lib</Bundle-SymbolicName>
             <Bundle-RequiredExecutionEnvironment>JavaSE-1.7</Bundle-RequiredExecutionEnvironment>
+            <Export-Package>
+              com.google.cloud.tools.app.impl.*;x-internal:=true;-noimport:=true,
+              com.google.cloud.tools.app.*
+            </Export-Package>
           </instructions>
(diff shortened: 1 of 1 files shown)


meltsufin:
Can you explain why `.impl` would be marked as internal? Would you prefer those classes to be in a different package?

We have #101 open to address the package structure. Maybe this should be part of it.

meltsufin:
@briandealwis Do you want to revisit this since #105 has been merged?

briandealwis:
@meltsufin I'll take a look and revisit

elharo:
The Felix maven plugin we use to generate OSGI bundles marks packages named impl or internal as private by default:

http://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html#default-behavior

meltsufin:
With the latest changes, you shouldn't need to use any impl or internal packages.

briandealwis:
Closing as no longer necessary.  Thanks guys,

PR #105, opened 2016-06-06
reorg packages so that .internal is not part of the API

### LINKED ISSUE #101: revisit package structure so that public API is clearer

meltsufin:
Fixes #101. This ensures that the clients of the library do not need to import anything from .internal or .impl. It also adds a factory that allows a single point from which to discover and use all of  the library functionality.

* created an AppEngineFactory and a CloudSDK implementation for it.
* moved default configuration implementations into the .api apackage.

@loosebazooka @patflynn


src/main/java/com/google/cloud/tools/app/api/AppEngineFactory.java
```diff
@@ -0,0 +1,50 @@
...
+import com.google.cloud.tools.app.api.genconfig.GenConfigUtility;
+import com.google.cloud.tools.app.api.instances.AppEngineInstances;
+import com.google.cloud.tools.app.api.logs.AppEngineLogs;
+import com.google.cloud.tools.app.api.services.AppEngineServices;
+import com.google.cloud.tools.app.api.versions.AppEngineVersions;
+
+/**
+ * A factory interface for creating App Engine services.
+ */
+public interface AppEngineFactory {
```
loosebazooka: I'd rather not include this in this change. We should really discuss how we want to expose the api through a design doc and review.

meltsufin: I agree that these changes are expanding the scope of the original issue, but the reason for including the factory in this PR is to be able to keep the `CloudSdk` and all `CloudSdkAppEngine*` implementation classes internal. We can review the idea of exposing the functionality to clients through this factory offline.

loosebazooka: I'd like to hold off then, prefer not expanding the scope till we decide what we want. The package naming doesn't seem like a pressing issue.

briandealwis:
I'm only starting to play with the APIs, but using a Builder to create a Factory to create command objects that take configurations feels convoluted.

src/main/java/com/google/cloud/tools/app/api/DefaultConfiguration.java
```diff
@@ -14,9 +14,7 @@
  * limitations under the License.
  */
 
-package com.google.cloud.tools.app.impl.config;
-
-import com.google.cloud.tools.app.api.Configuration;
+package com.google.cloud.tools.app.api;
```
elharo: as long as we're renaming everything what is "app" in the package name? It seems very generic. Is this really "appengine"?

meltsufin: It does refer to appengine. The source of the abbreviation it is really the `gcloud preview app ...` command.

patflynn: @meltsufin 

I think we made a mistake making this the same name as the gcloud 'app' command. gcloud is a cli that should be minimally verbose,  I think with package names in java we should be as descriptive as we can.   What do you think of changing from app to appengine?  This will also match the package naming in other tools better.

meltsufin: I don't mind renaming app -> appengine, but in order to be consistent, wouldn't we also want to rename the library itself to appengine-tools-lib-for-java, not to mention the plugins?

src/main/java/com/google/cloud/tools/app/cloudsdk/CloudSdkAppEngineFactory.java
```diff
@@ -0,0 +1,216 @@
...
+      cloudSdkBuilder.appCommandCredentialFile(appCommandCredentialFile);
+      return this;
+    }
+
+    /**
+     * Sets the format for printing command output resources. The default is a command-specific
+     * human-friendly output format. The supported formats are: csv, default, flattened, json, list,
+     * multi, none, table, text, value, yaml. For more details run $ gcloud topic formats.
+     */
+    public Builder appCommandOutputFormat(String appCommandOutputFormat) {
```
patflynn: shouldn't the parameter to this be an enum?

src/main/java/com/google/cloud/tools/app/cloudsdk/CloudSdkAppEngineFactory.java
```diff
@@ -0,0 +1,216 @@
...
+    public Builder async(boolean async) {
+      cloudSdkBuilder.async(async);
+      return this;
+    }
+
+    /**
+     * Adds a client consumer of process standard output. If none, output will be inherited by
+     * parent process.
+     */
+    public Builder addStdOutLineListener(ProcessOutputLineListener stdOutLineListener) {
```
patflynn: for future reference, this kind of config for IDE usecases probably needs to be part of the specific service invocation, and not at the 'SDK' level.

src/main/java/com/google/cloud/tools/app/cloudsdk/CloudSdkAppEngineFactory.java
```diff
@@ -0,0 +1,216 @@
...
+      cloudSdkBuilder.addStdOutLineListener(stdOutLineListener);
+      return this;
+    }
+
+    /**
+     * Adds a client consumer of process error output. If none, output will be inherited by parent
+     * process.
+     */
+    public Builder addStdErrLineListener(ProcessOutputLineListener stdErrLineListener) {
```
patflynn: Ditto for all these listeners.

src/main/java/com/google/cloud/tools/app/cloudsdk/CloudSdkAppEngineFactory.java
meltsufin: Sure, but we'll then need to sync up with gcloud, if they ever add or remove any of the formats.

src/main/java/com/google/cloud/tools/app/cloudsdk/CloudSdkAppEngineFactory.java
```diff
@@ -0,0 +1,216 @@
...
+public class CloudSdkAppEngineFactory implements AppEngineFactory {
+
+  private CloudSdk cloudSdk;
+
+  private CloudSdkAppEngineFactory(CloudSdk cloudSdk) {
+    this.cloudSdk = cloudSdk;
+  }
+
+  @Override
+  public AppEngineDeployment deployment() {
```
patflynn: 'factory' methods  should be named in such a way as it's clear that it's creating something.  I also think the naming of the commands is a bit weird.  They sound like things rather than actions.

But more broadly, I do agree with @briandealwis and @loosebazooka that we may be struggling with getting the right abstractions and names here.

meltsufin: The methods can be renamed to `create...`, but in general, should we keep the factory or revert to exposing the `CloudSdk` along with the `CloudSdkAppEngine....(CloudSdk)` classes instead?

src/main/java/com/google/cloud/tools/app/api/DefaultConfiguration.java
patflynn: i agree that the actual artifact name of this library needs work, but I
don't think that has to be tied to this PR.

meltsufin: Should we then hold off renaming the `app` package as well?

src/main/java/com/google/cloud/tools/app/cloudsdk/CloudSdkAppEngineFactory.java
patflynn: If you think that there's value in exposing the new package structure
independently of this Factory change, I'd remove it and go ahead with the
package refactor.  Otherwise I'd work with Appu on a design that you can
agree on.

src/main/java/com/google/cloud/tools/app/api/DefaultConfiguration.java
patflynn: Not on my account. I'm fine with you changing it now.

src/main/java/com/google/cloud/tools/app/cloudsdk/CloudSdkAppEngineFactory.java
meltsufin: this actually reminds me that we should remove the `add` prefix and use varargs instead

src/main/java/com/google/cloud/tools/app/api/DefaultConfiguration.java
meltsufin: @loosebazooka What do you think of the package renaming to `appengine`?

loosebazooka: Yeah that's cool.

src/main/java/com/google/cloud/tools/app/cloudsdk/CloudSdkAppEngineFactory.java
meltsufin: Created a separate issue: #107

loosebazooka:
:ship: lets also make sure all consumers are aware when we push to snapshots.

PR #106, opened 2016-06-06
Add directory copy for docker flexible staging

loosebazooka:
The pkg might be weird, but otherwise trying to enable us to copy directories without commons-io. Only enabled the change on the flexible staging step


loosebazooka:
@patflynn @meltsufin @etanshaul

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
```diff
@@ -57,6 +58,12 @@ public void stageFlexible(StageFlexibleConfiguration config) throws AppEngineExc
 
     try {
 
+      // Copy docker context to staging
+      if (config.getDockerDirectory() != null && config.getDockerDirectory().exists()) {
```
meltsufin: Should we check if the directory contains Dockerfile? Otherwise, it's probably a user error.

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/internal/util/FileUtil.java
```diff
@@ -0,0 +1,110 @@
...
+ */
+public class FileUtil {
+
+  /**
+   * We really want to preserve file attributes and permissions on linux and windows.
+   *
+   * @param source an existing source directory to copy from
+   * @param destination an existing destination directory to copy to
+   */
+  public static void copyDirectory(final Path source, final Path destination) throws IOException {
```
meltsufin: Have you looked at commons-io FileUtils.copyDirectory()? Any reason not to use it?

loosebazooka: Well partially because we don't want to include that dependency. Secondly because I don't think it preserves attributes/permissions, but that might have changed, going to have another look at their source.

meltsufin: Sorry, missed what you said in the main comment, but yeah, let's make sure commons-io is not doing what we want before providing our own impl.

loosebazooka: So as far as I can tell, it only attempts to preserve file modified time : http://grepcode.com/file/repo1.maven.org/maven2/commons-io/commons-io/2.4/org/apache/commons/io/FileUtils.java#FileUtils.copyDirectory%28java.io.File%2Cjava.io.File%2Cboolean%29

meltsufin: What about Java 7 `Files.copy(source, target, StandardCopyOption.COPY_ATTRIBUTES)`?

loosebazooka: Yeah I was just testing that out. Turns out COPY_ATTRIBUTES does copy file permissions, contrary to what some clown on the internet said. Anyway, we still have to implement directory copy.

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
loosebazooka: sure.

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
```diff
@@ -49,17 +49,21 @@ public void stageFlexible(StageFlexibleConfiguration config) throws AppEngineExc
...
       throw new AppEngineException("Staging location is not a directory. Location: "
-          + config.getStagingDirectory().toPath().toString());
+          + config.getStagingDirectory().toPath());
     }
 
     try {
 
       // Copy docker context to staging
       if (config.getDockerDirectory() != null && config.getDockerDirectory().exists()) {
+        if (!Files.isRegularFile(config.getDockerDirectory().toPath().resolve("dockerfile"))) {
```
meltsufin: Is it supposed to be uppercase "Dockerfile"?

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
```diff
@@ -48,15 +49,25 @@ public void stageFlexible(StageFlexibleConfiguration config) throws AppEngineExc
...
       throw new AppEngineException("Staging location is not a directory. Location: "
-          + config.getStagingDirectory().toPath().toString());
+          + config.getStagingDirectory().toPath());
     }
 
     try {
 
+      // Copy docker context to staging
+      if (config.getDockerDirectory() != null && config.getDockerDirectory().exists()) {
+        if (!Files.isRegularFile(config.getDockerDirectory().toPath().resolve("dockerfile"))) {
```
patflynn: case sensitivity is going to be an issue here too.  The convention is that the file is called 'Dockerfile'

src/main/java/com/google/cloud/tools/app/impl/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
loosebazooka: good catch.

src/test/java/com/google/cloud/tools/app/impl/cloudsdk/impl/util/FileUtilTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+    FileUtil.copyDirectory(src, dest);
+
+    Assert.assertTrue(Files.isRegularFile(dest.resolve(src.relativize(rootFile))));
+    Assert.assertTrue(Files.isDirectory(dest.resolve(src.relativize(subDir))));
+    Assert.assertTrue(Files.isRegularFile(dest.resolve(src.relativize(subFile))));
+  }
+
+  @Test
+  public void testCopyDirectory_posixPermissions() throws IOException {
```
meltsufin: will this test fail on Windows?

src/test/java/com/google/cloud/tools/app/impl/cloudsdk/impl/util/FileUtilTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+    FileUtil.copyDirectory(src, dest);
+
+    Assert.assertEquals(permission,
+        Files.getPosixFilePermissions(dest.resolve(src.relativize(rootFile))));
+  }
+
+  @Test
+  public void testCopyDirectory_aclPermissions() throws IOException {
+    assumeTrue(System.getProperty("os.name").startsWith("Windows"));
```
meltsufin: OS-specific tests probably don't belong in a unit test.

src/test/java/com/google/cloud/tools/app/impl/cloudsdk/impl/util/FileUtilTest.java
loosebazooka: no it's skipped on windows -> see assumeTrue

src/test/java/com/google/cloud/tools/app/impl/cloudsdk/impl/util/FileUtilTest.java
loosebazooka: maybe, any ideas where this should go then?

meltsufin: I think I would just try to rewrite this test using mocks so that it's not OS-specific.

loosebazooka: I mean I'm specifically confirming that linux permissions are preserved. Don't think I can mock in this test.

meltsufin: Then you're not really testing the library code, but rather the JDK.

loosebazooka: Sure, the idea was that we specifically needed permissions to be preserved and this satisfied our use case, however the test also ensures that the copy-options are set correctly.

meltsufin: Can you verify copy options being set by calling copyDirectory() with mocks?

loosebazooka: yeah, that's a good point.

loosebazooka: The problem here lies in the fact that it's a static utility method and I'm testing it as a black box.

I can either take it out or leave it in (and I'm leaning towards leaving it in). Mocking isn't a option here given the style, making it a singleton without state doesn't feel right either.

meltsufin: I see the difficulty with mocking here. Let's just leave it. It's better than not having the test at all.

meltsufin:
:ship:

PR #116, opened 2016-06-09
export cloudsdk package explicitly

elharo:
Not yet 100% sure if we need this; checking...


meltsufin:
Do you need this in the Sonatype snapshots repo to test?

elharo:
don't think this is actually needed

PR #119, opened 2016-06-15
Restore CloudSdk.validate()

### LINKED ISSUE #118: Restore CloudSdk#validate() to check a location
`CloudSdk` had a `validate()` method to check whether the location seemed like a real installation.  This check is useful for handling user-configured locations.

meltsufin:
Fixes #118.

@loosebazooka @briandealwis PTAL

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -193,6 +193,38 @@ public class CloudSdk {
     return getJavaAppEngineSdkPath().resolve(JAVA_TOOLS_JAR);
   }
 
+  /**
+   * Checks whether the configured Cloud SDK Path is valid.
+   *
+   * @throws AppEngineException when there is a validation error.
+   */
+  public void validate() throws AppEngineException {
+    if (sdkPath == null) {
+      throw new AppEngineException("Validation Error: Sdk path is null");
+    }
+    if (sdkPath.toFile().isDirectory()) {
+      throw new AppEngineException(
+          "Validation Error: Sdk directory '" + sdkPath + "' is not valid");
+    }
+    if (getGCloudPath().toFile().isFile()) {
+      throw new AppEngineException(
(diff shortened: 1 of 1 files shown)


loosebazooka:
where did it go?

meltsufin:
I removed it previously, since the comment on it said "no one should be using it". Sorry :-(

briandealwis:
So the only issue I have is that the _reason_ strings passed into `AppEngineException` aren't great.  It's useful to report the problems found so those who encounter a problem  (e.g., a new version of gcloud is released that changes the organization) can patch it up sufficiently to pass.

meltsufin:
You can get the message from the exception. Is that not sufficient for your use-case?

briandealwis:
They're not translatable, but they're enough to report on logging.  I took a quick stab at making them a bit more reportable and it's difficult (_expected 'bin/cloud.jar' to be a file_). So LGTM.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -193,6 +193,38 @@ private Path getJavaToolsJar() {
...
+  /**
+   * Checks whether the configured Cloud SDK Path is valid.
+   *
+   * @throws AppEngineException when there is a validation error.
+   */
+  public void validate() throws AppEngineException {
+    if (sdkPath == null) {
+      throw new AppEngineException("Validation Error: SDK path is null");
+    }
+    if (sdkPath.toFile().isDirectory()) {
```
briandealwis: These test conditions are inverted: missing a '!'

meltsufin: :scream: how did this slip in! great catch!

briandealwis:
I put two fixes in https://github.com/briandealwis/app-tools-lib-for-java/tree/cloudsdk_validate

meltsufin:
@briandealwis Integrated your fixes. PTAL

briandealwis:
LGTM

PR #127, opened 2016-06-20
Copy all App Engine supported files for staging

### LINKED ISSUE #92: Copy all files from src/main/docker for staging
Not only the Dockerfile from src/main/docker should be copied, but all the files and subdir content if any, as these files can serve as context files for the docker image build step.

### LINKED ISSUE #96: Copy all App Engine supported files from src/main/appengine for staging
...Not just app.yaml

meltsufin:
Fixes #96. 

The clients now have to specify an App Engine config directory instead of an app.yaml path for flexible staging.


meltsufin:
@loosebazooka @patflynn

loosebazooka:
This seems okay, just wondering if the swagger files are standard or not, maybe @patflynn can answer that?

loosebazooka:
:ship: if we need to remove the swagger files later we can.

PR #128, opened 2016-06-20
Yaml search through deployable directory

loosebazooka:
@meltsufin open to suggestions on different ways to do this, basically if we get a directory, then look through it for yamls

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDeployment.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDeployment.java
@@ -61,8 +61,7 @@ public class CloudSdkAppEngineDeployment implements AppEngineDeployment {
           arguments.add(yamlFile.toPath().toString());
         }
       }
-    }
-    else {
+    } else {
       for (File deployable : config.getDeployables()) {
         if (!deployable.exists()) {
           throw new IllegalArgumentException(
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDeployment.java
```diff
@@ -48,12 +53,22 @@ public void deploy(DeployConfiguration config) throws AppEngineException {
...
     List<String> arguments = new ArrayList<>();
     arguments.add("deploy");
-    for (File deployable : config.getDeployables()) {
-      if (!deployable.exists()) {
-        throw new IllegalArgumentException(
-            "Deployable " + deployable.toPath().toString() + " does not exist.");
+    // If we get a single directory, then look in it for yamls
+    if (config.getDeployables().size() == 1 && config.getDeployables().get(0).isDirectory()) {
+      for (String filename : yamlFilenames) {
+        File yamlFile = new File(config.getDeployables().get(0), filename);
```
meltsufin: I would put `config.getDeployables().get(0)` into a local variable so you don't have to repeat it.

meltsufin:
I like that this logic is in the common lib, but I'm unsure about allowing a directory to be passed in as a "deployable" and that having a special meaning.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDeployment.java
```diff
@@ -24,14 +24,20 @@
...
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 /**
  * Cloud SDK based implementation of {@link AppEngineDeployment}.
  */
 public class CloudSdkAppEngineDeployment implements AppEngineDeployment {
 
+  private static final List<String> yamlFilenames = Arrays
```
meltsufin: Can you make the variable all upper case? Like APPENGINE_YAML_FILES?

loosebazooka: good call.

meltsufin:
:ship:

PR #140, opened 2016-06-22
Prepare pom.xml for release to Maven Central

### LINKED ISSUE #139: Prepare pom.xml for release to Maven Central
Must meet these requirements: http://central.sonatype.org/pages/requirements.html

meltsufin:
Fixes #139 .
@patflynn PTAL

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2,14 +2,12 @@
   <modelVersion>4.0.0</modelVersion>
 
   <groupId>com.google.cloud.tools</groupId>
-  <artifactId>app-tools-lib-for-java</artifactId>
+  <artifactId>appengine-plugins-core</artifactId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -82,6 +130,23 @@
       </plugin>
 
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
```
patflynn: what about attaching sources? Is that automatic?

meltsufin: We were doing it before https://github.com/GoogleCloudPlatform/app-tools-lib-for-java/blob/985c162d0dc472d273ace1aff5ea47fcd9a79477/pom.xml#L124

patflynn:
LGTM

PR #141, opened 2016-06-23
Bump current version to 0.2.0-SNAPSHOT

meltsufin:
...following the 0.1.0 release.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
 
   <groupId>com.google.cloud.tools</groupId>
   <artifactId>appengine-plugins-core</artifactId>
-  <version>0.1.0-SNAPSHOT</version>
+  <version>0.2.0-SNAPSHOT</version>
 
   <name>App Engine Plugins Core Library</name>
   <description>


meltsufin:
I wanted to reserve 0.1.x for emergency bug fixes that don't have breaking API changes.

patflynn:
wouldn't this scheme limit us to 10 released API breaking revisions before
1.0?  One level deeper would give us more breathing room.

meltsufin:
0.999.0 is still cool I think :)

patflynn:
ah fair enough. Personally, I would like us to stay in the very lows <1 for
as long as possible to make it very clear to potential users that this is
not ready and very unstable.

meltsufin:
Technically, with semantic versioning, we should be bumping the major version every time we make a breaking API change. However, since we want to stay below version 1.0.0, bumping the minor version instead seems like a reasonable compromise.

patflynn:
a version of 0.9 means that something is close to 1.0 in people's minds.  I
don't want to get there any time soon.

meltsufin:
Closing in favor of #142

PR #145, opened 2016-06-23
remove unused field

elharo:
@meltsufin


meltsufin:
I see, the parameter is only used in the constructor. LGTM.

PR #146, opened 2016-06-24
update --> install

elharo:
@meltsufin 

elharo@elharo:~$ gcloud components update app-engine-java
You have specified individual components to update.  If you are trying
 to install new components, use:
  $ gcloud components install app-engine-java

Do you want to run install instead (y/N)?  y


loosebazooka:
CLA problem closed in favor of #157

PR #150, opened 2016-06-24
Fixes #149 - PathResolver should implement an interface to enable mocking

### LINKED ISSUE #149: PathResolver should implement an interface to enable mocking
PathResolver is an enum and as such cannot be mocked, but it would be helpful to write unit tests that are independent of the executing environment.

akerekes:


elharo:
I'm still not convinced this is the right approach. Let's talk on Monday.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/IPathResolver.java
```diff
@@ -0,0 +1,14 @@
+package com.google.cloud.tools.appengine.cloudsdk;
+
+import java.nio.file.Path;
+
+public interface IPathResolver {
```
meltsufin: I would rename to plain PathResolver

src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
```diff
@@ -24,7 +24,7 @@
 /**
  * Resolve paths with CloudSdk and Python defaults.
  */
-public enum PathResolver {
+public enum PathResolver implements IPathResolver {
```
meltsufin: I think we should name it DefaultPathResolver and get rid of enum/singleton. CloudSdk will need to be modified to accept a PathResolver in the CloudSdk.Builder.

meltsufin:
@akerekes @elharo Any interest in this still?

akerekes:
@meltsufin yes, we're still interested in this, but not sure whether this approach is the best. If we'll have an interface, the singleton constraint will be much more difficult to maintain, so we could just change it to a regular class (and not make it a singleton).

meltsufin:
I don't think it needs to be a singleton at the API level. This class doesn't have any state. So, it can either be a static utility method, or an interface and a default implementation.  I prefer the latter, but making it a static method would not make it worse than it is now.

akerekes:
@elharo @meltsufin PTAL

elharo:
I'm not sure why we need the interface here. Just rename DefaultPathResolver to PathResolver, and we're good. It's no longer an unmockable singleton, which was the concern.

elharo:
LGTM

meltsufin:
:+1:

Issue #85, opened 2016-05-24
Use Varargs  for CloudSdk.Builder.add*


meltsufin:

elharo:
Obsolete. There are no longer any such methods.

PR #154, opened 2016-07-07
Make dev_appserver exit in asnyc on failure

loosebazooka:
- adds exitListener/startListener lists

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -68,7 +68,8 @@ public class CloudSdk {
                    boolean async,
                    List<ProcessOutputLineListener> stdOutLineListeners,
                    List<ProcessOutputLineListener> stdErrLineListeners,
-                   ProcessExitListener exitListener, ProcessStartListener startListener,
+                   List<ProcessExitListener> exitListeners,
+                   List<ProcessStartListener> startListeners,
                    int runDevAppServerWaitSeconds) {
     this.sdkPath = sdkPath;
     this.appCommandMetricsEnvironment = appCommandMetricsEnvironment;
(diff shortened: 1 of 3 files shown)


loosebazooka:
@meltsufin this doesn't expose the lists of exit/start listeners to the user

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -323,15 +325,15 @@ public Builder addStdErrLineListener(ProcessOutputLineListener stdErrLineListene
      * The client listener of the process exit with code.
      */
     public Builder exitListener(ProcessExitListener exitListener) {
```
meltsufin: Let's just make it varargs, and fix #85 at the same time.

loosebazooka: I was trying to do this change without changing the surface of the builder.

loosebazooka: potentially we would want this to be addExitListeners or whatever if we were exposing it as a list

meltsufin: I think we would make it plural, but no 'add' prefix. Just builder.exitListners(...)

In any case, if you want to keep the surface the same, you should first clear the exitListeners list in this method. Otherwise, the method no longer works like a setter.

loosebazooka: yeah I had the clear in there, but it looked dumb. Let me put it back in, we can change the surface later.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/DefaultProcessRunner.java
```diff
@@ -47,17 +45,17 @@
    *                            be inherited by parent process.
    * @param stdErrLineListeners Client consumers of process error output. If empty, output will be
    *                            inherited by parent process.
-   * @param exitListener        Client consumer of process onExit event.
+   * @param exitListeners        Client consumer of process onExit event.
```
meltsufin: consumers with plural now

meltsufin:
:ship:

PR #160, opened 2016-07-11
Add support for java8 in standard staging.

loosebazooka:
@meltsufin


meltsufin:
`useRuntimeJava` seems a little confusing. Maybe we should just allow the clients to set `javaVersion`, and the library will figure out what to do with that.

meltsufin:
@loosebazooka changed configuration to allow specifying any runtime instead of just boolean for java.

loosebazooka:
:runner:

PR #161, opened 2016-07-13
If runtime is java, set runtime config to openjdk8 in app.yaml

loosebazooka:
This is a change that we should push up to appcfg/staging or into
gcloud. No tool at the moment appears to be building a dockerfile
with runtime:java and vm:true that is correct. By forcing the jdk
to be openjdk8 we trigger the creation of a Dockerfile with the
correct java8jetty9 image.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineStandardStaging.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineStandardStaging.java
@@ -21,9 +21,12 @@ import com.google.cloud.tools.appengine.api.deploy.AppEngineStandardStaging;
 import com.google.cloud.tools.appengine.api.deploy.StageStandardConfiguration;
 import com.google.cloud.tools.appengine.cloudsdk.internal.args.AppCfgArgs;
 import com.google.cloud.tools.appengine.cloudsdk.internal.process.ProcessRunnerException;
+import com.google.common.base.Charsets;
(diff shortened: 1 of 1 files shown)


meltsufin:
:runner:

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineStandardStaging.java
```diff
@@ -80,6 +83,13 @@ public void stageStandard(StageStandardConfiguration config) throws AppEngineExc
 
       cloudSdk.runAppCfgCommand(arguments);
 
+      //TODO : Move this fix up the chain (appcfg)
+      if (config.getRuntime() != null && config.getRuntime().equals("java")) {
```
elharo: This continues to confuse me. I thought App Engine Standard Java did not use app.yaml?

meltsufin: Staging generates app.yaml from the appengine-web.xml, which is then used for deployment.

elharo: Where is this documented?

meltsufin: Good question, but I don't know. In any case, this is more of an implementation detail of staging and deployment in AppCfg and gcloud.

PR #164, opened 2016-07-13
Make getSdkPath and getJavaToolsJar public.

joaoandremartins:
getSdkPath should only be used for testing in https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/335

@meltsufin @loosebazooka PTAL

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -32,6 +32,7 @@ import java.io.File;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.logging.Logger;
@@ -52,6 +53,7 @@ public class CloudSdk {
       "platform/google_appengine/google/appengine/tools/java/lib";
   private static final String JAVA_TOOLS_JAR = "appengine-tools-api.jar";
   private static final String WINDOWS_BUNDLED_PYTHON = "platform/bundledpython/python.exe";
+  private static final Map<String, Path> JAR_LOCATIONS = new HashMap<>();
 
   private final Path sdkPath;
   private final ProcessRunner processRunner;
@@ -93,6 +95,12 @@ public class CloudSdk {
     this.processRunner = new DefaultProcessRunner(async, stdOutLineListeners, stdErrLineListeners,
         exitListeners, startListeners);
 
+    // Populate jar locations.
+    JAR_LOCATIONS.put("servlet-api.jar",
+            getJavaAppEngineSdkPath().resolve("shared/servlet-api.jar"));
+    JAR_LOCATIONS.put("jsp-api.jar", getJavaAppEngineSdkPath().resolve("shared/jsp-api.jar"));
+    JAR_LOCATIONS.put(JAVA_TOOLS_JAR,
+        sdkPath.resolve(JAVA_APPENGINE_SDK_PATH).resolve(JAVA_TOOLS_JAR));
   }
 
   /**
@@ -162,7 +170,7 @@ public class CloudSdk {
     command.add(
         Paths.get(System.getProperty("java.home")).resolve("bin/java").toAbsolutePath().toString());
     command.add("-cp");
-    command.add(getJavaToolsJar().toAbsolutePath().toString());
+    command.add(JAR_LOCATIONS.get(JAVA_TOOLS_JAR).toString());
     command.add("com.google.appengine.tools.admin.AppCfg");
     command.addAll(args);
 
@@ -195,12 +203,18 @@ public class CloudSdk {
     return getSdkPath().resolve(JAVA_APPENGINE_SDK_PATH);
   }
 
-  public Path getJavaToolsJar() {
-    return getJavaAppEngineSdkPath().resolve(JAVA_TOOLS_JAR);
-  }
-
   private Path getWindowsPythonPath() {
     return getSdkPath().resolve(WINDOWS_BUNDLED_PYTHON);
+  }
+
+  /**
+   * Gets the file system location for an SDK jar.
+   *
+   * @param jarName the jar file name. For example, "servlet-api.jar"
+   * @return the path in the file system
+     */
+  public Path getJarPath(String jarName) {
+    return JAR_LOCATIONS.get(jarName);
   }
 
   /**
@@ -229,9 +243,10 @@ public class CloudSdk {
           "Validation Error: Java App Engine SDK path '" + getJavaAppEngineSdkPath()
               + "' is not valid");
     }
-    if (!getJavaToolsJar().toFile().isFile()) {
+    if (!JAR_LOCATIONS.get(JAVA_TOOLS_JAR).toFile().isFile()) {
       throw new AppEngineException(
-          "Validation Error: Java Tools jar path '" + getJavaToolsJar() + "' is not valid");
+          "Validation Error: Java Tools jar path '"
+                  + JAR_LOCATIONS.get(JAVA_TOOLS_JAR) + "' is not valid");
     }
   }
 
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -251,10 +251,12 @@ public class CloudSdk {
 
     /**
      * The home directory of Google Cloud SDK.
+     *
+     * @param sdkPath the root path for the Cloud SDK
      */
-    public Builder sdkPath(Path sdkPathFile) {
-      if (sdkPathFile != null) {
-        this.sdkPath = sdkPathFile;
+    public Builder sdkPath(Path sdkPath) {
+      if (sdkPath != null) {
+        this.sdkPath = sdkPath;
       }
       return this;
     }
@@ -352,7 +354,7 @@ public class CloudSdk {
     /**
      * Create a new instance of {@link CloudSdk}.
      *
-     * If {@code sdKPath} is not set, this method will attempt to look for the SDK in known install
+     * <p>If {@code sdKPath} is not set, this method will look for the SDK in known install
      * locations.
      */
     public CloudSdk build() {
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -25,6 +25,7 @@ import com.google.cloud.tools.appengine.cloudsdk.internal.process.WaitingProcess
 import com.google.cloud.tools.appengine.cloudsdk.process.ProcessExitListener;
 import com.google.cloud.tools.appengine.cloudsdk.process.ProcessOutputLineListener;
 import com.google.cloud.tools.appengine.cloudsdk.process.ProcessStartListener;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import com.google.common.collect.Maps;
 
@@ -175,7 +176,7 @@ public class CloudSdk {
     logger.info("submitting command: " + WHITESPACE_JOINER.join(command));
   }
 
-  private Path getSdkPath() {
+  @VisibleForTesting public Path getSdkPath() {
     return sdkPath;
   }
 
@@ -195,7 +196,7 @@ public class CloudSdk {
     return getSdkPath().resolve(JAVA_APPENGINE_SDK_PATH);
   }
 
-  private Path getJavaToolsJar() {
+  public Path getJavaToolsJar() {
     return getJavaAppEngineSdkPath().resolve(JAVA_TOOLS_JAR);
   }
 
@@ -253,9 +254,9 @@ public class CloudSdk {
      * The home directory of Google Cloud SDK. If not set, will attempt to look for the SDK in known
      * install locations.
      */
-    public Builder sdkPath(File sdkPathFile) {
+    public Builder sdkPath(Path sdkPathFile) {
       if (sdkPathFile != null) {
-        this.sdkPath = sdkPathFile.toPath();
+        this.sdkPath = sdkPathFile;
       }
       return this;
     }
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -231,7 +231,7 @@ public class CloudSdk {
     }
     if (!Files.isDirectory(sdkPath)) {
       throw new AppEngineException(
-          "Validation Error: SDK location '" + sdkPath + "' is a directory.");
+          "Validation Error: SDK location '" + sdkPath + "' is not a directory.");
     }
     if (!Files.isRegularFile(getGCloudPath())) {
       throw new AppEngineException(
@@ -373,7 +373,7 @@ public class CloudSdk {
     /**
      * Create a new instance of {@link CloudSdk}.
      *
-     * <p>If {@code sdKPath} is not set, this method will look for the SDK in known install
+     * <p>If {@code sdkPath} is not set, this method will look for the SDK in known install
      * locations.
      */
     public CloudSdk build() {
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -29,6 +29,7 @@ import com.google.common.base.Joiner;
 import com.google.common.collect.Maps;
 
 import java.io.File;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -174,7 +175,7 @@ private void logCommand(List<String> command) {
     logger.info("submitting command: " + WHITESPACE_JOINER.join(command));
   }
 
-  private Path getSdkPath() {
+  @VisibleForTesting public Path getSdkPath() {
```
elharo: why @VisibleForTesting? It seems useful and safe enough.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -252,9 +253,9 @@ public void validate() throws AppEngineException {
      * The home directory of Google Cloud SDK. If not set, will attempt to look for the SDK in known
      * install locations.
      */
-    public Builder sdkPath(File sdkPathFile) {
+    public Builder sdkPath(Path sdkPathFile) {
```
elharo: The comment seems questionable or perhaps misplaced. This method doesn't look anywhere.

joaoandremartins:
@elharo PTAL

elharo:
Tests appear to be failing due to:

[INFO] --- maven-checkstyle-plugin:2.17:check (default) @ appengine-plugins-core ---
[WARNING] src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java:[354] (javadoc) JavadocParagraph: Empty line should be followed by <p> tag on the next line.

This shouldn't cuase a failure though.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -195,7 +195,7 @@ public Path getJavaAppEngineSdkPath() {
     return getSdkPath().resolve(JAVA_APPENGINE_SDK_PATH);
   }
 
-  private Path getJavaToolsJar() {
+  public Path getJavaToolsJar() {
```
briandealwis: I wonder if these type of methods always return a `Collection<Path>` or `Path[]` for future-proofing (what happens if AppEngine's tools is split into multiple jars)?  

I think there are two approaches:
  1. The caller should know where the particular jars are situated within the Cloud SDK, in which case we just need a `resolve()` type method.  We're doing this in the Eclipse classpath providers, where we're pulling out runtime jars (servlet-api.jar and asp-api.jar).

  2. The locations are masked by this library, and the caller instead requests them by purpose, like this method.  Means we need to continue to modify and update this class as new libraries become included in the Cloud SDK.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -250,12 +250,11 @@ public void validate() throws AppEngineException {
     private int runDevAppServerWaitSeconds;
 
     /**
-     * The home directory of Google Cloud SDK. If not set, will attempt to look for the SDK in known
-     * install locations.
+     * The home directory of Google Cloud SDK.
```
elharo: @param sdkPath the home ...

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -250,12 +250,11 @@ public void validate() throws AppEngineException {
...
      */
-    public Builder sdkPath(File sdkPathFile) {
+    public Builder sdkPath(Path sdkPathFile) {
```
elharo: sdkPathFile --> sdkPath

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -352,6 +351,9 @@ public Builder runDevAppServerWait(int runDevAppServerWaitSeconds) {
 
     /**
      * Create a new instance of {@link CloudSdk}.
+     *
```
elharo: for the moment get rid of this blank line until we can configure checkstyle to ignore this.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -352,6 +351,9 @@ public Builder runDevAppServerWait(int runDevAppServerWaitSeconds) {
...
+     * If {@code sdKPath} is not set, this method will attempt to look for the SDK in known install
```
elharo: delete "attempt to"

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
meltsufin: I think you need `<p>` tag here. See: https://google.github.io/styleguide/javaguide.html#s7.1.2-javadoc-paragraphs

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
loosebazooka: 1. Maybe the naming is wrong, but it's just a reference to the location of appcfg.

2. I think the actual expectation is that no one uses these. They really should be internal. Do you guys want to use this for something more than just testing?

joaoandremartins: I think it could be fair that the client has some knowledge of the SDK structure. The alternative would be to have a method to return each and every one of the SDK JARs, which might be a bit unsightly.
What do you guys think?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
joaoandremartins: Added a <p> tag at the new line beginning, let's see if it likes it.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
joaoandremartins: I would like to worry about the case where the SDK is split into different directories later, when we have to.

joaoandremartins: @loosebazooka it's being used for a few things in the Eclipse plugin.
Find shared/servlet-api.jar and shared/jsp-api.jar, display the current location on the text box, set it for internal use, check that the path exists (though this should really be being done at the library), etc.

joaoandremartins: My above response was actually for getJavaAppEngineSdkPath(). Eclipse is only using getJavaToolsJar() to add its location to the classpath IIUC.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
meltsufin: If you run "mvn clean install" or "mvn checkstyle:check" locally, it will do the checkstyle checking.

joaoandremartins:
@elharo @briandealwis PTAL. I added JAR location resolution and some unit tests.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -93,6 +95,12 @@ private CloudSdk(Path sdkPath, String appCommandMetricsEnvironment,
     this.processRunner = new DefaultProcessRunner(async, stdOutLineListeners, stdErrLineListeners,
         exitListeners, startListeners);
 
+    // Populate jar locations.
```
elharo: something I just discovered: not all google cloud SDK installs have this. The app-engine-java component needs to be explicitly installed. 

Not for this PR, but we need to think about what happens if these paths aren't present.

elharo:
LGTM

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -229,9 +243,10 @@ public void validate() throws AppEngineException {
           "Validation Error: Java App Engine SDK path '" + getJavaAppEngineSdkPath()
               + "' is not valid");
     }
-    if (!getJavaToolsJar().toFile().isFile()) {
+    if (!JAR_LOCATIONS.get(JAVA_TOOLS_JAR).toFile().isFile()) {
       throw new AppEngineException(
-          "Validation Error: Java Tools jar path '" + getJavaToolsJar() + "' is not valid");
+          "Validation Error: Java Tools jar path '"
+                  + JAR_LOCATIONS.get(JAVA_TOOLS_JAR) + "' is not valid");
```
elharo: we should have a more specific message explaining why the location is invalid

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -229,9 +243,10 @@ public void validate() throws AppEngineException {
           "Validation Error: Java App Engine SDK path '" + getJavaAppEngineSdkPath()
               + "' is not valid");
     }
-    if (!getJavaToolsJar().toFile().isFile()) {
+    if (!JAR_LOCATIONS.get(JAVA_TOOLS_JAR).toFile().isFile()) {
```
elharo: Files.isRegularFile(). You don't need to convert to a java.io.File with toFile here and above.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
briandealwis: And that'll be true for GKE support too.

joaoandremartins: Added a TODO. We should probably just detect if the Java SDK is installed and ask the user to install it if it isn't.

loosebazooka: @joaoandremartins that's a good idea. Can you put that down as an issue on the repository?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
joaoandremartins: Replaced all toFile() by calls to Files.isRegularFile()/isDirectory().

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
joaoandremartins: I wonder if we can piggyback on #165? I'll link the TODO I added there, and maybe even assign it to myself.

elharo: In the library we can't assume there's a user. Any asking of user needs to be done from the UI.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -212,26 +229,28 @@ public void validate() throws AppEngineException {
     if (sdkPath == null) {
       throw new AppEngineException("Validation Error: SDK path is null");
     }
-    if (!sdkPath.toFile().isDirectory()) {
+    if (!Files.isDirectory(sdkPath)) {
       throw new AppEngineException(
-          "Validation Error: SDK directory '" + sdkPath + "' is not valid");
+          "Validation Error: SDK location '" + sdkPath + "' is a directory.");
```
elharo: is not a directory

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -352,6 +372,9 @@ public Builder runDevAppServerWait(int runDevAppServerWaitSeconds) {
 
     /**
      * Create a new instance of {@link CloudSdk}.
+     *
+     * <p>If {@code sdKPath} is not set, this method will look for the SDK in known install
```
elharo: sdkPath

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -93,6 +96,14 @@ private CloudSdk(Path sdkPath, String appCommandMetricsEnvironment,
     this.processRunner = new DefaultProcessRunner(async, stdOutLineListeners, stdErrLineListeners,
         exitListeners, startListeners);
 
+    // Populate jar locations.
+    // TODO(joaomartins): Consider case where SDK doesn't contain these jars. Only App Engine
+    // SDK does.
+    JAR_LOCATIONS.put("servlet-api.jar",
```
loosebazooka: I know this has been submitted, but there is no guarantee these jars will exist in the future in the cloud SDK. What is the need for these?

This CL seems to be modifying more than is intended in it's title.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -250,12 +269,13 @@ public void validate() throws AppEngineException {
...
-     * The home directory of Google Cloud SDK. If not set, will attempt to look for the SDK in known
-     * install locations.
+     * The home directory of Google Cloud SDK.
+     *
+     * @param sdkPath the root path for the Cloud SDK
      */
-    public Builder sdkPath(File sdkPathFile) {
-      if (sdkPathFile != null) {
-        this.sdkPath = sdkPathFile.toPath();
+    public Builder sdkPath(Path sdkPath) {
```
loosebazooka: The Api was made to expose only File, hence sdkPath(File...)

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
elharo: These jars are needed to setup projects in Eclipse, and to run local servers.

loosebazooka: And on top of that, why provide this as a map?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
joaoandremartins: Are you suggesting any change is made here?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
joaoandremartins: It's implemented as a map, but provided as API methods. Do you have any alternative suggestion?

loosebazooka: It's dangerous to have this dependency, api tools jar was purely for running staging and nothing else. The appengine sdk dependency may or may not exist in the future, we expect a standalone staging command to replace going into the java SDK here. I would generally avoid depending on anything in the java SDK included in the cloud sdk.

loosebazooka: Would it be difficult to get these jars from somewhere else?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
loosebazooka: The Api to the library was accepting File instead of Path. Having this builder accept path affects that consistency. Having said that, it's not a really big deal. I would have liked this CL to be more constrained in it's purpose though. It really seems to have pushed a lot of things in that could have been discussed.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
joaoandremartins: The tools jar is used in https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineSdkClasspathContainer.java#L40. I just assumed it was needed and bundled the change due to the same requirement in https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProvider.java#L39.

If the jars are in fact needed, then I think they should come from the SDK, hence the library should give them.

Although I agree that any functionality should probably be provided through an API, not JAR imports, but that is a separate conversation to this PR.

joaoandremartins:
Paths were already used over Files on the proof-of-concept implementation, at first, as per Patrick's suggestion and best practices.
I seem to recall having commented on the usage of File on your big PR, but maybe that comment fell through the cracks.

I couldn't make out any other changes to the existing API that caused disruption.

elharo:
If we're going to run a local Java server, and if that is a function of gcloud (a critical function I would argue) then I don't see how it can be without servlet.jar and these other jars somewhere. If gcloud moves these, then this API gives us enough indirection to reflect the new location without changing clients.

joaoandremartins:
The API let's you run a local devserver. https://github.com/GoogleCloudPlatform/app-tools-lib-for-java/blob/master/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer.java#L53

loosebazooka:
Is there somewhere you can point me to where you are using these jars?

elharo:
We add these to the classpath when creating a new App Engine project in Eclipse, or adding the App Engine facet to an existing non-maven project.

They're also implicitly used when running a local development server.

joaoandremartins:
@loosebazooka @meltsufin do we have to wait for a new release to upload new changes to Sonatype? https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/335 depends on this change and I can't submit it until these changes are published.

PR #168, opened 2016-07-14
Allow configuring other strategies for finding the Google Cloud SDK.

### LINKED ISSUE #169: Provide hooks for locating the Google Cloud SDK in other locations
When an explicit location has not been specified, `CloudSdk.Builder` uses `PathResolver` to try to discover the location of the Google Cloud SDK based on a set of heurstics by looking in a few fixed locations and examining the directories on the `PATH`, or specified in `GOOGLE_CLOUD_SDK_HOME`.  Many developers work on locked-down machines, or manage their disk layouts in different fashions.  For example, many Unix users put installations under `/opt`.  Some package managers consider `/usr/local` to be their own managed area.  There is much evidence that many developers struggle to change their `PATH`.

So for the Eclipse tooling, we provide the user means to configure the locations through a preference page ([issue](https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/issues/231), [commit](https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/255)).  This solution requires our code to use two new classes to fetch a `CloudSdk` instance rather than use the library's `Cloudsdk.Builder` directly:
  - `CloudSdkProvider` is a factory that uses the user preference to set `CloudSdk.Builder#sdkPath()`, and otherwise delegates to `PathResolver`
  - `CloudSdkPrompter` is like `CloudSdkProvider` but will prompt the user if the preference is unset and `PathResolver`  is unable to find the SDK

The downside to this solution is that we must be vigilant and stamp out any use of `CloudSdk.Builder` and convert the code to use our `CloudSdkProvider` or `CloudSdkPrompter` instead.  Otherwise we risk the SDK not being found _despite the user preference_.  We've already hit occasions where someone has committed code that uses `new CloudSdk.Builder().build()` instead of our `CloudSdkProvider` and thus caused breakage.  This is somewhat less than optimal.

It could be argued that `PathResolver`'s heuristics should be improved by specifying additional locations.  But I don't think we can possibly satisfy everyone.  And this doesn't help the poor developer who uses yet another different location that we haven't supported until now.

I propose that `CloudSdk.Builder` instead support a mechanism to provide additional path resolvers for locating the Google Cloud SDK.  We would then rewrite the Eclipse `CloudSdkProvider` as such a resolver.  Ideally this mechanism should not require any changes for the consumers of the library, such that they can continue to simply call `new CloudSdk.Builder().build()`.

briandealwis:
Currently the Eclipse Tooling has wrapper classes (`CloudSdkProvider` and `CloudSdkPrompter`) for resolving the Google Cloud SDK using a user-provided location.  This solution is error-prone as all code that tries to obtain a `CloudSdk` instance must funnel through these wrapper classes.  Failures occur if there are any code paths that uses the `CloudSdk.Builder` directly.

This patch extracts an interface `CloudSdkResolver` modelled on PathResolver.  The `CloudSdk` instance looks up additional `CloudSdkResolver` classes using the standard `java.util.ServiceLoader` mechanism.  To support registering resolvers under Eclipse, where we have separate class loaders, we register the bundle using the Eclipse Buddy Policy; additional resolvers can also be contributed using OSGi fragments.

Update: Fixes #169

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -31,7 +31,6 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 
 import java.io.File;
-import java.io.Serializable;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -456,12 +455,11 @@ public class CloudSdk {
   /**
    * Compare two {@link CloudSdkResolver} instances by their rank.
    */
-  private static class ResolverComparator implements Comparator<CloudSdkResolver>, Serializable {
-    private static final long serialVersionUID = -4414214946468755954L;
-
+  private static class ResolverComparator implements Comparator<CloudSdkResolver> {
     @Override
     public int compare(CloudSdkResolver o1, CloudSdkResolver o2) {
       return o1.getRank() - o2.getRank();
     }
+
   }
 }
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -453,7 +453,7 @@ public class CloudSdk {
   }
 
   /**
-   * Compare two {@link CloudSdkResolver} instances by their rank
+   * Compare two {@link CloudSdkResolver} instances by their rank.
    */
   private static class ResolverComparator implements Comparator<CloudSdkResolver> {
     @Override
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -412,7 +412,7 @@ public class CloudSdk {
           if (discoveredSdkPath != null) {
             return discoveredSdkPath;
           }
-        } catch (Exception ex) {
+        } catch (RuntimeException ex) {
           // prevent interference from exceptions in other resolvers
           logger.log(Level.SEVERE, resolver.getClass().getName()
               + ": exception thrown when searching for Google Cloud SDK", ex);
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -31,6 +31,7 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 
 import java.io.File;
+import java.io.Serializable;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -455,11 +456,12 @@ public class CloudSdk {
   /**
    * Compare two {@link CloudSdkResolver} instances by their rank.
    */
-  private static class ResolverComparator implements Comparator<CloudSdkResolver> {
+  private static class ResolverComparator implements Comparator<CloudSdkResolver>, Serializable {
+    private static final long serialVersionUID = -4414214946468755954L;
+
     @Override
     public int compare(CloudSdkResolver o1, CloudSdkResolver o2) {
       return o1.getRank() - o2.getRank();
     }
-
   }
 }
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -25,7 +25,9 @@ import com.google.cloud.tools.appengine.cloudsdk.internal.process.WaitingProcess
 import com.google.cloud.tools.appengine.cloudsdk.process.ProcessExitListener;
 import com.google.cloud.tools.appengine.cloudsdk.process.ProcessOutputLineListener;
 import com.google.cloud.tools.appengine.cloudsdk.process.ProcessStartListener;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 
 import java.io.File;
@@ -36,8 +38,11 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.ServiceLoader;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 /**
@@ -266,6 +271,7 @@ public class CloudSdk {
     private List<ProcessOutputLineListener> stdErrLineListeners = new ArrayList<>();
     private List<ProcessExitListener> exitListeners = new ArrayList<>();
     private List<ProcessStartListener> startListeners = new ArrayList<>();
+    private List<CloudSdkResolver> resolvers;
     private int runDevAppServerWaitSeconds;
 
     /**
@@ -380,12 +386,7 @@ public class CloudSdk {
 
       // Default SDK path
       if (sdkPath == null) {
-        Path discoveredSdkPath = new PathResolver().getCloudSdkPath();
-        if (discoveredSdkPath == null) {
-          throw new AppEngineException("Google Cloud SDK path was not provided and could not be"
-              + " found in any known install locations.");
-        }
-        sdkPath = discoveredSdkPath;
+        sdkPath = discoverSdkPath();
       }
 
       return new CloudSdk(sdkPath, appCommandMetricsEnvironment,
@@ -394,5 +395,56 @@ public class CloudSdk {
           startListeners, runDevAppServerWaitSeconds);
     }
 
+    /**
+     * Attempt to find the Google Cloud SDK in various places.
+     * 
+     * @return the path to the root of the Google Cloud SDK
+     * @throws AppEngineException if not found
+     */
+    @Nonnull
+    private Path discoverSdkPath() {
+      for (CloudSdkResolver resolver : getResolvers()) {
+        try {
+          Path discoveredSdkPath = resolver.getCloudSdkPath();
+          if (discoveredSdkPath != null) {
+            return discoveredSdkPath;
+          }
+        } catch (Exception ex) {
+          logger.log(Level.SEVERE, resolver.getClass().getName()
+              + ": exception thrown when searching for Google Cloud SDK", ex);
+        }
+      }
+      throw new AppEngineException("Google Cloud SDK path was not provided and could not be"
+          + " found in any known install locations.");
+    }
+
+    /**
+     * Return the configured SDK resolvers.
+     */
+    @VisibleForTesting
+    public List<CloudSdkResolver> getResolvers() {
+      if (this.resolvers != null) {
+        return resolvers;
+      }
+      // Explicitly specify classloader rather than use the TCCL
+      ServiceLoader<CloudSdkResolver> services =
+          ServiceLoader.load(CloudSdkResolver.class, getClass().getClassLoader());
+      List<CloudSdkResolver> resolvers = Lists.newArrayList(services);
+      // Explicitly add the PATH-based resolver last as ServiceLoader
+      // does not provide any ordering
+      resolvers.add(new PathResolver());
+      return resolvers;
+    }
+
+    /*
+     * Set the list of path resolvers to locate the Google Cloud SDK. Intended for tests to
+     * precisely control where the SDK may be found.
+     */
+    @VisibleForTesting
+    public Builder resolvers(List<CloudSdkResolver> resolvers) {
+      this.resolvers = resolvers;
+      return this;
+    }
   }
+
 }
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -35,6 +35,8 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Collections;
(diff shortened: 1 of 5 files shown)


briandealwis:
We could consider splitting out the PathResolver into separate resolvers:
  - Windows resolver
  - Environment variable resolver (_GOOGLE_CLOUD_SDK_HOME_)
  - System property resolver
We'd need to have some way to rank and order the resolvers.  `CloudSdkResolver` could have a method to return a priority value.

briandealwis:
Mentioned to João on chat:
> The nice thing about these Resolver instances is that they're automatically registered.

I also wondered about turning the `CloudSdkPrompter`, which prompts the user to download and configure the Google Cloud SDK if required, into a last-ditch resolver:

> I had wondered if we could turn the [eclipse-side] CloudSdkPrompter into the last-ditch Resolver that would prompt the user *if* it was running on the SWT thread and *if* there wasn't a some specific ThreadLocal flag to disable prompting.
> Turning the [eclipse-side] Prompter into a Resolver also reduces the number of calling conventions to create / discover the SDK.

That would require some way to order the resolvers.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -371,5 +372,56 @@ public CloudSdk build() {
...
+     */
+    @Nonnull
+    private Path discoverSdkPath() {
+      for (CloudSdkResolver resolver : getResolvers()) {
+        try {
+          Path discoveredSdkPath = resolver.getCloudSdkPath();
+          if (discoveredSdkPath != null) {
+            return discoveredSdkPath;
+          }
+        } catch (Exception ex) {
```
elharo: more specific exceptions please

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -371,5 +372,56 @@ public CloudSdk build() {
...
+      throw new AppEngineException("Google Cloud SDK path was not provided and could not be"
+          + " found in any known install locations.");
+    }
+
+    /**
+     * Return the configured SDK resolvers.
+     */
+    @VisibleForTesting
+    public List<CloudSdkResolver> getResolvers() {
+      if (this.resolvers != null) {
```
elharo: even if it's empty?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkResolver.java
```diff
@@ -0,0 +1,32 @@
...
+ * Resolve paths with CloudSdk and Python defaults.
+ */
+public interface CloudSdkResolver {
+
+  /**
+   * Attempts to find the path to Google Cloud SDK.
+   *
+   * @return Path to Google Cloud SDK or null
+   */
+  public Path getCloudSdkPath();
```
elharo: public not needed here

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -357,12 +363,7 @@ public CloudSdk build() {
 
       // Default SDK path
       if (sdkPath == null) {
-        Path discoveredSdkPath = new PathResolver().getCloudSdkPath();
-        if (discoveredSdkPath == null) {
-          throw new AppEngineException("Google Cloud SDK path was not provided and could not be"
-              + " found in any known install locations.");
-        }
-        sdkPath = discoveredSdkPath;
+        sdkPath = discoverSdkPath();
```
loosebazooka: Seems a little overly complex when you can just specify the location of your sdk in your tests directly and appears to have some very osgi specific work arounds. Is there a reason this is absolutely required?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
joaoandremartins: How about turning this into a single "if (this.resolvers == null)" and only "return resolvers" once?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -371,5 +372,56 @@ public CloudSdk build() {
...
+
+    /**
+     * Return the configured SDK resolvers.
+     */
+    @VisibleForTesting
+    public List<CloudSdkResolver> getResolvers() {
+      if (this.resolvers != null) {
+        return resolvers;
+      }
+      // Explicitly specify classloader rather than use the TCCL
```
joaoandremartins: What's TCCL?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -371,5 +372,56 @@ public CloudSdk build() {
...
     }
 
+    /**
+     * Attempt to find the Google Cloud SDK in various places.
+     * 
+     * @return the path to the root of the Google Cloud SDK
+     * @throws AppEngineException if not found
+     */
+    @Nonnull
+    private Path discoverSdkPath() {
```
joaoandremartins: build() is simple enough that we could just dump this code there and get rid of this extra method?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -371,5 +372,56 @@ public CloudSdk build() {
...
+      resolvers.add(new PathResolver());
+      return resolvers;
+    }
+
+    /*
+     * Set the list of path resolvers to locate the Google Cloud SDK. Intended for tests to
+     * precisely control where the SDK may be found.
+     */
+    @VisibleForTesting
+    public Builder resolvers(List<CloudSdkResolver> resolvers) {
```
joaoandremartins: Should we instead turn `resolvers` into a `Set` and have an `addResolver(s)` and `removeResolver`?

As is, the client needs to first get the resolvers to grab PathResolver using `@VisibleForTesting getResolvers()`, add their resolvers and then set the final resolver list.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -371,5 +372,56 @@ public CloudSdk build() {
...
 
+    /**
+     * Attempt to find the Google Cloud SDK in various places.
+     * 
+     * @return the path to the root of the Google Cloud SDK
+     * @throws AppEngineException if not found
+     */
+    @Nonnull
+    private Path discoverSdkPath() {
+      for (CloudSdkResolver resolver : getResolvers()) {
```
joaoandremartins: I would like to see some kind of order here so that the ordered queried resolvers are, for example, in Eclipse's case:
1) Context resolver,
2) Preference store resolver,
3) PathResolver (we should rename this to AutoDiscoveryResolver or something else),
4) CloudSdkPrompter,
etc.

joaoandremartins:
@briandealwis I noticed you assigned this issue to me. Did you just want me to review this PR or do you expect me to follow up on your work here?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
briandealwis: I think it should be a `List` — there should be an order.

The real value here is the use of `ServiceLocator` such that we disconnect the consumer of the `CloudSdk` from the providers who know how to best find the SDK.  The only real users of `resolvers()` will be tests.  After all, if the creator of the `CloudSdk.Builder` wants to add a resolver, then they know (or have the resolver instance to compute) the path and can set `sdkPath` explicitly.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
briandealwis: ok

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
briandealwis: The reasoning is scattered across a couple of issues and branches.  Rather than re-summarize here, I've opened issue #169 as I should have done earlier.  This is much more than about tests.

`ServiceLocator` is a Java thing, not an OSGi thing.  The Eclipse Buddy manifest header makes our lives slightly easier; if it's really a blocker, I'll remove it and we'll work around it.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
briandealwis: This is to prevent inadvertent exceptions from breaking the flow (e.g., `IOException` or `IllegalArgumentException`).  Perhaps I should just trap Throwable.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
briandealwis: Elliotte: I was thinking of a test that wanted total control and have no resolvers.

João: my preference is for early exits over single-return. Makes for smaller, easier to understand methods.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
briandealwis: Thread Context Class Loader.  An attempt from JEE to get around some of the issues around managed class paths. Eclipse/Equinox has something that works, but it is best to specify a class loader if possible.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
briandealwis: I tried inlining it and I think `build()` becomes more complicated.  Small methods are easier to grok.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
briandealwis: To see the difference it makes in the Eclipse tools: https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/362/files

briandealwis:
@joaoandremartins to review, please — I've done the follow-on work in https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/362 (to be modified once https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/335 lands)

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
elharo: No, catching Exception is highly discouraged. (See Bloch, Effective Java). Catching Throwable is worse. In some cases it's OK to catch RuntimeException, but I'm not sure this is one of them. Usually you only want to do that when you don't know what code might be invoked. Here we should only catch what getCloudSdkPath might plausibly throw.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
elharo: setResolvers(). resolvers sounds more like a getter to me.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
briandealwis: You're right, it should be `RuntimeException`.  An exception here indicates a programming error in and should be treated as if the code had returned `null` — it didn't find an SDK in some other location.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
briandealwis: It's on the `Builder` and `resolvers()` seems more fluent-like.  At least `setResolvers()` looks weird with fluent use.

joaoandremartins:
@briandealwis GoogleCloudPlatform/gcloud-eclipse-tools#335 landed, let me know when this is ready for another review.

briandealwis:
Please do @joaoandremartins — PTAL. It's now a priority because of gcloud-eclipse-tools#335 as you cut over `CloudSdkProvider` calls to `Cloudsdk.Builder.build()`, which breaks me ;-)

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -394,5 +398,70 @@ public CloudSdk build() {
...
+    public Builder resolvers(List<CloudSdkResolver> resolvers) {
+      this.resolvers = resolvers;
+      return this;
+    }
+  }
+
+  /**
+   * Compare two {@link CloudSdkResolver} instances by their rank.
+   */
+  private static class ResolverComparator implements Comparator<CloudSdkResolver>, Serializable {
```
loosebazooka: why does this need to be serializable?

briandealwis: Good question but Findbugs failed the build without it :-S

elharo: we should remove that check from our FindBugs config. It's not relevant to us. 

http://findbugs.sourceforge.net/bugDescriptions.html#SE_COMPARATOR_SHOULD_BE_SERIALIZABLE

joaoandremartins:
Besides the resolver order issue, LGTM.

briandealwis:
> @joaoandremartins Besides the resolver order issue, LGTM.

I addressed the ordering issue in https://github.com/GoogleCloudPlatform/app-tools-lib-for-java/pull/168/commits/d6e67b68eeb762b41d042deb1dfb20cfd164beb4 — or was there something else?

joaoandremartins:
Ah, didn't see that.
LGTM

PR #170, opened 2016-07-18
version to 0.1.4

elharo:
@joaoandremartins

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
 
   <groupId>com.google.cloud.tools</groupId>
   <artifactId>appengine-plugins-core</artifactId>
-  <version>0.1.3-SNAPSHOT</version>
+  <version>0.1.4-SNAPSHOT</version>
 
   <name>App Engine Plugins Core Library</name>
   <description>


joaoandremartins:
@elharo can you help me remember the reason for the version change?
Should we come up with a policy for changing versions, or is there one already?

briandealwis:
There were some breaking changes that bit CT4E since we are pulling 0.1.3-SNAPSHOT.

  - `PathResolver.INSTANCE` removed
  - `CloudSdk.Builder.sdkPath(path)` argument type changes from `File` &rarr;  `Path`

> Should we come up with a policy for changing versions, or is there one already?

@elharo's proposing that CT4E use N-1 to shield us from breaking changes.  It'd be good if the library incremented the version on ABI or behavioural breaking changes.

joaoandremartins:
LGTM

elharo:
The proposal is that N-1 has a stable API while the current version can change. That way CT4E and CT4IJ can move when they're ready.

meltsufin:
We normally bump the version right after a release. If you would like to avoid breaking changes, pull in a release version. Similarly, if you need to lock in certain changes, request a release. The master version is intended to always be {latest-release-version + 1}-SNAPSHOT.

PR #176, opened 2016-07-21
Fail flexible staging when artifact does not exist

### LINKED ISSUE #175: Staging Flexible Apps should fail when artifact doesn't exist

meltsufin:
Fixes #175 
@loosebazooka


loosebazooka:
:ship:

PR #177, opened 2016-07-21
Configure Travis to Deploy Snapshots on commit to master

### LINKED ISSUE #86: Configure Travis to Deploy Snapshots on commit to master branch

meltsufin:
Fixes #86

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ cache:
 
 script:
   - |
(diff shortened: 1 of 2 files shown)


meltsufin:
@loosebazooka PTAL
I verified that it works using the previous commit to this branch.

loosebazooka:
can you ensure it's only snapshots?

meltsufin:
Maybe, but why? The master should never have a non-snapshot version. Even if it did, it would never make it out to Maven Central without additional steps.

loosebazooka:
Well, either way a check would be nice. Can you file a bug and then :ski:

PR #204, opened 2016-09-05
Project ID validator

elharo:
@akerekes @meltsufin 

I want to move this code out of Eclipse. 

This introduces a new package structure above app engine since this projects and project IDs are not limited to App Engine.

see https://github.com/GoogleCloudPlatform/cloud-tools-for-eclipse/issues/666

### DIFF
--- a/checkstyle-suppressions.xml
+++ b/checkstyle-suppressions.xml
@@ -19,5 +19,6 @@ limitations under the License.
 -->
 <suppressions>
   <!-- Suppress paragraph validation in copyright header -->
-  <suppress files=".*\.java" checks="JavadocParagraph"/>
+  <suppress files=".*\.java" checks="JavadocParagraph" lines="1-12"/>
+  <suppress files="ProjectIdValidator\.java" checks="JavadocParagraph"/>
 </suppressions>
--- a/checkstyle-suppressions.xml
+++ b/checkstyle-suppressions.xml
@@ -19,5 +19,5 @@ limitations under the License.
 -->
 <suppressions>
   <!-- Suppress paragraph validation in copyright header -->
-  <suppress files=".*\.java" checks="JavadocParagraph" lines="1-12"/>
+  <suppress files=".*\.java" checks="JavadocParagraph" lines="1-16"/>
 </suppressions>
--- a/checkstyle-suppressions.xml
+++ b/checkstyle-suppressions.xml
@@ -19,5 +19,5 @@ limitations under the License.
 -->
 <suppressions>
   <!-- Suppress paragraph validation in copyright header -->
-  <suppress files=".*\.java" checks="JavadocParagraph" lines="1-16"/>
+  <suppress files=".*\.java" checks="JavadocParagraph"/>
 </suppressions>
--- a/src/main/java/com/google/cloud/tools/project/ProjectIdValidator.java
+++ b/src/main/java/com/google/cloud/tools/project/ProjectIdValidator.java
@@ -41,8 +41,8 @@ public class ProjectIdValidator {
    * This method only checks syntax. It does not check that the ID
    * actually identifies a project in the Google Cloud Platform. 
    * 
-   * @param id the 
-   * @return true if it's correct
+   * @param id the alleged project ID
+   * @return true if it's correct, false otherwise
    */
   public static boolean validate(String id) {
     if (id == null) {
--- a/src/main/java/com/google/cloud/tools/project/ProjectIdValidator.java
+++ b/src/main/java/com/google/cloud/tools/project/ProjectIdValidator.java
@@ -17,7 +17,8 @@ import java.util.regex.Pattern;
  */
 public class ProjectIdValidator {
 
-  private static final Pattern PATTERN = Pattern.compile("([a-z\\d\\-\\.]{1,100}:)?[a-z\\d\\-\\.]{1,100}");
+  private static final Pattern PATTERN = Pattern.compile(
+      "([a-z\\d\\-\\.]{1,100}:)?[a-z\\d\\-\\.]{1,100}");
   
   /**
    * Check whether a string is a syntactically correct project ID.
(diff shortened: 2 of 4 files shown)


src/main/java/com/google/cloud/tools/project/ProjectIdValidator.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+public class ProjectIdValidator {
+
+  private static final Pattern PATTERN = Pattern.compile("([a-z\\d\\-\\.]{1,100}:)?[a-z\\d\\-\\.]{1,100}");
+  
+  public static boolean validate(String id) {
+    if (id == null) {
+      return false;
+    } else if (id.isEmpty()) {
+      return true;
```
meltsufin: Empty string is a valid project ID?

meltsufin:
Note the checkstyle errors.

src/main/java/com/google/cloud/tools/project/ProjectIdValidator.java
elharo: Good point. It is the way we use this class currently, but maybe shouldn't be when this gets moved to a more general utility. I'll change that.

elharo:
@meltsufin checkstyle errors resolved (though I have filed a bug against checkstyle with one I think is a false positive: 

https://github.com/checkstyle/checkstyle/issues/3427

not that they make it easy to send them a patch

src/main/java/com/google/cloud/tools/project/ProjectIdValidator.java
```diff
@@ -0,0 +1,39 @@
+package com.google.cloud.tools.project;
```
meltsufin: Missing copyright header.

checkstyle-suppressions.xml
```diff
@@ -19,5 +19,5 @@ limitations under the License.
 -->
 <suppressions>
   <!-- Suppress paragraph validation in copyright header -->
-  <suppress files=".*\.java" checks="JavadocParagraph" lines="1-12"/>
+  <suppress files=".*\.java" checks="JavadocParagraph" lines="1-16"/>
```
meltsufin: This was designed to exclude the copyright header. I think 1-12 is correct.

src/test/java/com/google/cloud/tools/project/ProjectIdValidatorTest.java
```diff
@@ -0,0 +1,46 @@
+package com.google.cloud.tools.project;
```
meltsufin: Copyright header missing

checkstyle-suppressions.xml
elharo: yes, I deliberately added it so it would workaround a buggy checkstyle requirement for <p> tags, though if I add the copyright header below that's not going to work anymore. 

Adding the <p> tag that checkstyle wants makes the code less legible. The problem is that checkstyle does not recognize other non-p block-level elements such as pre.

meltsufin:
The test is still missing the copyright header. Also, for the checkstyle, try this:

```
/**
 * Check if a string is a legal <a href='https://support.google.com/cloud/answer/6158840?hl=en'>Google
 * Cloud Platform Project ID</a>. Source: com.google.apphosting.base.AppId
 *
 * <p><pre>
 * app-id ::= [(domain):](display-app-id)
 * domain ::= r'(?!-)[a-z\d\-\.]{1,100}'
 * display-app-id ::= r'(?!-)[a-z\d\-]{1,100}'
 * </pre></p>
 *
 * <p>Note that in some older documentation this is referred to as an "application ID."</p>
 */
public class ProjectIdValidator {
```

elharo:
I added a copyright to the test. The fix you suggest will quiet checkstyle, but it violates the HTML spec, in which block level elements such as pre should not appear inside a p element. That's essentially the bug I filed against the checkstyle JavadocParagraph rule.

checkstyle-suppressions.xml
```diff
@@ -19,5 +19,5 @@ limitations under the License.
 -->
 <suppressions>
   <!-- Suppress paragraph validation in copyright header -->
-  <suppress files=".*\.java" checks="JavadocParagraph" lines="1-12"/>
+  <suppress files=".*\.java" checks="JavadocParagraph"/>
```
meltsufin: I would rather keep the JavadocParagraph check, with the copyright header exclusion, and a separate specific exclusion for the ProjectIdValidator.java.

src/main/java/com/google/cloud/tools/project/ProjectIdValidator.java
```diff
@@ -0,0 +1,55 @@
...
+ * app-id ::= [(domain):](display-app-id)
+ * domain ::= r'(?!-)[a-z\d\-\.]{1,100}'
+ * display-app-id ::= r'(?!-)[a-z\d\-]{1,100}'
+ * </pre>
+ * 
+ * Note that in some older documentation this is referred to as an "application ID."
+ */
+public class ProjectIdValidator {
+
+  private static final Pattern PATTERN = Pattern.compile(
```
akerekes: The validation implemented in [cloud-tools-for-eclipse](https://github.com/GoogleCloudPlatform/cloud-tools-for-eclipse/blob/59f8eea489f68b849ddfda02b91a0a22ba8bd2f5/plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectIdValidator.java#L22) is rather different than this. Which one is correct?

meltsufin:
LGTM

src/main/java/com/google/cloud/tools/project/ProjectIdValidator.java
elharo: done

src/main/java/com/google/cloud/tools/project/ProjectIdValidator.java
elharo: this one

src/test/java/com/google/cloud/tools/project/ProjectIdValidatorTest.java
elharo: done

elharo:
PTAL @meltsufin @akerekes

meltsufin:
LGTM

PR #206, opened 2016-09-07
notice when app engine java components aren't installed

### LINKED ISSUE #165: Provide more specific error when missing app-engine-java gcloud component
It would be nice if the library threw a specific exception if appcfg was unavailable (via the gcloud component). Currently this is what users are shown:

```
Error: Could not find or load main class com.google.appengine.tools.admin.AppCfg
Failed to deploy 'app-engine-random-service:war exploded': Deployment failed due to an unexpected error while staging the project.
```

elharo:
@meltsufin @etanshaul fix #165


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -59,6 +59,7 @@
   private static final String DEV_APPSERVER_PY = "bin/dev_appserver.py";
   private static final String JAVA_APPENGINE_SDK_PATH =
       "platform/google_appengine/google/appengine/tools/java/lib";
+  private static final String JAVA_COMPONENTS_PATH = "platform/gcd/.appengine";
```
meltsufin: Is it sufficient to just check of the existence of `JAVA_APPENGINE_SDK_PATH`?

elharo: No, that seems to exist regardless

meltsufin: We should probably check with the Cloud SDK team to see what they recommend using.

etanshaul: Would executing and parsing a `gcloud components list` be overkill?

meltsufin: @markpell, @znewman01 can you comment on this?

znewman01: Yeah, if you can run `gcloud components list` it would be more reliable/breakage-proof:

    $ gcloud components list --format='value(state)' --filter=id:app-engine-java
    Installed

(Or "Not Installed").

But `.install/google-app-engine.manifest` might be a better file to check the existence of. Mark will have to weigh in on how stable that file is.

markpell: I don't think we want anything depending on these internal paths as they are not necessarily stable.  I don't even think JAVA_APPENGINE_SDK_PATH is appropriate.  If you want to check components, you should probably shell out and use --format to get the output back in a machine readable format.

What exactly are you trying to detect the presence of?  The gcd directory is for the datastore emulator which is currently deprecated in favor of the new version.  Is this what you are looking for?

etanshaul: @markpell we mainly would like to detect that the java tools component is installed in gcloud since this is required for deployment from our tooling.

elharo: We want to know whether or not the user ran:

gcloud components install app-engine-java

We've seen multiple user reports that trace back to not having done this. 
This directory appears to be added/removed when the user installs/removes the component.

elharo:
$ gcloud components list --format='value(state)' --filter=id:app-engine-java

Your current Cloud SDK version is: 124.0.0
The latest available version is: 125.0.0

name=Not Installed

briandealwis:
```
$ gcloud components list --format='value(state)' --filter=id:app-engine-java

Your current Cloud SDK version is: 119.0.0
The latest available version is: 125.0.0

name=Update Available
```
Hmm, I should update.

etanshaul:
I think you'd want to set `--format=json`

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
briandealwis: FWIW: although my installation is out of date, I did have `app-engine-java`, but I don't have anything in `platform/gcd`.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -262,6 +268,10 @@ public void validate() throws AppEngineException {
           "Validation Error: Java Tools jar location '"
               + JAR_LOCATIONS.get(JAVA_TOOLS_JAR) + "' is not a file.");
     }
+    if (!Files.isDirectory(getJavaAppEngineComponentsPath())) {
```
patflynn: my concern about having this here is that we can support deploy without needing the java components installed.  A client should be able to get configure gcloud and do a deployment without install the components,

meltsufin: Right, this only applies to Standard and Compat deployments, not pure Flex.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
meltsufin: Since what we're actually using is the appengine-tools-api.jar, we should probably just check for its existence.

markpell: So again, this is just checking for the wrong thing.  That directory is for the legacy datastore emulator which has nothing to do with Java or any Java components.

How far are we from having the requiring staging code as separate artifacts on maven central?  I don't think it's appropriate for the plugin to be grabbing individual jars from the SDK as this is not a public interface.

patflynn: @ludoch who would know when staging is going to make it to maven central.

To summarize what's being discussed:

For component detection we should be calling gcloud components list and not digging into the cloud SDK packaged files. 

The staging question is not related to this CL but we are working on moving as many deps as possible away from using Cloud SDK paths.

elharo:
$ gcloud components list --format='json' --filter=id:app-engine-java

Your current Cloud SDK version is: 124.0.0
The latest available version is: 125.0.0

[
  {
    "current_version_string": null,
    "id": "app-engine-java",
    "is_configuration": false,
    "is_hidden": false,
    "latest_version_string": "1.9.42",
    "name": "gcloud app Java Extensions",
    "size": 142219489,
    "state": {
      "name": "Not Installed"
    }
  }
]

It really shouldn't be saying

Your current Cloud SDK version is: 124.0.0
The latest available version is: 125.0.0

znewman01:
It's only printing that to STDERR; it's very easy to ignore.

patflynn:
@elharo are we closing this to start over?

elharo:
Yes, the version that calls gcloud components shares very little with this branch.

elharo:
reopening since we may do this

PR #207, opened 2016-09-08
notice when app engine java components aren't installed

### LINKED ISSUE #165: Provide more specific error when missing app-engine-java gcloud component

elharo:
@meltsufin @joaoandremartins @etanshaul @markpell 

fix #165

Uses gcloud components list


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -262,11 +290,20 @@ public void validate() throws AppEngineException {
           "Validation Error: Java Tools jar location '"
               + JAR_LOCATIONS.get(JAVA_TOOLS_JAR) + "' is not a file.");
     }
+    try {
+      if (!isComponentInstalled("app-engine-java")) {
```
meltsufin: Isn't the block above that checks the existence of JAVA_TOOLS_JAR sufficient to tell if the component is installed? I'm still not a fan of doing this expensive check using `gcloud components list`.

briandealwis: :+1: to @meltsufin's comment: I tried running `gcloud` from a VM shared drive and it was glacially sloooow.  If JAVA_TOOLS_JAR (or one of the jars) isn't available then we're either dealing with a missing `app-engine-java` component, or `appengine-plugins-core` is out of sync with the installed version and

briandealwis: Could we use `JAVA_TOOLS_JAR` as a heuristic: if the path exists, then we seem to have `app-engine-java` installed, otherwise confirm with `gcloud components list`.

meltsufin: I've just tried running `gcloud components remove app-engine-java` on my Linux machine and it did delete `~/google-cloud-sdk/platform/google_appengine/google/appengine/tools/java/lib/appengine-tools-api.jar`. When installed it back using `gcloud components install app-engine-java`, it reappeared. So, I think checking for existence of `JAVA_TOOLS_JAR` is sufficient.
@elharo, do you see a different result when you run these commands?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/AppEngineComponentsNotInstalledException.java
```diff
@@ -0,0 +1,11 @@
+package com.google.cloud.tools.appengine.cloudsdk;
+
+import com.google.cloud.tools.appengine.api.AppEngineException;
+
+public class AppEngineComponentsNotInstalledException extends AppEngineException {
```
patflynn: Isn't AppEngineException for errors caused by gcloud invocations?

patflynn: also I haven't checked the styleguide recently, is the javadoc no longer required for all public classes?=

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -125,6 +128,31 @@ public void runAppCommand(List<String> args) throws ProcessRunnerException {
     logCommand(command);
     processRunner.setEnvironment(environment);
     processRunner.run(command.toArray(new String[command.size()]));
+    processRunner.run(command.toArray(new String[command.size()]));
+  }
+
+  /**
+   * @return true iff the specified component is installed in the local environment
```
patflynn: {@code true}

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -125,6 +128,31 @@ public void runAppCommand(List<String> args) throws ProcessRunnerException {
...
+   */
+  public boolean isComponentInstalled(String id) throws ProcessRunnerException {
+    List<String> command = new ArrayList<>();
+    command.add(getGCloudPath().toString());
+    command.add("components");
+    command.add("list");
+    command.add("--format=json");
+    command.add("--filter=id:" + id);
+    
+    if (stdOutput == null) {
```
patflynn: isn't stdOutput non-null? Should we just add Preconditions not null checks to the constructor?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -125,6 +128,31 @@ public void runAppCommand(List<String> args) throws ProcessRunnerException {
...
+      throw new IllegalStateException("won't be able to read output");
+    }
+    
+    stdOutput.clear();
```
patflynn: I guess this class is just generally not threadsafe?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
patflynn: The practice of depending on the Cloud SDKs internal directory structure is not OK with the Cloud SDK team.  They reserve the right to break this at their discretion.  We need to move away from depending on incidental files distributed with the SDK and moving to using their api/cli for this check is part of that.

This is also why we're moving to using Maven central as the source for classpath deps.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -233,7 +261,7 @@ public Path getJarPath(String jarName) {
   /**
    * Checks whether the configured Cloud SDK Path is valid.
    *
-   * @throws AppEngineException when there is a validation error.
+   * @throws AppEngineException when there is a validation error
    */
   public void validate() throws AppEngineException {
```
patflynn: if validate() expects errors why not just return the issues rather than throw them?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -70,22 +71,25 @@
...
 
   private CloudSdk(Path sdkPath,
                    String appCommandMetricsEnvironment,
                    String appCommandMetricsEnvironmentVersion,
                    @Nullable File appCommandCredentialFile,
                    String appCommandOutputFormat,
                    ProcessRunner processRunner,
-                   WaitingProcessOutputLineListener runDevAppServerWaitListener) {
+                   WaitingProcessOutputLineListener outputLineListener,
+                   AccumulatingLineListener stdOutput) {
```
patflynn: Do we really need two kinds of listener API's?  It seems unlikely the user would want to pass both?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
patflynn: what do you think of making the output handling completely internal to this method? It doesn't seem like clients should need to think about this.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/AppEngineComponentsNotInstalledException.java
elharo: added Javadoc. This seems close enough to AppEngineException to justify subclassing. I can change to a different exception if you like, but that would require clients to update.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
elharo: I would love to do that. I think it might require major refactoring of the API though, but let me give it a bit more thought and see if I can crack it.

elharo: There is a code path along which stdOutput is not set, basically if inheritProcessOutput is true. See line 472-478. I'm not sure if we really need inheritProcessOutput though. Maybe we can drop it?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
elharo: I think ProcessRunner is not thread safe.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
elharo: We can do that. It is a breaking API change.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
elharo: They do quite different things. It might make sense to move the AccumulatingLineListener into ProcessRunner itself. It's not exosed in the public API (constructor is private)

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
meltsufin: It's true that the Cloud SDK reserves the right to move the files, but until we drop the dependency on AppCfg in the app-engine-java component in the Cloud SDK, we can do the file-based check without making anything more fragile in the library.

patflynn: Staging is in the process of being distributed to Maven Central. We should
be extracting ourselves out of this mess, rather than digging deeper into
it.

pom.xml
```diff
@@ -93,6 +93,11 @@
       <artifactId>jsr305</artifactId>
       <version>3.0.1</version>
     </dependency>
+		<dependency>
+		    <groupId>org.json</groupId>
```
patflynn: I sort of half read the other json related thread you have going. Are we switching this to gson?

elharo: I don't think so. GSON would be dramatic overkill for reading a single value. We could use JSR 353 if the license (GPL, CDDL) isn't a problem.

loosebazooka:
I'm not fully sure what was discussed here, but we're already depending on the `jar` being in the directory. Adding a check doesn't change how we interact with it. Could we should just check the existence for now. Future options include using the published jar for staging in maven central or possibly via a staging command exposed in gcloud (maybe?)

patflynn:
This check though is for whether the java app engine component has been
installed. The contract for determining that through gcloud is by querying
the installed components through the CLI. Using a heuristic when we don't
need one doesn't seem like the right thing to do.

meltsufin:
When we move away from relying on the `jar` in the Cloud SDK, we should reevaluate the detection of the component as well. However, currently, there is no additional risk in using the existence of the `jar` as the marker for the component being installed.

elharo:
There appears to be something really weird going on with the backup gcloud creates on removing the app-engine-java component, at least on Mac OS X. Somehow my terminal gets redirected to look there instead of where it used to be. That explains why Mike thinks the tools jar is removed and I think it isn't. It's actually being moved form point A to point B, and perhaps a link left behind.

patflynn:
@briandealwis 
@elharo 
@etanshaul 
@meltsufin 
@loosebazooka 

After an offline discussion I've been convinced that checking the jar location is actually the best approach for now.  

Getting this behavior into the library is blocking our IJ release (since we want this feature for our next beta). Anybody have bandwidth to start on this today?

PR #208, opened 2016-09-12
more specific exception when app engine java components are not installed

elharo:
@patflynn @akerekes

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -231,9 +231,12 @@ public class CloudSdk {
   }
 
   /**
-   * Checks whether the configured Cloud SDK Path is valid.
+   * Checks whether the Cloud SDK Path with necessary components is installed in
+   * the expected location.
    *
-   * @throws AppEngineException when there is a validation error.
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -253,9 +256,8 @@ public void validate() throws AppEngineException {
               + getDevAppServerPath() + "' is not a file.");
     }
     if (!Files.isDirectory(getJavaAppEngineSdkPath())) {
-      throw new AppEngineException(
-          "Validation Error: Java App Engine SDK location '"
-              + getJavaAppEngineSdkPath() + "' is not a directory.");
+      throw new AppEngineJavaComponentsNotInstalledException(
+          "Validation Error: Java App Engine components not installed");
```
etanshaul: What do you think about adding information to the exception message on how to install it (e.g. 'run gcloud components install ...'). Or do you think its better left to the clients to do this if needed?

etanshaul: On second thought, an exception message probably isn't the best place for this.

elharo:
Yes, the clients will catch this exception and then tell the user what to do (or do it themselves)

PR #209, opened 2016-09-12
Validate Cloud SDK and Java components before use

meltsufin:
@etanshaul @patflynn @loosebazooka @elharo 

I split the validation up into validating Cloud SDK separately from the Java components, and created a separate exception for when Cloud SDK is not found.
I also made the checks happen by default in the methods where the corresponding components are used.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -141,11 +141,8 @@ public class CloudSdk {
    */
   public void runDevAppServerCommand(List<String> args) throws ProcessRunnerException {
     validateCloudSdk();
-    if (!Files.isRegularFile(getDevAppServerPath())) {
-      throw new AppEngineException(
-          "Validation Error: dev_appserver.py location '"
-              + getDevAppServerPath() + "' is not a file.");
-    }
+    // TODO: remove this check when the auto-install for Java is fixed in dev_appserver.py
+    validateAppEngineJavaComponents();
 
     List<String> command = new ArrayList<>();
 
@@ -263,6 +260,11 @@ public class CloudSdk {
     if (!Files.isRegularFile(getGCloudPath())) {
       throw new CloudSdkNotFoundException(
           "Validation Error: gcloud location '" + getGCloudPath() + "' is not a file.");
+    }
+    if (!Files.isRegularFile(getDevAppServerPath())) {
+      throw new CloudSdkNotFoundException(
+          "Validation Error: dev_appserver.py location '"
+              + getDevAppServerPath() + "' is not a file.");
     }
   }
 
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -251,7 +251,7 @@ public class CloudSdk {
    */
   public void validateCloudSdk() throws CloudSdkNotFoundException {
     if (sdkPath == null) {
-      throw new CloudSdkNotFoundException("Validation Error: SDK path is null");
+      throw new CloudSdkNotFoundException("Validation Error: Cloud SDK path is null");
     }
     if (!Files.isDirectory(sdkPath)) {
       throw new CloudSdkNotFoundException(
@@ -279,7 +279,8 @@ public class CloudSdk {
       throws AppEngineJavaComponentsNotInstalledException {
     if (!Files.isDirectory(getJavaAppEngineSdkPath())) {
       throw new AppEngineJavaComponentsNotInstalledException(
-          "Validation Error: Java App Engine components not installed");
+          "Validation Error: Java App Engine components not installed."
+              + " Fix by running 'gcloud components install app-engine-java' on command-line.");
     }
     if (!Files.isRegularFile(JAR_LOCATIONS.get(JAVA_TOOLS_JAR))) {
       throw new AppEngineJavaComponentsNotInstalledException(
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -101,8 +101,11 @@ public class CloudSdk {
    * Uses the process runner to execute the gcloud app command with the provided arguments.
    *
    * @param args The arguments to pass to "gcloud app" command.
+   * @throws CloudSdkNotFoundException when the Cloud SDK is not installed where expected
    */
   public void runAppCommand(List<String> args) throws ProcessRunnerException {
+    validateCloudSdk();
+
     List<String> command = new ArrayList<>();
     command.add(getGCloudPath().toString());
     command.add("app");
@@ -131,10 +134,19 @@ public class CloudSdk {
    * Uses the process runner to execute a dev_appserver.py command.
    *
    * @param args the arguments to pass to dev_appserver.py
-   * @throws InvalidPathException when Python can't be located
-   * @throws ProcessRunnerException when process runner encounters an error
+   * @throws InvalidPathException      when Python can't be located
+   * @throws ProcessRunnerException    when process runner encounters an error
+   * @throws CloudSdkNotFoundException when the Cloud SDK is not installed where expected
+   * @throws AppEngineException        when dev_appserver.py cannot be found
    */
   public void runDevAppServerCommand(List<String> args) throws ProcessRunnerException {
+    validateCloudSdk();
+    if (!Files.isRegularFile(getDevAppServerPath())) {
+      throw new AppEngineException(
+          "Validation Error: dev_appserver.py location '"
+              + getDevAppServerPath() + "' is not a file.");
+    }
+
     List<String> command = new ArrayList<>();
 
     if (IS_WINDOWS) {
@@ -156,8 +168,13 @@ public class CloudSdk {
 
   /**
    * Executes an App Engine SDK CLI command.
+   *
+   * @throws AppEngineJavaComponentsNotInstalledException when the App Engine Java components are
+   *                                                      not installed in the Cloud SDK
    */
   public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
+    validateAppEngineJavaComponents();
+
     // AppEngineSdk requires this system property to be set.
     System.setProperty("appengine.sdk.root", getJavaAppEngineSdkPath().toString());
 
@@ -210,13 +227,13 @@ public class CloudSdk {
         throw new InvalidPathException(cloudSdkPython, "python binary not in specified location");
       }
     }
-    
+
     Path pythonPath = getSdkPath().resolve(WINDOWS_BUNDLED_PYTHON);
     if (Files.exists(pythonPath)) {
       return pythonPath;
     } else {
       return Paths.get("python");
-    } 
+    }
 
   }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -131,10 +134,19 @@ public void runAppCommand(List<String> args) throws ProcessRunnerException {
...
-   * @throws InvalidPathException when Python can't be located
-   * @throws ProcessRunnerException when process runner encounters an error
+   * @throws InvalidPathException      when Python can't be located
+   * @throws ProcessRunnerException    when process runner encounters an error
+   * @throws CloudSdkNotFoundException when the Cloud SDK is not installed where expected
+   * @throws AppEngineException        when dev_appserver.py cannot be found
    */
   public void runDevAppServerCommand(List<String> args) throws ProcessRunnerException {
+    validateCloudSdk();
+    if (!Files.isRegularFile(getDevAppServerPath())) {
```
loosebazooka: perhaps this could be in validateDevAppServer()?

etanshaul: maybe wrap this in a validateDevAppServer() (or something like that) to keep the validation consistent?

meltsufin: Is dev_appserver.py always installed with Cloud SDK? If so, I can roll this into validateCloudSdk().

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -231,36 +248,39 @@ public Path getJarPath(String jarName) {
...
+   * Checks whether the Cloud SDK Path with is valid.
    *
-   * @throws AppEngineJavaComponentsNotInstalledException when the App Engine Java 
-   *     components are not installed in the Cloud SDK
-   * @throws AppEngineException when the Cloud SDK is not installed where expected
+   * @throws CloudSdkNotFoundException when the Cloud SDK is not installed where expected
    */
-  public void validate() throws AppEngineException {
+  public void validateCloudSdk() throws CloudSdkNotFoundException {
     if (sdkPath == null) {
```
patflynn: IT feels like each one of these conditions is different and a client would probably want to handle them differently.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
etanshaul: also are we sure that dev app server doesn't require the java components?

loosebazooka: it's part of the appengine components

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkNotFoundException.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package com.google.cloud.tools.appengine.cloudsdk;
+
+import com.google.cloud.tools.appengine.api.AppEngineException;
+
+/**
+ * User needs to use a valid Cloud SDK installation.
+ */
+public class CloudSdkNotFoundException extends AppEngineException {
```
patflynn: I'm not seeing a lot of value in the AppEngineException class of exceptions anymore.  It seems to be used by all exceptions that could possibly occur in the library rather than the unexpected errors caused by library deps.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -131,10 +134,19 @@ public void runAppCommand(List<String> args) throws ProcessRunnerException {
...
-   * @throws ProcessRunnerException when process runner encounters an error
+   * @throws InvalidPathException      when Python can't be located
+   * @throws ProcessRunnerException    when process runner encounters an error
+   * @throws CloudSdkNotFoundException when the Cloud SDK is not installed where expected
+   * @throws AppEngineException        when dev_appserver.py cannot be found
    */
   public void runDevAppServerCommand(List<String> args) throws ProcessRunnerException {
+    validateCloudSdk();
+    if (!Files.isRegularFile(getDevAppServerPath())) {
+      throw new AppEngineException(
```
etanshaul: Should this have a specific exception subclass like the others?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
meltsufin: Ss the app-engine-java component necessary and sufficient for running Java apps with dev_appserver.py?

patflynn: the dev-appserver is always included with the cloud sdk, the
app-engine-java component is not required for running the dev-appserver,
but apparently you need to invoke it correctly so that it'll auto-install
the java component when it detects a java runtime is needed.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
meltsufin: I think for now we just want to let clients know if the Cloud SDK is valid or not. Details are in the exception message.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkNotFoundException.java
meltsufin: It could be useful for a client that just needs to know broadly if there is a problem coming from the lib.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
meltsufin: We should just do the auto-install flag then.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -156,8 +168,13 @@ public void runDevAppServerCommand(List<String> args) throws ProcessRunnerExcept
 
   /**
    * Executes an App Engine SDK CLI command.
+   *
+   * @throws AppEngineJavaComponentsNotInstalledException when the App Engine Java components are
+   *                                                      not installed in the Cloud SDK
    */
   public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
```
etanshaul: shouldn't this method also `validateCloudSdk()` ?

meltsufin: Technically, the check for the existence of the jar is sufficient.

etanshaul: Yeah, that's true.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkNotFoundException.java
elharo: Might be a good time to change to a checked exception then since these are all environmental problems, not programming errors.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
patflynn: That's not useful for the IDE use-cases. We want to present actionable
errors to the user and we don't want to pass through your message strings
since we support localization.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
meltsufin: Found out that auto-install currently doesn't work for Java runtime.

etanshaul:
lgtm

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkNotFoundException.java
meltsufin: I still would rather do it in a separate PR, since it's would be more disruptive to clients.

elharo: I could wait on making the old AppEngineException checked, but this one should be checked form the get go since the point of this PR is that clients should be catching and handling this exception. Otherwise it's really not useful.

The alternative would be to rethink throwing the exception and have it return a boolean instead.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkNotFoundException.java
```diff
@@ -0,0 +1,30 @@
...
+
+  CloudSdkNotFoundException(String message) {
```
etanshaul: Please make this public as well per our offline discussion (so we can mock these exceptions in client unit tests).

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkNotFoundException.java
meltsufin: I agree that making the exception checked is the right approach here. I just want to make that change separate because it affects a lot of places in the code.

meltsufin:
@etanshaul can you take a look one more time before I merge?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -279,7 +279,8 @@ public void validateAppEngineJavaComponents()
       throws AppEngineJavaComponentsNotInstalledException {
     if (!Files.isDirectory(getJavaAppEngineSdkPath())) {
       throw new AppEngineJavaComponentsNotInstalledException(
-          "Validation Error: Java App Engine components not installed");
+          "Validation Error: Java App Engine components not installed."
+              + " Fix by running 'gcloud components install app-engine-java' on command-line.");
```
etanshaul: Didn't we say we wouldn't suggest fixes here in the exception message, but instead let the clients determine how to handle it? One reason would be if clients auto download the component, it might not make sense to have this here.

etanshaul:
Other than my comment on the updated exception message, lgtm.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
etanshaul: I am ok with it though - I don't see it as a huge issue. So feel free to merge once you decide what you want to do it it. Also please release the lib afterwards :)

meltsufin: I think when the auto-download is enabled, this exception simply won't show up. My motivation for including the fix in the message is to just short-circuit the resolution of the problem for users, if the library clients somehow fail to handle the exception correctly or decide to simply display the message contained in the exception.

elharo: I'm with @etanshaul on this one. This library should be quite agnostic about whether there's a user or not.

meltsufin: I would revisit this when these exceptions are made checked. For the time being, however, since the clients are not forced to handle the exception, I'd rather have this in there to potentially save time for users encountering this issue.

patflynn: LGTM

PR #212, opened 2016-09-14
make AppEngineException a checked exception

### LINKED ISSUE #84: Make AppEngineException checked
This indicates an unpredictable environmental condition, not a program bug. This is exactly what checked exceptions were designed for.

elharo:
@meltsufin fix #84

FYI, making this change immediately identified one bug in the Eclipse Plugin by turning it into a compile error.


elharo:
CliProcessManagerTest is failing on travis, not sure why. It passes locally.

patflynn:
Let's have a chat about this. 

I usually like reserving checked exceptions for meaningful conditions that the client can handle.  I thought the purpose of the AEException is to wrap unexpected errors that arose from calling APIs or gcloud.

elharo:
Checked exceptions are for environmental problems outside the control of the program that the client must handle because there's no way to detect them at compile time. E.g. an IOException because a socket times out.

Runtime exceptions are for bug in program code that could be detected with an appropriate source code change. e.g. null pointers.

java.lang.Errors are for things that really can't be handled at all. E.g out of memory.

patflynn:
I broadly agree with your categories. What I would add is that your checked
exception taxonomy needs to enable clients to react in a useful way to
those exceptions.

But what I'm seeing here is the AppEngineException being used in a broader
more generic way that isn't particularly useful to clients.  There are
instances in this PR where AppEngineException is being used as a catchall
rather than an indication that an App Engine transient service error
occured ( such as in discoverSdkPath() ).   Another example is public
CloudSdk build() throws AppEngineException where an AppEngineException is
used to indicate a validation issue rather than a transient error.

The pattern I often see is a library or module defining a base checked
exception and wrapping all error sources with it as a rote part of their
API.  This adds no value to clients and I want to avoid AppEngineException
sharing that fate.

Perhaps if we made the name more explicit about the category of errors it
represents? Which based off of the javadoc  is essentially some kind of
AppEngine service transient error.  Does that sound reasonable?

elharo:
We could split out the CloudSDK errors into a separate hierarchy. E.g. AppEngineException (or CloudException?) for problems that involve talking to the servers and CloudSDKExcepition for problems that involve a local install?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -480,7 +486,7 @@ public CloudSdk build() {
      * @throws AppEngineException if not found
      */
     @Nonnull
-    private Path discoverSdkPath() {
+    private Path discoverSdkPath() throws AppEngineException {
```
aslo: Not sure if this is outside the scope of this PR, but it would be nice if there were a more specific checked exception we could use here - something like `CloudSdkNotFoundException`

aslo: Just read the other comments on this PR.  FWIW, more generally I agree with the sentiment expressed in the main thread - more explicit categories for checked exceptions would really nice.

PR #224, opened 2016-09-28
Remove unnecessary code.

joaoandremartins:
See https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/937 on why this isn't needed anymore.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
@@ -98,17 +98,6 @@ public class CloudSdkAppEngineFlexibleStaging implements AppEngineFlexibleStagin
         Path destination = config.getStagingDirectory().toPath()
             .resolve(config.getArtifact().toPath().getFileName());
         Files.copy(config.getArtifact().toPath(), destination, REPLACE_EXISTING);
-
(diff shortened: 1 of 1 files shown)


patflynn:
Isn't this still an issue for people who specify their own dockerfile?

meltsufin:
Normally, the custom dockerfile users provide is based on one of the standard ones. So, it shouldn't be an issue for them.

patflynn:
Yeah but what about not normally?  Has anybody looked at the Docker issue
recently to see if they have a global fix for this coming?

meltsufin:
It doesn't look like it's a Docker issue. My understanding is that it's just an issue with Jetty running under the jetty user that doesn't have access to the files.

patflynn:
https://github.com/jboss-dockerfiles/wildfly/issues/19

patflynn:
Based off of that issue it seems like the right thing to do until it's
fixed is to continue to change the permission on the archive on the host
before we add it to the container.

meltsufin:
We're using the first workaround mentioned in that issue.

loosebazooka:
I still don't understand why this was triggered on windows though.

patflynn:
You are doing the 1st workaround by modifying the image.  We can solve this
problem for all our users by making sure we modify the file first.

patflynn:
As a possible explanation, I found this: https://github.com/docker/docker/issues/783.  If I were to hazard a guess, perhaps the host OS for Flex did not have the AUFS fix earlier this year when we did our initial testing.

PR #228, opened 2016-09-29
Tests : Wait for stderr listener to complete

loosebazooka:


loosebazooka:
@meltsufin should fix that test..

meltsufin:
Do we understand why this is happening? What's the rationale for 1000ms wait vs. some other number?

loosebazooka:
It's because the test thread ends before the "stderr" reading thread ends. It's a race condition in the test specifically. In normal use, stderr reading would just continue until it finishes writing to some console or window.

loosebazooka:
as for 1000 vs any other number, no real reason. I think 1000 is safe for this test though.

loosebazooka:
tests were also in the wrong package, so I moved those.

PR #229, opened 2016-09-29
Make JAVA_HOME configurable on the devappserver child process

aslo:
This was motivated by https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/927. These changes allow callers to specify a java_home directory in the devserver `RunConfiguration`. 

Note that up until now, RunConfiguration has been used to only specify arguments to `dev_appserver.py`. This change would be the first environment variable that is added to the RunConfiguration interface - I think it's kind of nice to abstract that from the user a bit; I'm interested to hear other thoughts on that.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
@@ -46,6 +46,7 @@ public class DefaultRunConfiguration implements RunConfiguration {
   private String devAppserverLogLevel;
   private Boolean skipSdkUpdateCheck;
   private String defaultGcsBucketName;
+  private String javaHomeDir;
 
   @Override
   public List<File> getAppYamls() {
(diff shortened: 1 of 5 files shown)


patflynn:
sorry. didn't mean to unassign everyone there. Some weird UI interaction with github.

patflynn:
@joaoandremartins @etanshaul

joaoandremartins:
Have you tried running dev_appserver with --dev_appserver_log_level debug and checking that the called java binary is indeed the {$JAVA_HOME}/bin/java one?

briandealwis:
@joaoandremartins it does work — it's the only workaround we found for the Java8/JSP problem https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/590#issuecomment-241177482

joaoandremartins:
Oh, OK. It should work, though I tried calling dev_appserver.py from a shell where JAVA_HOME was set and the wrong java binary was still being called. i.e., java from PATH, not JAVA_HOME.

briandealwis:
Are there any other environment variables that may make sense to allow the user to modify?  I guess any other environment variables should be put in the `appengine-web.xml` or `app.yaml`?

aslo:
@joaoandremartins From my testing, the JAVA_HOME environment variable was being respected. Did you `export` it when setting the variable for testing? 

@briandealwis I'm not sure of any other env variables that it makes sense to add right now.  If we ever have a need, we can always add more using this mechanism in the future.

joaoandremartins:
Ah, I might have forgotten to export yes.
LGTM

PR #231, opened 2016-10-06
update dependency scopes

loosebazooka:
@meltsufin @patflynn @elharo 

I think we had misconfigured some scopes earlier.
I'm pretty sure about the junit scope change
findbugs to provided seems to be what the internet [says](http://stackoverflow.com/questions/26867614/what-is-correct-maven-scope-of-findbugs-annotations) is needed, however this I'm a lot less sure about. Findbugs annotation are only at compile time right, they're not preserved into runtime?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,6 +76,7 @@
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>4.12</version>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>com.google.guava</groupId>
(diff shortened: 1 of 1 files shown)


meltsufin:
This looks good to me. I think either `provided` or `test` is fine for findbugs.

loosebazooka:
This will also affect transitive dependencies in your project. You might need to add junit back into the test compile scope for projects that depends on this. Certainly appengine-gradle-plugin does.

meltsufin:
None of our projects should depend on JUnit in the compile scope. appengine-maven-plugin already specifies it in the `test` scope.

loosebazooka:
Yeah I mean it came in as a transitive dependency, so if you're building with gradle, it might have snuck into your compile dependencies without you knowing. I'm not sure if transitive is the default behavior in maven though.

PR #234, opened 2016-10-11
Add environment variable on Windows to set file upload processes to 1

joaoandremartins:
This is to fix https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/985 while the definite fix isn't in place.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -148,6 +148,12 @@ public class CloudSdk {
     if (appCommandMetricsEnvironmentVersion != null) {
       environment.put("CLOUDSDK_METRICS_ENVIRONMENT_VERSION", appCommandMetricsEnvironmentVersion);
     }
+    // This is to ensure IDE credentials get correctly passed to the gcloud commands, in Windows.
+    // It's a temporary workaround until a fix is released in b/31683722.
+    // https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/985
+    if (System.getProperty("os.name").contains("Windows")) {
+      environment.put("CLOUDSDK_APP_NUM_FILE_UPLOAD_PROCESSES", "1");
+    }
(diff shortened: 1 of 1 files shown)


joaoandremartins:
This is the fix that I have in mind, though it's being a pain to test this on Windows...

etanshaul:
lgtm assuming the lib guys are ok with it.

just wondering - why is it a pain to test on windows? Could you just use this change with IntelliJ (or raw gcloud) and see if the deployment works on windows?

joaoandremartins:
Just tested this with Windows, it works.

It was a pain to test on Windows because, for some reason, on my GCE VM, the runIdea gradle goal isn't even running and it's also a bit slow.

Also, I had to use the debugger to step in the library code and manually run the environment.put() command, which isn't a great way to run it. Let me know if you know better ways.

Testing it from a command line wouldn't work, since we wouldn't be testing the code that sends gcloud the IDEA credentials.

etanshaul:
A better way would be to have the updated core lib locally, update the pom
to some random version, do a mvn install to get it in the local repo, then
update Intellij to whatever version you chose.

etanshaul:
oh, and you need to build.gradle (the top level one) to also have `mavenLocal()` under the subprojects repositories section so that it pulls from your local mvn repo

joaoandremartins:
Yeah, I agree that would have been a better way. But both the VM and the laptop are really slow. I suspect checking out the library, adding the local repo, pointing to it and then debugging would've easily taken me another hour.

etanshaul:
sgtm

patflynn:
Shouldn't we only do this for standard deployment? Can you ask Zach what effect this will have globally?

joaoandremartins:
Just did. Though I would expect that the credentials aren't being properly passed down on every gcloud command, not just standard deployment. The reason we found this bug is because we were using accounts that couldn't write to GCS, which Flex probably doesn't require.

joaoandremartins:
@patflynn I think Zach left for the day -- let me know if you want me to https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/997 or not

PR #238, opened 2016-10-14
Adds support for '--name=true/false' flags.

joaoandremartins:
Required for the dev_appserver automatic_restart flag, which is set to `true` by default. Setting this flag explicitly to false is required.

Helps in https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/972

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer.java
@@ -88,7 +88,7 @@ public class CloudSdkAppEngineDevServer implements AppEngineDevServer {
     arguments.addAll(DevAppServerArgs.get("runtime", config.getRuntime()));
     arguments.addAll(DevAppServerArgs.get("allow_skipped_files", config.getAllowSkippedFiles()));
     arguments.addAll(DevAppServerArgs.get("api_port", config.getApiPort()));
-    arguments.addAll(DevAppServerArgs.get("automatic_restart", config.getAutomaticRestart()));
+    arguments.addAll(DevAppServerArgs.get("automatic_restart", config.getAutomaticRestart(), true));
     arguments
         .addAll(DevAppServerArgs.get("dev_appserver_log_level", config.getDevAppserverLogLevel()));
     arguments.addAll(DevAppServerArgs.get("skip_sdk_update_check", config.getSkipSdkUpdateCheck()));
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer.java
@@ -88,7 +88,7 @@ public class CloudSdkAppEngineDevServer implements AppEngineDevServer {
     arguments.addAll(DevAppServerArgs.get("runtime", config.getRuntime()));
     arguments.addAll(DevAppServerArgs.get("allow_skipped_files", config.getAllowSkippedFiles()));
     arguments.addAll(DevAppServerArgs.get("api_port", config.getApiPort()));
-    arguments.addAll(DevAppServerArgs.get("automatic_restart", config.getAutomaticRestart(), true));
+    arguments.addAll(DevAppServerArgs.get("automatic_restart", config.getAutomaticRestart()));
     arguments
         .addAll(DevAppServerArgs.get("dev_appserver_log_level", config.getDevAppserverLogLevel()));
     arguments.addAll(DevAppServerArgs.get("skip_sdk_update_check", config.getSkipSdkUpdateCheck()));
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/Args.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/Args.java
@@ -97,48 +97,6 @@ class Args {
   }
 
   /**
-   * Produces the flag form of a boolean value.
-   *
-   * @return {@code [--name]} if value is {@code true}, {@code [--no-name]} if value is
-   * {@code false}, {@code []} if value is {@code null}.
-   */
-  static List<String> boolWithNo(String name, Boolean value) {
-    if (value != null) {
-      if (value) {
-        return Collections.singletonList("--" + name);
-      }
-      return Collections.singletonList("--no-" + name);
-    }
-    return Collections.emptyList();
-  }
-
-  /**
-   * Produces the flag form of a boolean value.
-   *
-   * @return {@code [--name=true]} if value=true, {@code [--name=false]} if value=false,
-   *     {@code []} if value=null.
-   */
-  static List<String> boolWithTrueFalse(String name, Boolean value) {
-    if (value == null) {
-      return Collections.emptyList();
-    }
-    return Arrays.asList("--" + name + "=" + value.toString());
-  }
-
-  /**
-   * Produces the flag form of a boolean value.
-   *
-   * @return {@code [--name]} if value is {@code true}, {@code []} if value is {@code false} or
-   * {@code null}.
-   */
-  static List<String> bool(String name, Boolean value) {
-    if (Boolean.TRUE.equals(value)) {
-      return Collections.singletonList("--" + name);
-    }
-    return Collections.emptyList();
-  }
-
-  /**
    * Produces the flag form of a file object, using {@link File#toPath()}.
    *
    * @return {@code [--name, file.toPath().toString()]} or {@code []} if file is null.
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/Args.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/Args.java
@@ -112,6 +112,13 @@ class Args {
     return Collections.emptyList();
   }
 
+  static List<String> boolWithTrueFalse(String name, Boolean value) {
+    if (value == null || !value) {
+      return Arrays.asList("--" + name + "=false");
+    }
+    return Arrays.asList("--" + name + "=true");
+  }
+
   /**
    * Produces the flag form of a boolean value.
    *
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/Args.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/Args.java
@@ -112,11 +112,17 @@ class Args {
     return Collections.emptyList();
   }
 
+  /**
+   * Produces the flag form of a boolean value.
+   *
+   * @return {@code [--name=true]} if value=true, {@code [--name=false]} if value=false,
+   *     {@code []} if value=null.
+   */
   static List<String> boolWithTrueFalse(String name, Boolean value) {
-    if (value == null || !value) {
-      return Arrays.asList("--" + name + "=false");
+    if (value == null) {
+      return Collections.emptyList();
     }
-    return Arrays.asList("--" + name + "=true");
+    return Arrays.asList("--" + name + "=" + value.toString());
   }
 
   /**
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/DevAppServerArgs.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/DevAppServerArgs.java
@@ -45,21 +45,10 @@ public class DevAppServerArgs {
   }
 
   /**
-   * @return {@code [--name]} if value=true, {@code []} if value=false/null.
+   * @return {@code [--name=true]} if value=true, {@code [--name=false]} if value=false,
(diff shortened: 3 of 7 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/DevAppServerArgs.java
```diff
@@ -24,30 +24,42 @@
...
   /**
-   * @return [--name] if value=true, [] if value=false/null.
+   * @return {@code [--name]} if value=true, {@code []} if value=false/null.
    */
   public static List<String> get(String name, Boolean value) {
+    return get(name, value, false);
+  }
+
+  /**
+   * @return if produceTrueFalse=true, {@code [--name=true]} if value=true, {@code [--name=false]}
```
elharo: This argument confuses me.

patflynn: These methods give me the heebeegeebeez.   Are there  any downsides to simply always returning a flag value whether the boolean argument is true or false?

joaoandremartins: For boolean flags, I kinda prefer either having a flag or no flag at all, for true or false.

@elharo would renaming it to "withTrueOrFalse" make it any better?

patflynn: @joaoandremartins but the issue with that is that defaults are not always 'false'. Which means the caller has to know which one of these methods to call based on the defaults in gcloud.  That seems like the wrong thing to do.

joaoandremartins: Indeed. On the flipside, there might also be commands that don't support --flag=false/true notation (I haven't checked).
I can either add to this new method's Javadoc that this method should only be used when a boolean flag default is true, or we can file a new issue to investigate getting rid of the method that produces either --flag or nothing.

patflynn: @joaoandremartins the current API behavior was non-obvious and caused an important bug, but with this change we just add to the complexity.  Let's fix this properly now. It's not a lot of work to look at the devappserver flags to see if there exists a flag that are 'existence' based rather than true 'boolean'.  We should also note other boolean flags that default to true, as they are all broken as of now.

patflynn: @elharo 
@meltsufin 
@loosebazooka 
@briandealwis 
@GoogleCloudPlatform/cloud-tools-for-java 

Making the change as I propose it would effect all clients. I'm including Eclipse and build plugin team members too to get their input.

elharo: I will reiterate: I am very confused by the PR. I am not sure what it's trying to do or how it's trying to accomplish that, but surely there's a less confusing way to get where we want to be?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/DevAppServerArgs.java
```diff
@@ -24,30 +24,42 @@
 public class DevAppServerArgs {
 
   /**
-   * @return [--name=value] or [] if value=null.
+   * @return {@code [--name=value]} or {@code []} if value=null.
```
elharo: returns an empty list, not an empty array

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/DevAppServerArgs.java
```diff
@@ -24,30 +24,42 @@
...
    */
   public static List<String> get(String name, String value) {
     return Args.stringWithEq(name, value);
   }
 
   /**
-   * @return [--name=value1, --name=value2, ...] or [] if value=null.
+   * @return {@code [--name=value1, --name=value2, ...]} or {@code []} if value=null.
```
elharo: returns an empty list, not an empty array

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/DevAppServerArgs.java
```diff
@@ -24,30 +24,42 @@
...
-   * @return [--name=value] or [] if value=null.
+   * @return {@code [--name=value]} or {@code []} if value=null.
    */
   public static List<String> get(String name, Integer value) {
     return Args.integerWithEq(name, value);
   }
 
   /**
-   * @return [--name] if value=true, [] if value=false/null.
+   * @return {@code [--name]} if value=true, {@code []} if value=false/null.
```
elharo: returns an empty list, not an empty array

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/DevAppServerArgs.java
```diff
@@ -24,30 +24,42 @@
...
-   * @return [--name=value1, --name=value2, ...] or [] if value=null.
+   * @return {@code [--name=value1, --name=value2, ...]} or {@code []} if value=null.
    */
   public static List<String> get(String name, List<String> values) {
     return Args.stringsWithEq(name, values);
   }
 
   /**
-   * @return [--name=value] or [] if value=null.
+   * @return {@code [--name=value]} or {@code []} if value=null.
```
elharo: returns an empty list, not an empty array

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/DevAppServerArgs.java
```diff
@@ -24,30 +24,42 @@
...
   public static List<String> get(String name, Boolean value) {
+    return get(name, value, false);
+  }
+
+  /**
+   * @return if produceTrueFalse=true, {@code [--name=true]} if value=true, {@code [--name=false]}
+   *     if value=false/null. If produceTrueFalse=false, {@code [--name]} if value=true, {@code []}
+   *     if value=false/null.
+   */
+  public static List<String> get(String name, Boolean value, Boolean produceTrueFalse) {
```
elharo: This is the method where you lose me. What if we eliminated the use of java.lang.Boolean and null values? All values must be explicitly specified.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/DevAppServerArgs.java
joaoandremartins: @elharo I don't know how to be clearer here :(. We're adding library support for boolean flags of the kind `--flag=false `.

Perhaps if you tell me if anything specific is generating that confusion, I can be more helpful.

@patflynn as you mentioned, this change is indeed an API change, which goes way beyond the scope of what this PR was supposed to be. Creating a new issue doesn't mean the issue can't be worked on now -- in practice, the same thing will happen in both cases, but without this code committed, and the dev_appserver reload fix will have to wait for that work to be finished in consequence. As you wish, though!

patflynn: I'm confused.  Why can't you leave the API unchanged, and always return --name=false or --name=true?

joaoandremartins: Actually, scratch that bit. The API wouldn't be changed, just the behaviour of the method, but that's a separate issue.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer.java
```diff
@@ -88,7 +88,7 @@ public void run(RunConfiguration config) throws AppEngineException {
     arguments.addAll(DevAppServerArgs.get("runtime", config.getRuntime()));
     arguments.addAll(DevAppServerArgs.get("allow_skipped_files", config.getAllowSkippedFiles()));
     arguments.addAll(DevAppServerArgs.get("api_port", config.getApiPort()));
-    arguments.addAll(DevAppServerArgs.get("automatic_restart", config.getAutomaticRestart()));
+    arguments.addAll(DevAppServerArgs.get("automatic_restart", config.getAutomaticRestart(), true));
```
joaoandremartins: The way I see it, the real issue is fundamentally `--automatic_restart` being `true` by default in dev_appserver. My preferred fix would be to just default it to false on dev_appserver and move on, but that change would take longer. (Manifestly, I don't know why automatic restart defaults to true, or can think of any case where that behaviour is helpful)

That being said, I like that we're making up in the library for that issue in dev_appserver, and prefer we stick to default false.

Regarding your first point, checking the state of the other dev_appserver flags seems to be a popular request these days, so I'll go ahead and do that in the scope of this PR.

patflynn: @chanseokoh I think you're right that we should be setting those defaults. @meltsufin do you know if these defaults are set elsewhere?

```
paflynn@paflynn1:~/google-cloud-sdk/bin $ dev_appserver.py -h | grep -B 5 'default: True'
                        password to use when connecting to the SMTP server
                        specified in --smtp_host and --smtp_port (default: )
  --smtp_allow_tls [SMTP_ALLOW_TLS]
                        Allow TLS to be used when the SMTP server announces
                        TLS support (ignored if --smtp_host is not set)
                        (default: True)
--
                        clear the search indexes (default: False)

Task Queue API:
  --enable_task_running [ENABLE_TASK_RUNNING]
                        run "push" tasks created using the taskqueue API
                        automatically (default: True)
--
                        (default: False)
  --api_port API_PORT   port to which the server for API calls should bind
                        (default: 0)
  --automatic_restart [AUTOMATIC_RESTART]
                        restart instances automatically when files relevant to
                        their module are changed (default: True)
```

joaoandremartins: Should I go ahead and set defaults in this PR too?

patflynn:
FYI we had an offline discussion where we established that the existing behavior exists because AppCfg only supports existence/omission boolean flag types. We've agreed that we should change the dev appserver flags parser to pass through 'false' flags.  We also decided not to set defaults in the libraries since they are big 'B' and passing null will do the right thing. 

These are the affected flags:
smtp_allow_tls
enable_task_running 
automatic_restart

We only set automatic_restart in IJ, Eclipse team should review the user of these flags, and I'll also check the Gradle plugin.  Maven leaves unset values as null so is not affected.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer.java
joaoandremartins: Actually, since a null automaticRestart value still produces no flag and uses the default (i.e., keeps the previous behaviour), I'm going to punt setting the defaults to another PR.
This PR remains "add support for --name=true/false flags"

chanseokoh:
Eclipse doesn't set automatic_restart, so we are good. (Seems like the core library doesn't have APIs to set the other two flags.)

I agree we don't have to set the default of automatic_restart in `DefaultRunConfiguration`, because, as before, the last commit ensures that `null` will omit the flag entirely.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/DevAppServerArgs.java
joaoandremartins: This is the norm in every other similar comment for this, and 3 other classes. I agree we should improve these comments, but how do you feel about doing this in another PR?

joaoandremartins:
@patflynn @meltsufin @elharo @chanseokoh PTAL

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/Args.java
```diff
@@ -115,6 +115,19 @@
   /**
    * Produces the flag form of a boolean value.
    *
+   * @return {@code [--name=true]} if value=true, {@code [--name=false]} if value=false,
+   *     {@code []} if value=null.
```
elharo: I still don't understand why we support nulls here. Just use a primitive boolean that is explicitly true or false, never null; per Effective Java Item 49

joaoandremartins:
@meltsufin want to go over this just once more? Didn't need to move any unit tests in the end, since there were none to Args directly.

joaoandremartins:
...or @patflynn ?

patflynn:
@joaoandremartins 
looks like you need the lgtm from @elharo

joaoandremartins:
@elharo and I discussed and I think we're on the same page regarding Boolean vs boolean.

The review I'm requesting is regarding the last commit where I extracted the bool generation logic from Args to the other *Args classes.

PR #242, opened 2016-10-27
Implement cloudSdk.getVersion() and cloudSdk.getComponents()

### LINKED ISSUE #241: CloudSdk.getVersion, CloudSdk.getComponents
CT4J would like to check the installed version of the Cloud SDK (131, 132, etc.) as well as the version s of various components. I'm not immediately sure what this feature should look like but it's probably part of the CloudSdk class.

aslo:
fixes #241 

This PR implements 2 new methods: `cloudSdk.getVersion()` and `cloudSdk.getComponents()`.  Unlike other methods in the `CloudSdk` class that shell out to a subprocess, these methods do not use the CloudSdk's processRunner that is configured at construction time. This allows the methods to have far simpler interfaces and return types. 

PTAL and let me know if this is the right approach. My main concern is that as a user of the CloudSdk class, it's a little weird that you might configure a processRunner at class construction time, but a totally separate one might be used for certain operations.


src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/StringBuilderProcessOutputLineListener.java
```diff
@@ -0,0 +1,38 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.appengine.cloudsdk.internal.process;
+
+import com.google.cloud.tools.appengine.cloudsdk.process.ProcessOutputLineListener;
+
+public class StringBuilderProcessOutputLineListener implements ProcessOutputLineListener {
```
joaoandremartins: I think we want to delegate creating listeners like this to the clients.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -235,6 +245,59 @@ public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
     processRunner.run(command.toArray(new String[command.size()]));
   }
 
+  /**
+   * Returns the version of the Cloud Sdk installation. Unlike other methods in this class that call
```
elharo: SDK

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -235,6 +245,59 @@ public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
...
+
+    SynchronousOutputProcessRunner runner = new SynchronousOutputProcessRunner.Builder().build();
+
+    // gcloud info --format="value(basic.version)"
+    List<String> command = Lists.newArrayList(
+        getGCloudPath().toString(), "info");
+
+    command.addAll(GcloudArgs.get("format", "value(basic.version)"));
+
+    // TODO handle error exit codes
```
elharo: shoul be done in this PRd

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -235,6 +245,59 @@ public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
...
+    runner.run(command.toArray(new String[command.size()]));
+
+    return runner.getProcessStdOut();
+  }
+
+  /**
+   * Returns the list of CloudSdkComponents and their settings, reported by the current gcloud
```
elharo: Cloud SDK Components (see https://cloud.google.com/sdk/docs/managing-components for spelling)

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -235,6 +245,59 @@ public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
...
+   * installation. Unlike other methods in this class that call gcloud, this method always uses a
+   * synchronous ProcessRunner and will block until the gcloud process returns.
+   *
+   * @throws ProcessRunnerException when process runner encounters an error
+   * @throws JsonSyntaxException    when the cloud SDK output cannot be parsed
```
elharo: one space before when

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -235,6 +245,59 @@ public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
...
+    SynchronousOutputProcessRunner runner = new SynchronousOutputProcessRunner.Builder().build();
+
+    // gcloud components list --show-versions --format=json
+    List<String> command = Lists.newArrayList(
+        getGCloudPath().toString(), "components", "list");
+
+    command.addAll(GcloudArgs.get("show-versions", true));
+    command.addAll(GcloudArgs.get("format", "json"));
+
+    // TODO handle error exit codes
```
elharo: do now

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponent.java
```diff
@@ -0,0 +1,116 @@
...
+
+  public String getName() {
+    return name;
+  }
+
+  public void setName(String name) {
+    this.name = name;
+  }
+
+  public String getCurrentVersionString() {
```
elharo: I don't think you need to include the type in the method names. E.g. getLatestVersion instead of getLatestVersionString

joaoandremartins:
I'm not convinced by not following the same approach here as for the other commands. i.e., not calling something like `runGcloudCommand(args, "");`. That takes care of all the processRunner plumbing for you, why not use it?

I don't think you need to specify a Configuration interface either -- other run*Command() methods are already adding custom flags within the method's body, you could do the same with --version.

I think it's fair to have a public getVersion() method in the CloudSdk object, I don't see any advantage in creating a new object for the operation as we do for every other operation (e.g., CloudSdkAppEngineDeployment).

Another thing I'm unsure, but I don't think we have a lot of workarounds, is that we're making it clear that the library runs on top of gcloud, which we want to hide as much as possible. If instead of gcloud CLI we had a Java API, how would we check the API version? We probably wouldn't need to, as that would be taken care of by e.g. Maven... So I can't really see a better way for handling this than what you're doing...

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
joaoandremartins: Actually, I think this should be handled in the same way as the other commands, i.e., add exit hooks and let the client handle that any way they want. This is related to my higher level comment on the PR main thread.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponent.java
```diff
@@ -0,0 +1,116 @@
...
+ */
+
+package com.google.cloud.tools.appengine.cloudsdk.serialization;
+
+import com.google.gson.annotations.SerializedName;
+
+/**
+ * Representation of a CloudSdkComponent. Used for json serialzation/deserialization.
+ */
+public class CloudSdkComponent {
```
joaoandremartins: Would it make sense for this to be an enum?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/StringBuilderProcessOutputLineListener.java
joaoandremartins: I'm not sure I follow. the client should just need to call sdk.getVersion() and nothing else. Likewise for getComponents() (as I think it's happening now).
If you really want to use this, I think the class should not be public but package-protected (likewise for WaitingProcessOutputLineListener). But I don't feel strongly about this.

aslo:
> I'm not convinced by not following the same approach here as for the other commands. i.e., not calling something like runGcloudCommand(args, "");. That takes care of all the processRunner plumbing for you, why not use it?

The main reason is that it is *too* flexible to be useful here.  It's possible for a user of the CloudSdk class to configure an asyc processRunner that makes it difficult for us to also handle the json parsing of the response.  

>I don't think you need to specify a Configuration interface either -- other run*Command() methods are already adding custom flags within the method's body, you could do the same with --version.
> 
>I think it's fair to have a public getVersion() method in the CloudSdk object, I don't see any advantage in creating a new object for the operation as we do for every other operation (e.g., CloudSdkAppEngineDeployment).

I'm not sure I understand what you mean, can you clarify?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
aslo: :+1:

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponent.java
joaoandremartins: At the same time, it's not possible to have more than any one component, the components don't change that often, and you can always catch those JsonParseExceptions so they're not thrown.
How does a class allow for more flexibility?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -235,6 +245,59 @@ public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
...
+
+  /**
+   * Returns the list of CloudSdkComponents and their settings, reported by the current gcloud
+   * installation. Unlike other methods in this class that call gcloud, this method always uses a
+   * synchronous ProcessRunner and will block until the gcloud process returns.
+   *
+   * @throws ProcessRunnerException when process runner encounters an error
+   * @throws JsonSyntaxException    when the cloud SDK output cannot be parsed
+   */
+  public List<CloudSdkComponent> getComponents()
```
joaoandremartins: Should be a Set since components are unique?

joaoandremartins:
Why does asynchrony make parsing the Json harder? Apart from asynchrony, what other issues are there with reusing the existing model?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponent.java
```diff
@@ -0,0 +1,116 @@
...
+
+  public String getLatestVersionString() {
+    return latestVersionString;
+  }
+
+  public void setLatestVersionString(String latestVersionString) {
+    this.latestVersionString = latestVersionString;
+  }
+
+  public Integer getSizeInBytes() {
```
aslo: I think this is more of a stylistic thing and I prefer to keep it in the method name. (It saves users a trip to the javadoc and makes it harder to make a unit conversion error).

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
aslo: I'd prefer to keep it as a list. Technically they aren't guaranteed to be unique. I'd like to avoid making that assumption if possible.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponent.java
aslo: Because if new components are added, we will be able to deserialize them into this class without making any code changes. Even if they don't change that often, I really like having the guarantee that parsing won't break when they do.

You can certainly catch JsonParseExceptions, but it would still be better to be able to just return an actual object right?

elharo:
Haven't followed all the comments, but I do vote for synchrony here. Not sure why you'd want this to be asynchronous.

patflynn:
can we remove the component checking entirely?  We don't actually have a current use case for that.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
briandealwis: The JSON itself is an array so a `List` seems right.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponent.java
briandealwis: I'm :+1: to an object: having to muck about inside JSON objects is not attractive.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponent.java
```diff
@@ -0,0 +1,116 @@
...
+
+  public State getState() {
+    return state;
+  }
+
+  public void setState(State state) {
+    this.state = state;
+  }
+
+  static class State {
```
briandealwis: In keeping with objectifying the JSON, it would be useful to either turn this into an enum or have a predicate method `isInstalled()`.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponent.java
```diff
@@ -0,0 +1,116 @@
...
+
+  public void setIsHidden(Boolean isHidden) {
+    this.isHidden = isHidden;
+  }
+
+  public State getState() {
+    return state;
+  }
+
+  public void setState(State state) {
```
briandealwis: Do we need setters on this object?

aslo:
@elharo Does the eclipse plugin currently have a use case for `getComponents`? If not, perhaps we should remove it.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponent.java
aslo: Perhaps not, but they do make unit testing far easier. Do you think they should be removed?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponent.java
aslo: I prefer adding `isInstalled()` to making State an enum. I think it's best to keep the deserialization as flexible as possible.

aslo:
@briandealwis Regarding caching the result of a single `gcloud info` call, I don't think that would be a good idea, if for no other reason than the fact that we would need to worry about cache invalidation.

briandealwis:
Re: usecases for `getComponents()`: we have nothing so far, but I can imagine a simple pref page that lists the current components that allows the user to add/update components.

Re: caching: it could be a short-term caching of the info. Calling out to gcloud has some overhead (almost 600ms on my machine for a simple `gcloud info`).  That information is unlikely to change for a couple of minutes.

elharo:
On Fri, Oct 28, 2016 at 4:29 PM, Brian de Alwis <notifications@github.com>
wrote:

> Re: usecases for getComponents(): we have nothing so far, but I can
> imagine a simple pref page that lists the current components that allows
> the user to add/update components.
>

Can't we use the components list to check the version of the app engine
java component? That would be useful.


-- 
Elliotte Rusty Harold
elharo@macfaq.com

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponent.java
briandealwis: The object seems immutable, and the setters don't do anything useful. But I'm not fussed.  Does Github have a /shrug ? :-)

briandealwis:
`gcloud info —format=json` lists the installed components which is sufficient to determine if `app-engine-java` is installed.  That's our only concrete usecase currently.  Are the components updated in lockstep with `gcloud`?  Or can they be released independently?

elharo:
On Fri, Oct 28, 2016 at 4:41 PM, Brian de Alwis <notifications@github.com>
wrote:

> gcloud info —format=json lists the installed components which is
> sufficient to determine if app-engine-java is installed. That's our only
> concrete usecase currently. Are the components updated in lockstep with
> gcloud? Or can they be released independently?


They can be released, updated, and installed independently, I think. E.g.
we need 1.9.44 of the java component, nothing earlier.


-- 
Elliotte Rusty Harold
elharo@macfaq.com

joaoandremartins:
The components are updated together with gcloud. They can't be updated separately, found that out this week. That might mean we don't actually need the app-engine-java version.

briandealwis:
`gcloud components list —show-versions` provides a meaningful name for each component, useful for UIs, whereas `gcloud info` just lists the component identifier.

aslo:
@briandealwis Regarding caching, I think that's outside the scope of this PR, but I created an enhancement issue for it: https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/243

aslo:
FYI, since the first round of reviews, I've implemented [process exit code checking](https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/242/files#diff-562793416b98626facf7820af2a81c16R185), which was my final remaining TODO.  

If anyone wants to take a look and give some feedback on that, I'd appreciate it. Other than that, I believe I've addressed all other feedback.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponent.java
aslo: ¯\\_(ツ)_/¯

chanseokoh:
> The components are updated together with gcloud. They can't be updated separately, found that out this week. That might mean we don't actually need the app-engine-java version.

I'm not certain if this is 100% true. I'm having gcloud 131, but I have latest versions for other components including `app-engine-java`.

```
Your current Cloud SDK version is: 131.0.0
The latest available version is: 132.0.0

┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                                  Components                                                                 │
├──────────────────┬──────────────────────────────────────────────────────┬────────────┬───────────────┬──────────────────────────┬───────────┤
│      Status      │                         Name                         │ Installed  │     Latest    │            ID            │    Size   │
├──────────────────┼──────────────────────────────────────────────────────┼────────────┼───────────────┼──────────────────────────┼───────────┤
│ Update Available │ Cloud SDK Core Libraries                             │ 2016.10.17 │    2016.10.24 │ core                     │   5.0 MiB │
│ Update Available │ Cloud Storage Command Line Tool                      │       4.21 │          4.22 │ gsutil                   │   2.8 MiB │
│ Not Installed    │ App Engine Go Extensions                             │            │               │ app-engine-go            │  47.3 MiB │
│ Not Installed    │ Cloud Datastore Emulator                             │            │         1.2.1 │ cloud-datastore-emulator │  15.4 MiB │
│ Not Installed    │ Cloud Datastore Emulator (Legacy)                    │            │ v1beta3-1.0.0 │ gcd-emulator             │  38.1 MiB │
│ Not Installed    │ Cloud Pub/Sub Emulator                               │            │    2016.08.19 │ pubsub-emulator          │  16.3 MiB │
│ Not Installed    │ Google Container Registry's Docker credential helper │            │               │ docker-credential-gcr    │   2.2 MiB │
│ Not Installed    │ kubectl                                              │            │               │ kubectl                  │  15.9 MiB │
│ Installed        │ BigQuery Command Line Tool                           │     2.0.24 │        2.0.24 │ bq                       │   < 1 MiB │
│ Installed        │ Default set of gcloud commands                       │            │               │ gcloud                   │           │
│ Installed        │ gcloud Alpha Commands                                │ 2016.01.12 │    2016.01.12 │ alpha                    │   < 1 MiB │
│ Installed        │ gcloud Beta Commands                                 │ 2016.01.12 │    2016.01.12 │ beta                     │   < 1 MiB │
│ Installed        │ gcloud app Java Extensions                           │     1.9.44 │        1.9.44 │ app-engine-java          │ 132.0 MiB │
│ Installed        │ gcloud app Python Extensions                         │     1.9.40 │        1.9.40 │ app-engine-python        │   7.2 MiB │
└──────────────────┴──────────────────────────────────────────────────────┴────────────┴───────────────┴──────────────────────────┴───────────┘
To install or remove components at your current SDK version [131.0.0], run:
  $ gcloud components install COMPONENT_ID
  $ gcloud components remove COMPONENT_ID

To update your SDK installation to the latest version [132.0.0], run:
  $ gcloud components update
```

aslo:
@chanseokoh I think that's just because a new version of app-engine-java was not released with gcloud 132.0.0. My understanding is that each gcloud release points to a specific version of each component. If you downgrade versions to 131.0.0, for example, you'll see that a number of other components will be automatically downgraded to older versions.

chanseokoh:
@aslo Ah, I see. So it's just that `app-engine-java` 1.9.44 is bound to 131.0.0. Makes sense.

aslo:
I'd like to go ahead and merge this, but wanted to make one final request for comments. 

There was one somewhat significant addition since the last round of comments (see commit  https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/242/commits/bc99df4624be7272a579709a1908f9c753790906):  `getVersion` now returns a `CloudSdkVersion`. It's a `Comparable` that wraps the version and takes the burden off of callers to compare two versions.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  @Override
+  public String toString() {
+    return version;
+  }
+
+  @Override
+  public int compareTo(CloudSdkVersion otherVersion) {
+    List<Integer> mine = new LinkedList<>(this.versionComponents);
```
elharo: ArrayLIst is almost always preferable

src/test/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponentTest.java
```diff
@@ -0,0 +1,154 @@
+/*
+ * Copyright 2016 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the \"License");
```
elharo: no backslash before "License"

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -159,6 +170,26 @@ private void runGcloudCommand(List<String> args, String... topLevelCommand)
     processRunner.run(command.toArray(new String[command.size()]));
   }
 
+  private String runSynchronousGcloudCommand(List<String> args)
```
joaoandremartins: This method could use some Javadoc explaining why it follows a separate convention to the other methods here.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -235,6 +266,45 @@ public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
...
+   * gcloud, this method always uses a synchronous ProcessRunner and will block until the gcloud
+   * process returns.
+   *
+   * @throws ProcessRunnerException when process runner encounters an error
+   */
+  public CloudSdkVersion getVersion() throws ProcessRunnerException {
+    validateCloudSdk();
+
+    // gcloud info --format="value(basic.version)"
+    List<String> command = Lists.newArrayList("info");
```
joaoandremartins: Sounds like Lists.newArrayList() could be replaced by ImmutableList.of().

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -235,6 +266,45 @@ public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
...
+   *
+   * @throws ProcessRunnerException when process runner encounters an error
+   * @throws JsonSyntaxException when the cloud SDK output cannot be parsed
+   */
+  public List<CloudSdkComponent> getComponents()
+      throws ProcessRunnerException, JsonSyntaxException {
+    validateCloudSdk();
+
+    // gcloud components list --show-versions --format=json
+    List<String> command = Lists.newArrayList("components", "list");
```
joaoandremartins: Same as above about ImmutableList.of().

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -235,6 +266,45 @@ public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
...
+    command.addAll(GcloudArgs.get("show-versions", true));
+    command.addAll(GcloudArgs.get("format", "json"));
+
+    String componentsJson = runSynchronousGcloudCommand(command);
+    Type type = new TypeToken<List<CloudSdkComponent>>(){}.getType();
```
joaoandremartins: Is there a simpler way to do this?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/SynchronousOutputProcessRunner.java
```diff
@@ -0,0 +1,67 @@
...
+import com.google.cloud.tools.appengine.cloudsdk.process.ProcessExitListener;
+import com.google.cloud.tools.appengine.cloudsdk.process.ProcessOutputLineListener;
+import com.google.cloud.tools.appengine.cloudsdk.process.ProcessStartListener;
+import com.google.common.collect.Lists;
+
+/**
+ * A ProcessRunner that runs synchronous processes and makes their stderr and stdout streams
+ * available as Strings.
+ */
+public class SynchronousOutputProcessRunner extends DefaultProcessRunner implements ProcessRunner {
```
joaoandremartins: I'm still convinced this class is redundant. It also manages to completely ignore the listeners set by the user when sdk is constructed.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
```diff
@@ -0,0 +1,118 @@
...
+  public int hashCode() {
+    return Objects.hash(versionComponents);
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (obj == null) {
+      return false;
+    }
+    if (obj == this) {
```
joaoandremartins: Is this ever going to be true?

src/test/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponentTest.java
```diff
@@ -0,0 +1,154 @@
...
+  private Gson gson;
+
+  @Before
+  public void setup() {
+    this.gson = new Gson();
+  }
+
+  @Test
+  public void testGsonDeserializeJson() {
+    CloudSdkComponent result = gson.fromJson(getJson(), CloudSdkComponent.class);
```
joaoandremartins: Aren't result and expected swapped here?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
aslo: I omitted the javadoc since this method is private, but I'll add some regular comments explaining.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
```diff
@@ -0,0 +1,118 @@
...
+    while (first.size() > second.size()) {
+      second.add(0);
+    }
+  }
+
+  private List<Integer> buildVersionComponents(String version) throws NumberFormatException {
+    String[] components = version.split("\\.");
+    ImmutableList.Builder builder = ImmutableList.builder();
+    for (String num : components) {
+      builder.add(Integer.parseInt(num));
```
patflynn: do you have a commitment from the cloud SDK team that their component versions will always be \d+(\.d+)*?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
patflynn: @joaoandremartins this is standard equals boilerplate. it can also be true simply by passing a reference to the same object as an argument to equals.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+    if (this.getClass() != obj.getClass()) {
+      return false;
+    }
+    CloudSdkVersion otherVersion = (CloudSdkVersion) obj;
+
+    // if both versions don't have the same number of components, pad the smaller one with zeros
+    List<Integer> mine = new ArrayList<>(this.versionComponents);
+    List<Integer> other = new ArrayList<>(otherVersion.getVersionComponents());
+    rightPadZerosUntilSameLength(mine, other);
```
aslo: Not necessarily; the way I implemented this, "131" is equal to "131.0.0".

src/test/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponentTest.java
```diff
@@ -0,0 +1,154 @@
...
+    String regex = String.format(".*%s\":\\s*%s.*", expectedKey, expectedValue);
+    assertTrue(result.matches(regex));
+  }
+
+  private void assertJsonKeyValueExists(String expectedKey, boolean expectedValue, String result) {
+    String regex = String.format(".*%s\":\\s*%s.*", expectedKey, expectedValue);
+    assertTrue(result.matches(regex));
+  }
+
+  private String getJson() {
```
patflynn: what type of information is this json representing?

src/test/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponentTest.java
```diff
@@ -0,0 +1,154 @@
...
+
+  private Gson gson;
+
+  @Before
+  public void setup() {
+    this.gson = new Gson();
+  }
+
+  @Test
+  public void testGsonDeserializeJson() {
```
patflynn: this test name sounds like it's testing gson deserialization.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
```diff
@@ -0,0 +1,118 @@
...
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Represents the version of the Cloud SDK.
```
briandealwis: Could you please describe the format and how the version components are compared.  That's to ensure that this class is used to represent versions of Cloud SDK components.  For example, _Cloud Datastore Emulator (Legacy)_ (`gcd-emulator`) has version `v1beta3-1.0.0`.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
aslo: Yep. See also https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/242/commits/617892efc00ddfffdb1b37d29b94bea74b573a3b which demonstrates.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
aslo: I will add a comment to clarify, but FYI this class is intended only to represent the version of `gcloud`  (i.e. 131.0.0). Since it implements `Comparable`, I'm not even going to attempt to use this for the component versions, to avoid comparing versions with human-readable strings like `v1beta3-1.0.0`.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
aslo: FYI this class is intended only to wrap the gcloud version, i.e. `131.0.0`, not the component versions, which have been shown not to follow this pattern (see the Cloud Datastore Emulator (Legacy) component, whose current version is ` v1beta3-1.0.0 `).

I will seek confirmation from them that this pattern can be counted on for the output of `gcloud info --format="value(basic.version)"`.

src/test/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponentTest.java
aslo: It's testing that the class can be deserialized from json and that all of its expected fields are present, etc. The `CloudSdkComponent` class itself contains some Gson-specific annotations that make it deserialize properly when using the gson library, which is why I thought to include "Gson" in the test names. 

I can just rename to `testDeserialization`, etc if you prefer.

src/test/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponentTest.java
aslo: This is the expected json-ified version of the CloudSdkComponent test fixture in this class. To clarify, I'm updating these two method names to `getCloudSdkComponentTestFixture` and `getCloudSdkComponentTestFixtureAsJson`.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
aslo: :+1:

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
joaoandremartins: I noticed you have Javadoc on the public methods that invoke this one, so I'm not too fussed about Javadoc here.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
joaoandremartins: I thought that this would actually always be false because the passed object is the same, but the reference is different, since arguments are passed as copies of the reference in Java.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/SynchronousOutputProcessRunner.java
aslo: I don't think it's redundant. It's a specific configuration of a `ProcessRunner`, that encapsulates a nontrivial configuration (see the Builder), and it adds behaviors that pertain to the synchronous execution of processes. If this class did not exist, all of this plumbing would have to take place in the `CloudSdk` class, which IMO should not need to take on the responsibility of knowing how to configure the ProcessRunner correctly, how to determine if the process exited successfully, etc. I feel pretty strongly that this class has value.

Ignoring the listeners set by the user is a different story. I agree that it's feels a bit weird, but I still think it's better than the alternative. A user of the CloudSdk would expect that the listeners that they configure are for piping the output from a long-running process somewhere. Think about this usage pattern:
```
ProcessOutputLineListener someListener = // define a listener that pipes output directly to the user
CloudSdk sdk = new CloudSdk.Builder().addStdOutLineListener(someListener).build();

if (!sdk.getComponents().contains(someRequiredComponent)) {
  // assume this code path is not followed
} else {
  sdk.runDevAppServerCommand("some command");
}
```
Assuming that the SynchronousOutputProcessRunner were to use the `someListener` that the CloudSdk was configured with, the user would see the output of the `gcloud components list` command that is called internally by `getComponents()`. I think that's bad.

By ignoring the user's configured output listeners, we can fully encapsulate the plumbing involved in finding the gcloud version. I think that's a very good thing, as it vastly simplifies the use of this method.

src/test/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkComponentTest.java
patflynn: I feel strongly that the serialization test needs to include the driving code.  Your other option is to move the driving code into this class.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/SynchronousOutputProcessRunner.java
```diff
@@ -0,0 +1,80 @@
...
+  private StringBuilderProcessOutputLineListener stdErrListener;
+  private ExitCodeRecorderProcessExitListener exitListener;
+
+  @VisibleForTesting
+  SynchronousOutputProcessRunner(StringBuilderProcessOutputLineListener stdOutListener,
+                                         StringBuilderProcessOutputLineListener stdErrListener,
+                                         ExitCodeRecorderProcessExitListener exitListener) {
+    super(
+        false                                                         /* async */,
+        Lists.<ProcessExitListener>newArrayList(exitListener)         /* exitListeners */,
```
joaoandremartins: These should be ImmutableLists.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/SynchronousOutputProcessRunner.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+public class SynchronousOutputProcessRunner extends DefaultProcessRunner implements ProcessRunner {
+
+  private StringBuilderProcessOutputLineListener stdOutListener;
+  private StringBuilderProcessOutputLineListener stdErrListener;
+  private ExitCodeRecorderProcessExitListener exitListener;
+
+  @VisibleForTesting
+  SynchronousOutputProcessRunner(StringBuilderProcessOutputLineListener stdOutListener,
+                                         StringBuilderProcessOutputLineListener stdErrListener,
```
aslo: :+1:

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/SynchronousOutputProcessRunner.java
aslo: :+1:

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/SynchronousOutputProcessRunner.java
aslo: Of the 3 suggestions, I actually like the second one the best. I see your point, it does save a lot of LOC. I'll update.

PR #245, opened 2016-11-02
Ignore suffixes

aslo:
quick followup to https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/242

According to the Cloud SDK team, the cloud sdk version follows semantic versioning (see semver.org). This change is necessary so that parsing doesn't break if a pre-release version string is used, e.g. `0.1.0-rc1`.  

Note that with this change, any version suffix is simply ignored for comparison purposes. Full semver parsing and comparison will be implemented later.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
@@ -115,12 +115,14 @@ public class CloudSdkVersion implements Comparable<CloudSdkVersion> {
     }
   }
 
+  // TODO(alexsloan): implement and assert true semver comparisons, such that prerelease suffixes
+  // are compared according to the semver spec (semver.org)
   private List<Integer> parseVersionComponents(String version) throws NumberFormatException {
     // just strip out any suffixes
     version = ignoreBuildOrPrereleaseSuffix(version);
 
     String[] components = version.split("\\.");
-    ImmutableList.Builder builder = ImmutableList.builder();
+    ImmutableList.Builder<Integer> builder = ImmutableList.builder();
     for (String num : components) {
       builder.add(Integer.parseInt(num));
     }
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
@@ -45,9 +45,9 @@ public class CloudSdkVersion implements Comparable<CloudSdkVersion> {
   /**
    * Constructs a CloudSdkVersion from a version string.
    * @param version a non-null, nonempty string of the form "\d+(\.\d+)*[+-].*".
-   * @throws NumberFormatException if the string cannot be parsed
+   * @throws IllegalArgumentException if the string cannot be parsed
    */
-  public CloudSdkVersion(String version) throws NumberFormatException {
+  public CloudSdkVersion(String version) throws IllegalArgumentException {
     Preconditions.checkArgument(!Strings.isNullOrEmpty(version));
 
     this.version = version;
@@ -117,7 +117,7 @@ public class CloudSdkVersion implements Comparable<CloudSdkVersion> {
 
   private List<Integer> parseVersionComponents(String version) throws NumberFormatException {
     // just strip out any suffixes
-    version = removeSuffix(version);
+    version = ignoreBuildOrPrereleaseSuffix(version);
 
     String[] components = version.split("\\.");
     ImmutableList.Builder builder = ImmutableList.builder();
@@ -127,7 +127,10 @@ public class CloudSdkVersion implements Comparable<CloudSdkVersion> {
     return builder.build();
   }
 
-  private String removeSuffix(String version) {
+  // Returns the version string without its prerelease and/or build suffix. Any characters following
+  // (and including) the first occurrence of either the BUILD_SEPARATOR or the PRERELEASE_SEPARATOR
+  // will be ignored
+  private String ignoreBuildOrPrereleaseSuffix(String version) {
     List<Character> separators = ImmutableList.of(BUILD_SEPARATOR, PRERELEASE_SEPARATOR);
     for (int i = 0; i < version.length(); i++) {
       if (separators.contains(version.charAt(i))) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
```diff
@@ -24,27 +24,36 @@
...
   private final List<Integer> versionComponents;
   private final String version;
 
   /**
    * Constructs a CloudSdkVersion from a version string.
-   * @param version a non-null, nonempty string of the form 130.0.0
+   * @param version a non-null, nonempty string of the form "\d+(\.\d+)*[+-].*".
    * @throws NumberFormatException if the string cannot be parsed
    */
   public CloudSdkVersion(String version) throws NumberFormatException {
```
patflynn: should this continue to throw NumberFormatException?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
```diff
@@ -108,12 +117,26 @@ private void rightPadZerosUntilSameLength(List<Integer> first, List<Integer> sec
...
+
     String[] components = version.split("\\.");
     ImmutableList.Builder builder = ImmutableList.builder();
     for (String num : components) {
       builder.add(Integer.parseInt(num));
     }
     return builder.build();
   }
+
+  private String removeSuffix(String version) {
```
patflynn: It would be useful to describe what a suffix is for this method.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
```diff
@@ -24,27 +24,36 @@
...
+  private static final char BUILD_SEPARATOR = '+';
+  private static final char PRERELEASE_SEPARATOR = '-';
+
   private final List<Integer> versionComponents;
   private final String version;
 
   /**
    * Constructs a CloudSdkVersion from a version string.
-   * @param version a non-null, nonempty string of the form 130.0.0
+   * @param version a non-null, nonempty string of the form "\d+(\.\d+)*[+-].*".
```
patflynn: So we will still blow up on something like 132.alpha-1.0 and the like?

src/test/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersionTest.java
```diff
@@ -56,6 +57,18 @@ public void testToString() {
   public void testEquals_differentSizes() {
     assertTrue(new CloudSdkVersion("0.1.0").equals(new CloudSdkVersion("0.1")));
     assertTrue(new CloudSdkVersion("0").equals(new CloudSdkVersion("0.0.0.0")));
+    assertFalse(new CloudSdkVersion("1.1").equals(new CloudSdkVersion("1.10")));
+  }
+
+  @Test
+  public void testEquals_preRelease() {
```
patflynn: If we do want to blow up on some kinds of input we should have tests to verify that.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
aslo: We will, as that is not a valid semver. Per your comment below, I'll add tests to verify that we do blow up.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
aslo: :+1:

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
aslo: Good catch. I think an IllegalArgumentException would be more appropriate. I'll add tests to verify that behavior.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
patflynn: I'm just taking a look at the semver spec, and get what's going on here.  

Might this code be a little simpler if we just matched the input string with the digit part of your regex: e.g. \d+(\.\d+)*, throw an IllegalArgumentException if there's no match, and then parse that knowing it can be safely broken into a list of numbers?

aslo:
@patflynn Any other comments? If not, I'd love an LGTM so I can merge.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
```diff
@@ -108,12 +115,29 @@ private void rightPadZerosUntilSameLength(List<Integer> first, List<Integer> sec
     }
   }
 
-  private List<Integer> buildVersionComponents(String version) throws NumberFormatException {
+  private List<Integer> parseVersionComponents(String version) throws NumberFormatException {
+    // just strip out any suffixes
+    version = ignoreBuildOrPrereleaseSuffix(version);
+
     String[] components = version.split("\\.");
     ImmutableList.Builder builder = ImmutableList.builder();
```
patflynn: I must be getting old.  This doesn't need a LHS type argument?

src/test/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersionTest.java
```diff
@@ -56,6 +68,18 @@ public void testToString() {
...
+    // TODO(alexsloan): implement and assert true semver comparisons, such that prerelease suffixes
```
patflynn: you probably don't want this todo in the test case. It should be in the non-test code.

src/test/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersionTest.java
```diff
@@ -41,9 +44,18 @@ public void testConstructor_emptyString() {
     new CloudSdkVersion("");
   }
 
-  @Test(expected = NumberFormatException.class)
-  public void testConstructor_nonNumeric() {
-    new CloudSdkVersion("v1beta3-1.0.0");
+  @Test
+  public void testConstructor_invalid() {
```
patflynn: This is going to feel academic, but generally with unit tests, each test case should be as specific as possible. Is there something qualitatively different about the inputs provided?  Are they different edges?  If so, put them in different tests and label them clearly.

aslo: :+1: I'll tighten up these tests

PR #249, opened 2016-11-13
File Utilities that don't depend on eclipse

elharo:
@akerekes @briandealwis @meltsufin as requested, I've pulled these out into an independent, tested utility.

After this bakes a bit more, I want to push this all the way up to Guava, but that's not going to happen on our time frame.

### DIFF
--- a/src/main/java/com/google/cloud/tools/io/FilePermissions.java
+++ b/src/main/java/com/google/cloud/tools/io/FilePermissions.java
@@ -21,12 +21,14 @@ import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 
+import com.google.common.annotations.Beta;
+
 /**
  * Non-IDE specific utility methods for working with file systems.
  * We hope to migrate these into Guava. Clients should beware that these methods are
  * not yet stable and may change without notice. 
  */
-// todo Beta annotation
+@Beta
 public class FilePermissions {
 
   /**
--- a/src/main/java/com/google/cloud/tools/io/FilePermissions.java
+++ b/src/main/java/com/google/cloud/tools/io/FilePermissions.java
@@ -16,12 +16,11 @@
 
 package com.google.cloud.tools.io;
 
+import com.google.common.annotations.Beta;
 import java.io.IOException;
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-
-import com.google.common.annotations.Beta;
 
 /**
  * Non-IDE specific utility methods for working with file systems.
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/cloud/tools/io/FilePermissions.java
```diff
@@ -0,0 +1,66 @@
...
+        throw new FileAlreadyExistsException(segment + " is a file");
+      }
+    }
+    
+    // Can't create a directory if the bottom most currently existing directory in
+    // the path is not writable. 
+    for (Path segment = path; segment != null; segment = segment.getParent()) {
+      if (Files.isDirectory(segment)) {
+        if (!Files.isWritable(segment)) {
+          throw new IOException(path + " is not writable");
```
briandealwis: This should be `segment`, not `path`?

src/main/java/com/google/cloud/tools/io/FilePermissions.java
```diff
@@ -0,0 +1,66 @@
...
+   * 
+   * @param path tentative location for directory
+   * @throws IOException if a directory cannot be created at the specified path
+   */
+  public static void verifyDirectoryCreatable(Path path) throws IOException {
+    
+    // Can't create a directory if a non-directory file already exists with that name
+    // somewhere in the path. 
+    for (Path segment = path; segment != null; segment = segment.getParent()) {
+      if (Files.isRegularFile(segment)) {
```
briandealwis: I don't think `isRegularFile()` handles situations where the segment is a special file (e.g., a device node like `/dev/sd0a`, a unix socket, or FIFO).

Could these be combined into something like:

```
for (Path segment = path; segment != null; segment = segment.getParent()) {
  if (Files.isDirectory(segment)) {
    if (!Files.isWritable(segment)) {
      throw new IOException(segment + " is not writable");
    }
    break;
  } else if(Files.exists(segment)) {
    throw new FileAlreadyExistsException(segment + " is not a directory");
  }
}
```

elharo: I'm not sure I follow. Can you give an example of a case where this method returns the wrong answer?

These two loops can't easily be combined, because the breaks don't line up.

elharo:
PTAL @briandealwis now flags /dev/null etc. as non-writable

src/test/java/com/google/cloud/tools/io/FilePermissionsTest.java
```diff
@@ -45,6 +45,16 @@ public void testSubDirectoryCanBeCreated() throws IOException {
   }
   
   @Test
+  public void testSubDirectoryCannotBeCreatedInDevNull()  {
+    try {
+      FilePermissions.verifyDirectoryCreatable(Paths.get("/dev/null/foo/bar"));
```
briandealwis: This path is Unix specific. It seems Windows 7 or 8 made `C:\` read-only, but the test will likely succeed if run from a different drive.  Probably just skip the test if on Windows.

elharo: done

elharo:
OK, pulled it into a single loop. Looking at it, I'm not sure this is actually easier to read.

briandealwis:
It matches how [you described it](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/888#issuecomment-257258899):
> a full solution that walks up the path until it finds a directory it either can or cannot write to.

PR #252, opened 2016-11-16
Implement CloudSdkVersion to the semver spec

aslo:
This PR is a followup to https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/242, in which a `CloudSdkVersion` class was implemented in order to make it easy to compare versions of the Cloud SDK. 

For completeness, this PR modifies the CloudSDK class so that it now implements the remainder of the [Semantic Version spec](http://semver.org/). The main addition here is that now the pre-release portions of semantic version strings can be compared. (i.e. `"0.1.0-alpha" < "0.1.0-beta" < "0.1.0"`).


pom.xml
```diff
@@ -7,7 +7,7 @@
 
   <groupId>com.google.cloud.tools</groupId>
   <artifactId>appengine-plugins-core</artifactId>
-  <version>0.2.1-SNAPSHOT</version>
+  <version>0.2.2-SNAPSHOT</version>
```
patflynn: was there a release where they forgot to update this?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
```diff
@@ -20,71 +20,119 @@
...
+  private static final Pattern SEMVER_PATTERN = Pattern.compile(getSemVerRegex());
 
-  private final List<Integer> versionComponents;
   private final String version;
 
+  private final int majorVersion;
+  private final int minorVerion;
+  private final int patchVersion;
+  private final CloudSdkVersionPreRelease preRelease;
+  private final String build;
```
patflynn: maybe just a comment to indicate that the 'build' variable is a component of the version string.  I was kind of thrown until I read code further down.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
```diff
@@ -20,71 +20,119 @@
...
+
+    preRelease = matcher.group("prerelease") != null
+        ? new CloudSdkVersionPreRelease(matcher.group("prerelease")) : null;
+    build = matcher.group("build");
+
     this.version = version;
-    this.versionComponents = parseVersionComponents(version);
+  }
+
+  private static String getSemVerRegex() {
```
patflynn: this is kinda hardcore.  Did we use anything as a basis for it?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
```diff
@@ -20,71 +20,119 @@
...
   public int hashCode() {
-    return Objects.hash(versionComponents);
+    return Objects.hash(majorVersion, minorVerion, patchVersion, preRelease);
   }
 
   /**
-   * Compares objects for equality. Two CloudSdkVersions are considered equal if, from left to
-   * right, their version integers are equal.
+   * Compares this to another CloudSdkVersion for equality. Note that the build identifier field is
+   * excluded for comparison.
```
patflynn: but compareto doesn't take Object does it?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
```diff
@@ -102,44 +150,24 @@ public boolean equals(Object obj) {
...
-  // will be ignored
-  private String ignoreBuildOrPrereleaseSuffix(String version) {
-    List<Character> separators = ImmutableList.of(BUILD_SEPARATOR, PRERELEASE_SEPARATOR);
-    for (int i = 0; i < version.length(); i++) {
-      if (separators.contains(version.charAt(i))) {
-        return version.substring(0, i);
-      }
-    }
-    return version;
+  public String getBuild() {
```
patflynn: I think javadoc for what 'build' represents in semver would be helpful here.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersionPreRelease.java
```diff
@@ -0,0 +1,181 @@
...
+
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Represents the pre-release field in a Cloud SDK Version. The pre-release field is comprised of
```
patflynn: could you provide examples?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersionPreRelease.java
```diff
@@ -0,0 +1,181 @@
...
+ * one or more dot-separated segments, which are can be treated as either numeric-only, or
+ * alphanumeric, depending on their contents.
+ */
+class CloudSdkVersionPreRelease implements Comparable<CloudSdkVersionPreRelease> {
+
+  private List<PreReleaseSegment> segments;
+  private final String preRelease;
+
+  /**
+   * Constructs a new CloudSdkVersionPreRelease from a string representation.
```
patflynn: an example here of what a normal full semver string would be and which part is the pre-release could be helpful.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
patflynn: Also if compare ignores the build string than is equals equivalent to compareTo == 0?

pom.xml
elharo: We're currently at 0.2.0. We usually only change this as part of a release.

aslo: Ah whoops, that's right. I'll revert this.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
aslo: Yep, this is similar to the regex that the Cloud SDK uses for version comparison.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
aslo: > but compareto doesn't take Object does it?

Nope, compareTo takes a CloudSdkVersion. I'm not sure what you're getting at here?


> Also if compare ignores the build string than is equals equivalent to compareTo == 0?

Yep, both equals and compareTo ignore the build string, so equals is equivalent to compareTo==0.  (The semver spec dictates that the build id must be ignored for comparison, so as a result, I chose to also ignore it in equals, in order to respect the convention that  `(x.compareTo(y)==0) == (x.equals(y))`

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
aslo: :+1:

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersionPreRelease.java
aslo: :+1: I'll add this to the class-level javadoc.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersionPreRelease.java
aslo: added an example to class-level javadoc.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
patflynn: The confusion was mine wrgt compareTo() signature. I didn't notice github had folded everything before the return statement.

WRGT equals vs compareto,  I don't think that it's a requirement of the compareTo contract that compareTo == 0 always implies the two objects are equal.   It's odd to have two objects be equals with different toString values and/or responses to getters.

patflynn:
To me the issue is that this class behaves worse than the example in the
doc you link. Two different CloudSdkVersions with different build variants
but the same comparable fields would be treated as the same in both set/map
types that respected the general set/map contract, as well as the Sorted
map/set types which violate it.

Perhaps if we really want to ignore the variant we should just throw it
away at object initialization.

patflynn:
in a way the BigDecimal example they provide as a good case for violating
consistency between equals/compareTo is very similar to our case.

aslo:
Good point re: BigDecimal. I'll update the equals implementation to factor in the build identifier.

aslo:
@patflynn Mind giving a final review?

PR #255, opened 2016-11-19
check for minimum required Cloud SDK version

elharo:
@briandealwis @aslo Anything less than 131 causes issues.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -387,9 +387,10 @@ public class CloudSdk {
 
   private void validateCloudSdkVersion() {
     try {
-      if (getVersion().getMajorVersion() < MINIMUM_VERSION) {
-        throw new CloudSdkNotFoundException(
-            "Cloud SDK version " + getVersion() + " is too old. Please update.");        
+      CloudSdkVersion version = getVersion();
+      if (version.getMajorVersion() < MINIMUM_VERSION) {
+        throw new CloudSdkOutOfDateException(
+            "Cloud SDK version " + version + " is too old. Please update.");        
       }
     } catch (ProcessRunnerException ex) {
       throw new CloudSdkNotFoundException(ex);              
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -258,7 +258,7 @@ public class CloudSdk {
    * Executes an App Engine SDK CLI command.
    *
    * @throws AppEngineJavaComponentsNotInstalledException when the App Engine Java components are
-   *                                                      not installed in the Cloud SDK
+   *     not installed in the Cloud SDK
    */
   public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
     validateAppEngineJavaComponents();
@@ -380,7 +380,7 @@ public class CloudSdk {
    *
    * @throws CloudSdkNotFoundException when an up-to-date Cloud SDK is not installed where expected
    */
-  public void validateCloudSdk() {
+  public void validateCloudSdk() throws CloudSdkNotFoundException {
     validateCloudSdkLocation();
     validateCloudSdkVersion();
   }
@@ -389,8 +389,8 @@ public class CloudSdk {
     try {
       CloudSdkVersion version = getVersion();
       if (version.getMajorVersion() < MINIMUM_VERSION) {
-        throw new CloudSdkOutOfDateException(
-            "Cloud SDK version " + version + " is too old. Please update.");        
+        throw new CloudSdkOutOfDateException("Cloud SDK version " + version
+            + " is too old. Please update to at least " + MINIMUM_VERSION);
       }
     } catch (ProcessRunnerException ex) {
       throw new CloudSdkNotFoundException(ex);              
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -659,6 +659,7 @@ public class CloudSdk {
             ServiceLoader.load(CloudSdkResolver.class, getClass().getClassLoader());
         resolvers = Lists.newArrayList(services);
         // Explicitly add the PATH-based resolver
+        System.err.println("adding PathResolver...");
         resolvers.add(new PathResolver());
       }
       Collections.sort(resolvers, new ResolverComparator());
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkOutOfDateException.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.appengine.cloudsdk;
+
+/**
+ * The Cloud SDK that was found is too old (generally before 131.0).
+ */
+public class CloudSdkOutOfDateException extends CloudSdkNotFoundException {
```
patflynn: I think we're fudging things a bit to say that an incompatible version is 'not found'.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -377,11 +376,28 @@ public Path getJarPath(String jarName) {
...
+  public void validateCloudSdk() {
+    validateCloudSdkLocation();
+    validateCloudSdkVersion();
+  }
+
+  private void validateCloudSdkVersion() {
+    try {
+      CloudSdkVersion version = getVersion();
+      if (version.getMajorVersion() < MINIMUM_VERSION) {
+        throw new CloudSdkOutOfDateException(
```
patflynn: Should we indicate the required minimum version here?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -377,11 +376,28 @@ public Path getJarPath(String jarName) {
...
   /**
-   * Checks whether the Cloud SDK Path with is valid.
+   * Checks whether the Cloud SDK path and version are valid.
    *
-   * @throws CloudSdkNotFoundException when the Cloud SDK is not installed where expected
+   * @throws CloudSdkNotFoundException when an up-to-date Cloud SDK is not installed where expected
    */
-  public void validateCloudSdk() throws CloudSdkNotFoundException {
+  public void validateCloudSdk() {
+    validateCloudSdkLocation();
```
patflynn: Why remove the throws clause?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -172,8 +173,8 @@ private void runGcloudCommand(List<String> args, String... topLevelCommand)
   // class's main configured ProcessRunner should be used.
   private String runSynchronousGcloudCommand(List<String> args)
       throws ProcessRunnerException {
-    validateCloudSdk();
-
+    validateCloudSdkLocation();
```
patflynn: I'm not a big fan of the way we have the hidden runtime validation exceptions that can blow up callers of this. Not really this PRs problem though.

aslo: FYI I believe @joaoandremartins is working on some changes to validation in this library that uses an enum type for validation errors, building from [what we already do](https://github.com/GoogleCloudPlatform/google-cloud-intellij/blob/master/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/sdk/DefaultCloudSdkService.java#L104-L132) in the intellij plugin.  When that's done it would remove the need to use unchecked exceptions.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
briandealwis: We could but chances are the user will just update to the latest anyways.

patflynn:
Cloud SDK offers users the ability to pin to a version. There's a good
chance that companies do that.  We should make it clear what the minimum
version is that they need to pin to.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
elharo: We should probably not have used unchecked exceptions here in the first place, but c'est la code. I made the version exception a subclass of the not found exception so clients won't get immediately blown up.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
elharo: I'd have to dig it out, but my recollection is that convention for runtime exceptions is no throws clause but an @throws doc comment

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkOutOfDateException.java
elharo: I could make it a different exception, but that will require client updates.

patflynn: Since the CloudSdkNotFoundException is runtime, I'm not sure there's a problem.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
patflynn: I've seen using throws for a runtime exception be a kind of middle ground between the checked/unchecked exception crowds.  I don't think we need to remove it, and having it there makes it harder to miss and the IDE try/catch generation work better.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkOutOfDateException.java
elharo: Their compiles wouldn't break, but they might experience runtime exceptions they don't have a catch block for. Arguably, it would be better if their compiles did break in this case since devs would be notified they needed to update their code.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
elharo: How do IDEs treat it differently if it's in a throws clause?

patflynn: In IntelliJ at least, if I 'surround with try/catch' on a block, it'll automatically put the exceptions listed in the method declaration into the catch.  It does not do that for exceptions listed in @throws.

elharo: Interesting. Eclipse doesn't do that to my knowledge. Restored.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
elharo: added the minimum version to the message. can't hurt.

PR #256, opened 2016-11-21
Check Windows per-user install

elharo:
@briandealwis

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
@@ -59,6 +59,9 @@ public class PathResolver implements CloudSdkResolver {
     return searchPaths(possiblePaths);
   }
 
+  /** 
+   * The default location for a single-user install of Cloud SDK on Windows.
+   */
   private static String getLocalAppDataLocation() {
     String localAppData = System.getenv("LOCALAPPDATA");
     if (localAppData != null) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
```diff
@@ -56,7 +59,16 @@ public Path getCloudSdkPath() {
     return searchPaths(possiblePaths);
   }
 
-  private void getLocationsFromPath(List<String> possiblePaths) {
+  private static String getLocalAppDataLocation() {
```
briandealwis: I'd add a comment here that the single-user installation option (the default) places the sdk in the LocalAppData

src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
```diff
@@ -43,7 +45,8 @@ public Path getCloudSdkPath() {
 
     // search program files
     if (System.getProperty("os.name").contains("Windows")) {
-      possiblePaths.add(getProgramFilesLocation());
+        possiblePaths.add(getLocalAppDataLocation());
```
patflynn: Weren't we just using the 'where/which' commands for this?  Or is that some place else and this is a fallback?

elharo: is there such a command on Windows?

PR #258, opened 2016-11-26
Replaced 'swagger.yaml' with openapi.yaml. Refer to b/32670418.

jasminejaksic:


elharo:
What is the situation on the server? What does it do with a file named swagger.yaml? I'm fine with adding openapi.yaml to to the whitelist, but if clients are still uploading and the server is still processing swagger.yaml we should probably keep that too.

jasminejaksic:
The file name should not matter on the server side. 
All the external doc references will also be named (refer to CL/140226260).

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
```diff
@@ -39,7 +39,7 @@
 public class CloudSdkAppEngineFlexibleStaging implements AppEngineFlexibleStaging {
   protected static final Set<String> APP_ENGINE_CONFIG_FILES_WHITELIST = ImmutableSet.of("app.yaml",
       "cron.yaml", "queue.yaml", "dispatch.yaml", "index.yaml", "dos.yaml", "swagger.json",
-      "swagger.yaml");
+      "openapi.yaml");
```
elharo: why did you change swagger.yaml --> openapi.yaml and not change swagger.json --> openapi.json?

jasminejaksic: All the references will be changed sooner or later. For now I have edited the comments and replaced swagger.yaml.

loosebazooka: I'm not sure I understand the explanation. Preferably would like to see everything changed that needs to be changed.

loosebazooka:
I'm going to append to this pr to change the json file as well.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
elharo: This is not an example. This is production code. It needs to support the file names that actually appear in the wild.

patflynn:
presumably somewhere in gcloud there's a switch that tells it what to do
when it's deploying 'swagger.yam' or 'openapi.yaml'.  It seems like it
would have to be filename specific.

PR #259, opened 2016-11-27
It's the Cloud SDK that is not found, not the Cloud SDK path.

elharo:
@briandealwis also uses a more specific exception

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -641,7 +641,7 @@ public class CloudSdk {
               + ": exception thrown when searching for Google Cloud SDK", ex);
         }
       }
-      throw new CloudSdkNotFoundException("The Google Cloud SDK could not be found in the default"
+      throw new CloudSdkNotFoundException("The Google Cloud SDK could not be found in the customary"
           + " locations and no path was provided.");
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -641,8 +641,8 @@ private Path discoverSdkPath() {
               + ": exception thrown when searching for Google Cloud SDK", ex);
         }
       }
-      throw new AppEngineException("Google Cloud SDK path was not provided and could not be"
-          + " found in any known install locations.");
+      throw new CloudSdkNotFoundException("The Google Cloud SDK could not be found in the default"
+          + " locations and no path was provided.");
```
briandealwis: Not sure if _default_ is correct here as the different resolvers  may have searched in more than the default locations.

_The Google Cloud SDK could not be located_?

elharo: how about default --> customary?

briandealwis: "customary" works for me.  I was unsure about _no path was provided_ but that seems covered with `GOOGLE_CLOUD_SDK_HOME` and the preferences.

elharo:
customary it is

PR #260, opened 2016-11-27
validator for App Engine service (a.k.a. module) names

elharo:
@briandealwis We'll need this or something like it to add service names to our new application wizards

### DIFF
--- a/src/main/java/com/google/cloud/tools/project/ServiceNameValidator.java
+++ b/src/main/java/com/google/cloud/tools/project/ServiceNameValidator.java
@@ -16,12 +16,19 @@
 
 package com.google.cloud.tools.project;
 
+import com.google.common.annotations.Beta;
+
 /**
  * Each service and each version must have a name. A name can contain numbers, letters, and hyphens.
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/cloud/tools/project/ServiceNameValidator.java
```diff
@@ -0,0 +1,53 @@
...
+      for (char c : name.toCharArray()) {
+        if (Character.isAlphabetic(c)) {
+          continue;
+        } else if (Character.isDigit(c)) {
+          continue;
+        } else if (c == '-') {
+          continue;
+        } else {
+          return false;
+        }
```
elharo: What's the difference between a regular expression and line noise?
Line noise is more compressible.

briandealwis: I don't know what that means. This method as is took me more mental effort to read and understand than expected.  But I don't feel strongly about this.

elharo: Cryptographer joke, but the short answer is no, I don't find the regex easier to read.

PR #263, opened 2016-12-01
Places quiet flag in right place.

joaoandremartins:
Flag had no effect at the end of the gcloud command.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -132,12 +132,12 @@ public class CloudSdk {
 
     List<String> command = new ArrayList<>();
     command.add(getGCloudPath().toString());
+    command.add("--quiet");
+
     for (String commandToken : topLevelCommand) {
       command.add(commandToken);
     }
     command.addAll(args);
-
-    command.add("--quiet");
     command.addAll(GcloudArgs.get("format", appCommandOutputFormat));
 
     Map<String, String> environment = Maps.newHashMap();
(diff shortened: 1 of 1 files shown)


chanseokoh:
> Could we test this by running a command and verifying the output is quiet as expected?

Not sure if this is about making the output quiet. `--help` gives below:

```
     --quiet, -q
        Disable all interactive prompts when running gcloud commands. If input
        is required, defaults will be used, or an error will be raised.
```

Anyway, it's a bit surprising that the location of the flag matters. Maybe adding `args` is the culprit, e.g., the given `args` are sometimes not really complete in that they consume the next token that follows (`--quiet` in this case)?

aslo:
I'm not convinced that this is always correct. For example, `gcloud app deploy` needs its args in the following order. Note that the global flags including --quiet are required to be at the end.

```
    gcloud app deploy [DEPLOYABLES ...] [--bucket=BUCKET]
        [--image-url=IMAGE_URL] [--no-promote] [--no-stop-previous-version]
        [--version=VERSION, -v VERSION] [GLOBAL-FLAG ...]
```

joaoandremartins:
Hum, actually, I might have misread gcloud --help...

But I tried the gcloud beta debug source gen-repo-info-file command with the --quiet flag at the beginning and end, and it only worked with --quiet right after gcloud.

i.e.,

Works: gcloud --quiet beta debug source gen-repo-info-file --source-directory ~

Doesn't work: gcloud beta debug source gen-repo-info-file --source-directory ~ --quiet

I'd be inclined to agree with @chanseokoh, except on my previous example, there is no args after the quiet flag.

joaoandremartins:
For deploy, I just verified that the flag works in both places. I prefer to stick to what the docs say, and have it at the end. But sticking to this will delay https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1096...

What I suggest is to file an issue with gcloud so the --quiet flag works as intended for the gen-repo-info-file command and, temporarily, place it at the beginning of the command.

joaoandremartins:
Opened internal issue 33285287.

In the meantime, would be great to get your guys' ideas on temporarily relocating the flag to the beginning of the command (I verified it works for deploy, what other gcloud do we use that I should test?), since a fix will still require at least one Cloud SDK version.

joaoandremartins:
@chanseokoh @etanshaul @elharo @aslo  can anyone review? Ideally, this would be in the 0.2.2 release.

PR #268, opened 2016-12-02
Change swagger -> openapi

loosebazooka:
I thought I could build on top of #258 but it's in a user's branch. So this is meant to replace that.


elharo:
We still haven't established that we should delete the swagger file names.

loosebazooka:
add openapi, leave swagger

patflynn:
I think we still need to talk about this. I don't understand what the point
of having this file in the staging directory since gcloud app deploy
doesn't understand it.

PR #270, opened 2016-12-03
add test for getVersion

elharo:
@joaoandremartins @meltsufin @briandealwis @chanseokoh

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/DefaultProcessRunner.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/DefaultProcessRunner.java
@@ -106,9 +106,9 @@ public class DefaultProcessRunner implements ProcessRunner {
       }
 
       processBuilder.command(command);
-
+      
       Process process = processBuilder.start();
-
+      
       // Only handle stdout or stderr if there are listeners.
       if (!stdOutLineListeners.isEmpty()) {
         handleStdOut(process);
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/DefaultProcessRunner.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/DefaultProcessRunner.java
@@ -141,7 +141,7 @@ public class DefaultProcessRunner implements ProcessRunner {
     this.environment = environment;
   }
 
-  private void handleStdOut(final Process process) {
+  private void handleStdOut(Process process) {
     final Scanner stdOut = new Scanner(process.getInputStream(), Charsets.UTF_8.name());
     Thread stdOutThread = new Thread("standard-out") {
       @Override
@@ -159,7 +159,7 @@ public class DefaultProcessRunner implements ProcessRunner {
     stdOutThread.start();
   }
 
-  private void handleErrOut(final Process process) {
+  private void handleErrOut(Process process) {
     final Scanner stdErr = new Scanner(process.getErrorStream(), Charsets.UTF_8.name());
     Thread stdErrThread = new Thread("standard-err") {
       @Override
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/StringBuilderProcessOutputLineListener.java
```diff
@@ -19,23 +19,23 @@
...
+  private final StringBuffer buffer = new StringBuffer();
 
   public StringBuilderProcessOutputLineListener() {
-    stringBuilder = new StringBuilder();
   }
 
   @Override
-  public void onOutputLine(String line) {
-    stringBuilder.append(line);
+  public synchronized void onOutputLine(String line) {
```
chanseokoh: I'm not sure if `synchronized` is necessary. Particularly for `onOutputLine()`, only one thread calls this method if I'm not mistaken. For `toString()` below, `buffer` is thread-safe, so I think we are good.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/DefaultProcessRunner.java
```diff
@@ -106,9 +106,9 @@ public void run(String[] command) throws ProcessRunnerException {
       }
 
       processBuilder.command(command);
-
+      
       Process process = processBuilder.start();
-
+
```
briandealwis: Looks like whitespace introduced?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/StringBuilderProcessOutputLineListener.java
```diff
@@ -19,23 +19,23 @@
...
+    buffer.append(line);
   }
 
   @Override
-  public String toString() {
-    return stringBuilder.toString();
+  public synchronized String toString() {
```
briandealwis: [@chanseokoh's logic](https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/270/files#r90747480) seems to apply here.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/process/ProcessOutputLineListener.java
```diff
@@ -20,6 +20,7 @@
 
 /**
  * Line output listener for {@link DefaultProcessRunner}.
+ * Warning: all implementations must be thread safe.
```
briandealwis: An explanation here would be useful: it took me a minute or two to puzzle through.  _The process writes lines, and any consumers listen from the lines._

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/StringBuilderProcessOutputLineListener.java
```diff
@@ -19,23 +19,23 @@
...
 
   public StringBuilderProcessOutputLineListener() {
-    stringBuilder = new StringBuilder();
   }
 
   @Override
-  public void onOutputLine(String line) {
-    stringBuilder.append(line);
+  public synchronized void onOutputLine(String line) {
+    buffer.append(line);
```
briandealwis: Hmm, this implementation has a flaw in that multiple lines will be catenated together without a newline. Doesn't matter for JSON or single-line output.  Might be worth adding a note to the class javadoc to this effect rather than trying to fix this.

elharo: I thought maybe the nextLine method included the line terminator, but looking at the JavaDoc, nope. So that's another bug. Next PR, I think.

PR #271, opened 2016-12-03
Wait for output handlers and fix asyncRun() code

### LINKED ISSUE #266: Worth looking at syncRun() and asyncRun() regarding InterruptedException
#### Unnecessary Throw Declaration of `asyncRun()`

[`asyncRun()`](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/master/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/DefaultProcessRunner.java#L187) declares that it can throw `InterruptedException`, but in fact, it can never.

#### `InterruptedException` from `waitFor()`

Note that when `process.waitFor()` throws `InterruptedException`, it *does not* mean that the `process` is somehow terminated or "interrupted." Rather, it means that the thread executing `waitFor()` (which is the thread with the name `"wait-for-exit"`) has been interrupted.

So, I believe the following code in the `finally` block is a bit problematic too (or at least very confusing) when the thread is interrupted, because the target `process` will be still running.

```java
          } finally {
            int exitCode = process.exitValue();
            for (ProcessExitListener exitListener : exitListeners) {
              exitListener.onExit(exitCode);
            }
          }
```

The first problem is that [`process.exitValue()` will throw `IllegalThreadStateException` if the process is still running](https://docs.oracle.com/javase/7/docs/api/java/lang/Process.html#exitValue()). If terminating the `"wait-for-exit"` thread in this fashion in this case was really intended, I strongly suggest moving the above code in the `finally` block into the main `try` block, following `process.waitFor();`.

Honestly, however, I'm not sure if the thread `"wait-for-exit"` can ever be interrupted.

#### `InterruptedException` in `syncRun()`

If `waitFor()` is interrupted, it means the main thread executing the library is somehow interrupted. Currently, we eventually throw `ProcessRunnerException` in this case, which may make sense.

### LINKED ISSUE #269: Race condition in CloudSdk.getVersion
Tentative hypothesis: CloudSdk.getVersion sometimes returns before the stdOut line listener thread has started. This causes an IllegalArgumentException.

1. CloudSdk.getVersion calls runSynchronousGcloudCommand("gcloud info --format="value(basic.version)")
2. ProcessRunner runner = new DefaultProcessRunner(various args including the listeners)
3. runner.run(command.toArray(new String[command.size()]));
4. processBuilder.command(command);
    Process process = processBuilder.start();
5. handleStdOut(process);
6. stdOutThread.start();
7. syncRun(process)

  At this point the gcloud info command is off and running, and can return at any time. 
  However I see no guarantee that the thread started in step 6 has actually kicked off and is listening before the process from step 7 finishes. I think sometimes the process finishes before the stdOut thread starts, and thus we get no output on stdOut, which returns an empty string as the version and thus causes an IllegalArgumentException.

chanseokoh:
Fixes #269.
Fixes #266.

It will be much much easier to understand the PR if you look at the changes incrementally made by the two commits: https://github.com/GoogleCloudPlatform/appengine-plugins-core/commit/374d4677cdff11fc770244574e25df4f6b971e56, https://github.com/GoogleCloudPlatform/appengine-plugins-core/commit/53b48c5f6b72b6c1cde9d09856b3d18b89e2b8c9

The first commit https://github.com/GoogleCloudPlatform/appengine-plugins-core/commit/374d4677cdff11fc770244574e25df4f6b971e56 joins the output handlers for `syncRun()`. As discussed in #269, the output handlers will be able to consume all (buffered) output of the sub-process until EOF and eventually die gracefully, [*even if the handlers start running well after the sub-process dies*](https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/269#issuecomment-264614059) at least for small output. (I think it will also work for excessively large output, as long as the handlers keep consuming output, according to the [`Process` javadoc](https://docs.oracle.com/javase/7/docs/api/java/lang/Process.html):)

> Because some native platforms only provide limited buffer size for standard input and output streams, failure to promptly write the input stream or read the output stream of the subprocess may cause the subprocess to block, or even deadlock.

(As a side note, because we join the output handlers and ensure that access to the string built by the stdout handler happens only after the handler terminates, I believe the thread-safety fix in #270 is not really necessary. However, it's good to have the fix there.)

The second commit https://github.com/GoogleCloudPlatform/appengine-plugins-core/commit/53b48c5f6b72b6c1cde9d09856b3d18b89e2b8c9 fixes both #266 and #269 for `asyncRun()`, by actually invoking `syncRun()`.

I hope this fixes the issues, but there may be things that I am missing.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/DefaultProcessRunner.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/DefaultProcessRunner.java
@@ -183,6 +183,7 @@ public class DefaultProcessRunner implements ProcessRunner {
 
   private void syncRun(Process process, Thread stdOutThread, Thread stdErrThread)
       throws InterruptedException {
+    int exitCode = process.waitFor();
     // https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/269
     if (stdOutThread != null) {
       stdOutThread.join();
@@ -191,7 +192,6 @@ public class DefaultProcessRunner implements ProcessRunner {
       stdErrThread.join();
     }
 
-    int exitCode = process.waitFor();
     for (ProcessExitListener exitListener : exitListeners) {
       exitListener.onExit(exitCode);
     }
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/DefaultProcessRunner.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/DefaultProcessRunner.java
@@ -109,12 +109,14 @@ public class DefaultProcessRunner implements ProcessRunner {
 
       Process process = processBuilder.start();
 
+      Thread stdOutHandler = null;
+      Thread stdErrHandler = null;
       // Only handle stdout or stderr if there are listeners.
       if (!stdOutLineListeners.isEmpty()) {
-        handleStdOut(process);
+        stdOutHandler = handleStdOut(process);
       }
       if (!stdErrLineListeners.isEmpty()) {
-        handleErrOut(process);
+        stdErrHandler = handleErrOut(process);
       }
 
       for (ProcessStartListener startListener : startListeners) {
@@ -125,7 +127,7 @@ public class DefaultProcessRunner implements ProcessRunner {
         asyncRun(process);
       } else {
         shutdownProcessHook(process);
-        syncRun(process);
+        syncRun(process, stdOutHandler, stdErrHandler);
       }
 
     } catch (IOException | InterruptedException | IllegalThreadStateException e) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/DefaultProcessRunner.java
```diff
@@ -175,29 +178,36 @@ public void run() {
     };
     stdErrThread.setDaemon(true);
     stdErrThread.start();
+    return stdErrThread;
   }
 
-  private void syncRun(final Process process) throws InterruptedException {
+  private void syncRun(Process process, Thread stdOutThread, Thread stdErrThread)
+      throws InterruptedException {
+    // https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/269
```
elharo: Should we call process.waitFor *before* we join to these threads?

chanseokoh: Alright, it's conceptually better.

Issue #273, opened 2016-12-05
Consider making getVersion() a straight blocking call


chanseokoh:
See https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/269#issuecomment-264614059 and the following discussions.

chanseokoh:
Invalidated by #276. Closing.

PR #276, opened 2016-12-06
Improve CloudSdk.getVersion

aslo:
This PR makes a few enhancements to CloudSdk.getVersion that are required in order for the intellij plugin to use the version-compatibility logic in this library and not need to perform an expensive operation twice. (see https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/1127 for more context).

Quick summary of changes:
- Cloud SDK version is now determined by reading the VERSION file at the root of the cloud sdk installation, rather than calling gcloud in a subprocess. This is more performant and the code is less complex. Note that this means we can't determine the version of any cloud sdk installation that is earlier than 133.0.0 (when VERSION files first started being used)
- CloudSdkOutOfDateException no longer inherits from CloudSdkNotFoundException. This makes it a lot easier to validate based on these exceptions. 
- CloudSdkOutOfDateException now has the minimum required version as a member field. This allows clients of this library to construct their own validation messages using the data from the exception.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -287,10 +287,8 @@ public class CloudSdk {
     Path versionFile = getSdkPath().resolve(VERSION_FILE);
 
     if (!Files.isRegularFile(versionFile)) {
-      throw new CloudSdkOutOfDateException("Cloud SDK version is out of date. Please update to "
-          + "at least " + MINIMUM_VERSION, MINIMUM_VERSION);
+      throw new CloudSdkOutOfDateException(MINIMUM_VERSION);
     }
-    // TODO charset?
     List<String> lines = Files.readAllLines(versionFile, StandardCharsets.UTF_8);
     // expect only a single line
     String contents = lines.get(0);
@@ -298,8 +296,7 @@ public class CloudSdk {
     try {
       return new CloudSdkVersion(contents);
     } catch (IllegalArgumentException e) {
-      throw new CloudSdkOutOfDateException("Cloud SDK version is out of date. Please update to "
-          + "at least " + MINIMUM_VERSION, MINIMUM_VERSION);
+      throw new CloudSdkOutOfDateException(MINIMUM_VERSION);
     }
   }
 
@@ -391,12 +388,11 @@ public class CloudSdk {
     validateCloudSdkVersion();
   }
 
-  private void validateCloudSdkVersion() {
+  private void validateCloudSdkVersion() throws CloudSdkOutOfDateException {
     try {
       CloudSdkVersion version = getVersion();
       if (version.compareTo(MINIMUM_VERSION) < 0) {
-        throw new CloudSdkOutOfDateException("Cloud SDK version " + version
-            + " is too old. Please update to at least " + MINIMUM_VERSION, MINIMUM_VERSION);
+        throw new CloudSdkOutOfDateException(MINIMUM_VERSION);
       }
     } catch (IllegalArgumentException | IOException ex) {
       throw new CloudSdkNotFoundException("Could not determine Cloud SDK version", ex);
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -59,7 +59,9 @@ import javax.annotation.Nullable;
  * Cloud SDK CLI wrapper.
  */
 public class CloudSdk {
-  private static final CloudSdkVersion MINIMUM_VERSION = new CloudSdkVersion("133.0.0");
+
+  public static final CloudSdkVersion MINIMUM_VERSION = new CloudSdkVersion("133.0.0");
+
   private static final Logger logger = Logger.getLogger(CloudSdk.class.toString());
   private static final Joiner WHITESPACE_JOINER = Joiner.on(" ");
 
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -36,6 +36,8 @@ import com.google.common.collect.Maps;
 import com.google.gson.JsonSyntaxException;
 
 import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
@@ -57,7 +59,7 @@ import javax.annotation.Nullable;
  * Cloud SDK CLI wrapper.
  */
 public class CloudSdk {
-  private static final int MINIMUM_VERSION = 131;
+  private static final CloudSdkVersion MINIMUM_VERSION = new CloudSdkVersion("133.0.0");
   private static final Logger logger = Logger.getLogger(CloudSdk.class.toString());
   private static final Joiner WHITESPACE_JOINER = Joiner.on(" ");
 
@@ -69,6 +71,7 @@ public class CloudSdk {
   private static final String JAVA_TOOLS_JAR = "appengine-tools-api.jar";
   private static final Map<String, Path> JAR_LOCATIONS = new HashMap<>();
   private static final String WINDOWS_BUNDLED_PYTHON = "platform/bundledpython/python.exe";
+  private static final String VERSION_FILE = "VERSION";
 
   private final Path sdkPath;
   private final ProcessRunner processRunner;
@@ -274,19 +277,28 @@ public class CloudSdk {
   }
 
   /**
-   * Returns the version of the Cloud SDK installation. Unlike other methods in this class that call
-   * gcloud, this method always uses a synchronous ProcessRunner and will block until the gcloud
-   * process returns.
+   * Returns the version of the Cloud SDK installation.
    *
-   * @throws ProcessRunnerException when process runner encounters an error
+   * @throws CloudSdkOutOfDateException if the Cloud SDK is too out of date to determine its version
    */
-  public CloudSdkVersion getVersion() throws ProcessRunnerException {
-    // gcloud info --format="value(basic.version)"
-    List<String> command = new ImmutableList.Builder<String>().add("info")
-        .addAll(GcloudArgs.get("format", "value(basic.version)")).build();
+  public CloudSdkVersion getVersion() throws IOException {
+    Path versionFile = getSdkPath().resolve(VERSION_FILE);
 
-    String output = runSynchronousGcloudCommand(command);
-    return new CloudSdkVersion(output);
+    if (!Files.isRegularFile(versionFile)) {
+      throw new CloudSdkOutOfDateException("Cloud SDK version is out of date. Please update to "
+          + "at least " + MINIMUM_VERSION, MINIMUM_VERSION);
+    }
+    // TODO charset?
+    List<String> lines = Files.readAllLines(versionFile, StandardCharsets.UTF_8);
+    // expect only a single line
+    String contents = lines.get(0);
+
+    try {
+      return new CloudSdkVersion(contents);
+    } catch (IllegalArgumentException e) {
+      throw new CloudSdkOutOfDateException("Cloud SDK version is out of date. Please update to "
+          + "at least " + MINIMUM_VERSION, MINIMUM_VERSION);
+    }
   }
 
   /**
@@ -369,9 +381,10 @@ public class CloudSdk {
   /**
    * Checks whether the Cloud SDK path and version are valid.
    *
-   * @throws CloudSdkNotFoundException when an up-to-date Cloud SDK is not installed where expected
+   * @throws CloudSdkNotFoundException when Cloud SDK is not installed where expected
+   * @throws CloudSdkOutOfDateException when Cloud SDK is out of date
    */
-  public void validateCloudSdk() throws CloudSdkNotFoundException {
+  public void validateCloudSdk() throws CloudSdkNotFoundException, CloudSdkOutOfDateException {
     validateCloudSdkLocation();
     validateCloudSdkVersion();
   }
@@ -379,11 +392,11 @@ public class CloudSdk {
   private void validateCloudSdkVersion() {
     try {
       CloudSdkVersion version = getVersion();
-      if (version.getMajorVersion() < MINIMUM_VERSION) {
+      if (version.compareTo(MINIMUM_VERSION) <= 0) {
         throw new CloudSdkOutOfDateException("Cloud SDK version " + version
-            + " is too old. Please update to at least " + MINIMUM_VERSION);
+            + " is too old. Please update to at least " + MINIMUM_VERSION, MINIMUM_VERSION);
       }
-    } catch (ProcessRunnerException | IllegalArgumentException ex) {
+    } catch (IllegalArgumentException | IOException ex) {
       throw new CloudSdkNotFoundException("Could not determine Cloud SDK version", ex);
     }
   }
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -36,6 +36,7 @@ import com.google.common.collect.Maps;
 import com.google.gson.JsonSyntaxException;
 
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
@@ -72,7 +73,7 @@ public class CloudSdk {
       "platform/google_appengine/google/appengine/tools/java/lib";
   private static final String JAVA_TOOLS_JAR = "appengine-tools-api.jar";
   private static final String WINDOWS_BUNDLED_PYTHON = "platform/bundledpython/python.exe";
-  private static final String VERSION_FILE = "VERSION";
+  private static final String VERSION_FILE_NAME = "VERSION";
 
   private final Map<String, Path> jarLocations = new HashMap<>();
   private final Path sdkPath;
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -274,19 +279,28 @@ public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
...
-    // gcloud info --format="value(basic.version)"
-    List<String> command = new ImmutableList.Builder<String>().add("info")
-        .addAll(GcloudArgs.get("format", "value(basic.version)")).build();
+  public CloudSdkVersion getVersion() throws IOException {
+    Path versionFile = getSdkPath().resolve(VERSION_FILE);
 
-    String output = runSynchronousGcloudCommand(command);
-    return new CloudSdkVersion(output);
+    if (!Files.isRegularFile(versionFile)) {
+      throw new CloudSdkOutOfDateException("Cloud SDK version is out of date. Please update to "
```
patflynn: I'd change this warning to 'Cloud SDK versions below %MINIMUM_VERSION% are not supported by this library.'

'Cloud SDK version is 'out of date'' is too general in the sense that it could be out of date but supported.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -274,19 +279,28 @@ public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
...
+          + "at least " + MINIMUM_VERSION, MINIMUM_VERSION);
+    }
+    // TODO charset?
```
patflynn: I'm not sure what this todo means.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -369,21 +383,22 @@ public Path getJarPath(String jarName) {
...
     try {
       CloudSdkVersion version = getVersion();
-      if (version.getMajorVersion() < MINIMUM_VERSION) {
+      if (version.compareTo(MINIMUM_VERSION) < 0) {
         throw new CloudSdkOutOfDateException("Cloud SDK version " + version
-            + " is too old. Please update to at least " + MINIMUM_VERSION);
+            + " is too old. Please update to at least " + MINIMUM_VERSION, MINIMUM_VERSION);
       }
-    } catch (ProcessRunnerException | IllegalArgumentException ex) {
+    } catch (IllegalArgumentException | IOException ex) {
```
patflynn: perhaps we need to distinguish between the expected validation exception of being unsupported by the caller, and the exception of being unsupported by the core library.  And the messages should be built into the exception types with just the parameters passed into the constructor.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
aslo: :+1:

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -57,7 +59,9 @@
  * Cloud SDK CLI wrapper.
  */
 public class CloudSdk {
-  private static final int MINIMUM_VERSION = 131;
+
+  public static final CloudSdkVersion MINIMUM_VERSION = new CloudSdkVersion("133.0.0");
```
chanseokoh: `private`?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -274,19 +279,28 @@ public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
   }
 
   /**
-   * Returns the version of the Cloud SDK installation. Unlike other methods in this class that call
-   * gcloud, this method always uses a synchronous ProcessRunner and will block until the gcloud
-   * process returns.
+   * Returns the version of the Cloud SDK installation.
    *
-   * @throws ProcessRunnerException when process runner encounters an error
+   * @throws CloudSdkOutOfDateException if the Cloud SDK is too out of date to determine its version
```
chanseokoh: Javadoc about `@throws IOException`? Or, does it make sense to throw `IOException` from a client's perspective?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -274,19 +279,28 @@ public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
...
-    String output = runSynchronousGcloudCommand(command);
-    return new CloudSdkVersion(output);
+    if (!Files.isRegularFile(versionFile)) {
+      throw new CloudSdkOutOfDateException("Cloud SDK version is out of date. Please update to "
+          + "at least " + MINIMUM_VERSION, MINIMUM_VERSION);
+    }
+    // TODO charset?
+    List<String> lines = Files.readAllLines(versionFile, StandardCharsets.UTF_8);
+    // expect only a single line
+    String contents = lines.get(0);
```
chanseokoh: It shouldn't happen, so I'm not sure but this might throw `IndexOutOfBoundsException`.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -274,19 +279,28 @@ public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
...
+    Path versionFile = getSdkPath().resolve(VERSION_FILE);
 
-    String output = runSynchronousGcloudCommand(command);
-    return new CloudSdkVersion(output);
+    if (!Files.isRegularFile(versionFile)) {
+      throw new CloudSdkOutOfDateException("Cloud SDK version is out of date. Please update to "
+          + "at least " + MINIMUM_VERSION, MINIMUM_VERSION);
+    }
+    // TODO charset?
+    List<String> lines = Files.readAllLines(versionFile, StandardCharsets.UTF_8);
```
aslo: Actually, nevermind. I believe UTF-8 should work in either case, since the same file is present in either a windows or linux installation of the cloud sdk. I'll test on windows to confirm.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
aslo: See my above comment on the line below this one.

chanseokoh:
Closes #273.

Remove `StringBuilderProcessOutputLineListener` and `runGcloudSynchronously()` (which is private)?

chanseokoh:
NVM, Is see `runSynchronousGcloudCommand()` is used in `getComponents()`.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
chanseokoh: Got it.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
chanseokoh: Sounds good. Also, a client may wonder why a seemingly trivial call `getVersion()` can throw `IOException`, so a brief mention that it tries to read a VERSION file would be nice in Javadoc.

aslo: [Here](https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/276/files#diff-562793416b98626facf7820af2a81c16R286) is what I went with. Looks good?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -68,8 +72,9 @@
       "platform/google_appengine/google/appengine/tools/java/lib";
   private static final String JAVA_TOOLS_JAR = "appengine-tools-api.jar";
   private static final String WINDOWS_BUNDLED_PYTHON = "platform/bundledpython/python.exe";
-  private final Map<String, Path> jarLocations = new HashMap<>();
+  private static final String VERSION_FILE = "VERSION";
```
patflynn: VERSION_FILE_NAME

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkOutOfDateException.java
```diff
@@ -16,13 +16,31 @@
...
- * The Cloud SDK that was found is too old (generally before 131.0).
+ * The Cloud SDK that was found is too old (generally before 133.0.0).
  */
-public class CloudSdkOutOfDateException extends CloudSdkNotFoundException {
+public class CloudSdkOutOfDateException extends AppEngineException {
+
+  private static final String MESSAGE
+      = "Cloud SDK versions below %s are not supported by this library.";
+
+  private CloudSdkVersion requiredVersion;
```
patflynn: private final

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -274,19 +279,32 @@ public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
...
+    List<String> lines = Files.readAllLines(versionFile, StandardCharsets.UTF_8);
+    if (lines.size() > 0) {
+      // expect only a single line
+      contents = lines.get(0);
+    }
+
+    try {
+      return new CloudSdkVersion(contents);
+    } catch (IllegalArgumentException e) {
+      throw new CloudSdkOutOfDateException(MINIMUM_VERSION, e);
```
patflynn: This case doesn't seem like the exception is appropriate to the error.  Isn't the problem here that for some weird reason the contents of the file is not a valid Cloud SDK version string as far as we know?  Perhaps we should just throw IllegalStateException?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -274,19 +279,32 @@ public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
...
-    return new CloudSdkVersion(output);
+    if (!Files.isRegularFile(versionFile)) {
+      throw new CloudSdkOutOfDateException(MINIMUM_VERSION);
+    }
+
+    String contents = "";
+    List<String> lines = Files.readAllLines(versionFile, StandardCharsets.UTF_8);
+    if (lines.size() > 0) {
+      // expect only a single line
+      contents = lines.get(0);
```
chanseokoh: If `lines.size() == 0` (or could do `!lines.isEmpty()`), it will reach`return new CloudSdkVersion(contents);`. If `contents` is am empty string, the `Precondition` check inside will throw `IllegalArgumentException`.

Maybe it's simpler to catch `IndexOutOfBoundsException` in the following `try` block?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
chanseokoh: This may also be relevant: https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/276/files#diff-562793416b98626facf7820af2a81c16R405

chanseokoh: Okay, the link is broken. So, in this PR, the new code is

```java
  private void validateCloudSdkVersion() throws CloudSdkOutOfDateException {
    try {
      CloudSdkVersion version = getVersion();
      if (version.compareTo(MINIMUM_VERSION) < 0) {
        throw new CloudSdkOutOfDateException(MINIMUM_VERSION);
      }
    } catch (IllegalArgumentException | IOException ex) {
      throw new CloudSdkNotFoundException("Could not determine Cloud SDK version", ex);
    }
  }
```

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -274,19 +279,32 @@ public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
...
 
   /**
-   * Returns the version of the Cloud SDK installation. Unlike other methods in this class that call
-   * gcloud, this method always uses a synchronous ProcessRunner and will block until the gcloud
-   * process returns.
+   * Returns the version of the Cloud SDK installation. Version is determined by reading the VERSION
+   * file located in the Cloud SDK directory.
    *
-   * @throws ProcessRunnerException when process runner encounters an error
+   * @throws CloudSdkOutOfDateException if the Cloud SDK is too out of date to determine its version
```
joaoandremartins: I'm not sure if CloudSdkOutOfDateException conveys when the exception is thrown. Maybe `NoVersionFileException` instead?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
aslo: :+1:

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -369,21 +387,21 @@ public Path getJarPath(String jarName) {
...
     try {
       CloudSdkVersion version = getVersion();
-      if (version.getMajorVersion() < MINIMUM_VERSION) {
-        throw new CloudSdkOutOfDateException("Cloud SDK version " + version
-            + " is too old. Please update to at least " + MINIMUM_VERSION);
+      if (version.compareTo(MINIMUM_VERSION) < 0) {
+        throw new CloudSdkOutOfDateException(MINIMUM_VERSION);
       }
-    } catch (ProcessRunnerException | IllegalArgumentException ex) {
+    } catch (IllegalArgumentException | IOException ex) {
```
chanseokoh: I think we now catch `IllegalArgumentException` inside `getVersion()`, we don't need to catch it here.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
aslo: Just so we're all on the same page, let's discuss the all the possible exception cases so we can use a consistent strategy. Here my new proposal for all cases:

### in CloudSdk.getVersion
1. The version file does not exist:
Throw a CloudSdkOutOfDateException. It's too old to even know what version it is.
2. The version file exists, but we gen an IOException when reading it:
throw IOException
3. The version file exists, but its contents are not parseable:
throw IllegalStateException

### in CloudSdk.validateCloudSdkVersion
1. CloudSdk.getVersion throws a CloudSdkOutOfDateException:
do nothing; callers deal with the CloudSdkOutOfDateException
2. CloudSdk.getVersion throws anything else
wrap in a CloudSdkOutOfDateException. Within the context of specifically only validating the version, perhaps it doesn't make sense to propagate all of the above exception types to the caller, but to just consider this a case where the version is wrong.

@chanseokoh @patflynn What do you think?

joaoandremartins: As I said on the code review, I'd have two exceptions, CloudSdkOutOfDateException and VersionFileNotFoundException. getVersion() would return a VersionFileNotFoundException instead of a CloudSdkOutOfDateException for the case where the file doesn't exist.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
aslo: Thanks for pointing out. Please take a look at my above [comment](https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/276#discussion_r91149616) for an overall proposal for the exception types to be used in this PR.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
patflynn: I also think that the file not existing is a different class of error from the validation exception.

The IO and ISE exceptions on getVersion I would propagate to clients but as runtime exceptions that cause unexpected errors. If the ISE ever happens I would want it reported in our feedback.

etanshaul:
Have you had a chance to test out the performance of this (is the noticeable slow down gone in the client UIs)?

aslo:
Yep. These changes, plus my changes to the intellij plugin (PR coming) make it so there is no noticeable lag in the UI.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
chanseokoh: Just don't forget `validateCloudSdkVersion()` is a private method. `validateCloudSdk()` is the public-facing method.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkVersionFileNotFoundException.java
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.appengine.cloudsdk;
+
+import com.google.cloud.tools.appengine.api.AppEngineException;
+
+public class CloudSdkVersionFileNotFoundException extends AppEngineException {
```
patflynn: javadoc

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -282,28 +283,31 @@ public void runAppCfgCommand(List<String> args) throws ProcessRunnerException {
...
+   * @throws IllegalStateException if the file content could not be parsed
    */
-  public CloudSdkVersion getVersion() throws IOException {
-    Path versionFile = getSdkPath().resolve(VERSION_FILE);
+  public CloudSdkVersion getVersion() {
+    Path versionFile = getSdkPath().resolve(VERSION_FILE_NAME);
 
     if (!Files.isRegularFile(versionFile)) {
-      throw new CloudSdkOutOfDateException(MINIMUM_VERSION);
+      throw new CloudSdkVersionFileNotFoundException("Cloud SDK version file not found at "
```
etanshaul: I may be missing something but I don't see much point to this exception if its just thrown and caught internally in this class.

etanshaul:
sounds good

PR #277, opened 2016-12-06
Makes JAR_LOCATIONS not static.

joaoandremartins:
When many CloudSdk objects with different sdkPaths are created in quick succession, the values in JAR_LOCATIONS take a little while to update. This is creating some weirdness in IntelliJ when specifying different sdkPaths and validating them. All validations succeed, except for `if (!Files.isRegularFile(jarLocations.get(JAVA_TOOLS_JAR))) {` which still considers the old stored static sdkPath value and fails.

The visible impact of this in IntelliJ is that when an invalid path is changed to a valid one, the warning message doesn't disappear.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -137,7 +137,6 @@ public class CloudSdk {
       command.add(commandToken);
     }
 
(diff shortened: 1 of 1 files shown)


joaoandremartins:
The problem is that the contents of the HashMap are dynamic (because of the SDK path that is only given at run time), so we can't initialize the variable correctly statically. But that is a good point, anyways.

PR #278, opened 2016-12-06
Remove unnecessary exception catch

chanseokoh:
[The only method that can throw `IllegalThreadStateException` is `exitValue()`.](https://docs.oracle.com/javase/7/docs/api/java/lang/Process.htm) We don't have `exitValue()`.


aslo:
:+1:

PR #281, opened 2016-12-08
Update test retrigger

loosebazooka:


loosebazooka:
This is just #228 which had problems with travis not triggering.

loosebazooka:
@meltsufin @patflynn @elharo

PR #283, opened 2016-12-16
Update README.md

elharo:
@patflynn don't need to login at command line for this library to work


chanseokoh:
@patflynn Yeah, I think so too. AFAIK, the Gradle and Maven appengine plugins use gcloud creds.

elharo:
This change doesn't say you can't login; just removes the incorrect statement that you have to. How and whether to login is properly part of the docs for the clients of this library--e.g. CT4E, CT4I, maven plugin, etc. This is the wrong place to include that information.

PR #284, opened 2016-12-16
supply environment variables

elharo:
@briandealwis @chanseokoh

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
@@ -274,8 +274,7 @@ public class DefaultRunConfiguration implements RunConfiguration {
     for (String key : environment.keySet()) {
       if (key.indexOf('\0') > -1) {
         throw new IllegalArgumentException("Environment variable name " + key + " contains a NUL");
-      }
-      if (key.indexOf('\0') > -1 || key.indexOf('=') > -1) {
+      } else if (key.indexOf('=') > -1) {
         throw new IllegalArgumentException("Environment variable name "
             + key + " must not contain =.");
       }
--- a/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
@@ -23,7 +23,7 @@ import java.util.Map;
 import com.google.common.collect.ImmutableMap;
 
 /**
- * Plain Java bean implementation of {@link RunConfiguration}.
+ * Plain Java bean implementation of {@link RunConfiguration}. This class is not thread safe.
  */
 public class DefaultRunConfiguration implements RunConfiguration {
 
@@ -50,7 +50,7 @@ public class DefaultRunConfiguration implements RunConfiguration {
   private Boolean skipSdkUpdateCheck;
   private String defaultGcsBucketName;
   private String javaHomeDir;
-  private ImmutableMap<String, String> environment;
+  private ImmutableMap<String, String> environment = ImmutableMap.of();
 
   @Override
   public List<File> getAppYamls() {
@@ -268,8 +268,19 @@ public class DefaultRunConfiguration implements RunConfiguration {
    * Makes an immutable copy of the specified map. 
    * 
    * @param environment the environment variables to store.
+   * @throws NullPointerException if any key or value in environment is null
+   * @throws IllegalArgumentException if any key contains the NUL or = character 
    */
   public void setEnvironmentVariables(Map<String, String> environment) {
+    for (String key : environment.keySet()) {
+      if (key.indexOf('\0') > -1) {
+        throw new IllegalArgumentException("Environment variable name " + key + " contains a NUL");
+      }
+      if (key.indexOf('\0') > -1 || key.indexOf('=') > -1) {
+        throw new IllegalArgumentException("Environment variable name "
+            + key + " must not contain =.");
+      }
+    }
     this.environment = ImmutableMap.copyOf(environment);
   }
 }
--- a/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
@@ -47,6 +47,7 @@ public class DefaultRunConfiguration implements RunConfiguration {
   private Boolean skipSdkUpdateCheck;
   private String defaultGcsBucketName;
   private String javaHomeDir;
+  private List<String> environment;
 
   @Override
   public List<File> getAppYamls() {
@@ -253,5 +254,14 @@ public class DefaultRunConfiguration implements RunConfiguration {
 
   public void setJavaHomeDir(String javaHomeDir) {
     this.javaHomeDir = javaHomeDir;
+  }
+
+  @Override
+  public List<String> getEnvironmentVariables() {
+    return environment;
+  }
+ 
+  public void setEnvironmentVariables(List<String> environment) {
+    this.environment = environment;
   }
 }
--- a/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
@@ -16,11 +16,10 @@
 
 package com.google.cloud.tools.appengine.api.devserver;
 
+import com.google.common.collect.ImmutableMap;
 import java.io.File;
 import java.util.List;
 import java.util.Map;
-
-import com.google.common.collect.ImmutableMap;
 
 /**
  * Plain Java bean implementation of {@link RunConfiguration}. This class is not thread safe.
--- a/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
@@ -18,6 +18,9 @@ package com.google.cloud.tools.appengine.api.devserver;
 
 import java.io.File;
 import java.util.List;
+import java.util.Map;
+
+import com.google.common.collect.ImmutableMap;
 
 /**
  * Plain Java bean implementation of {@link RunConfiguration}.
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer.java
```diff
@@ -64,11 +65,19 @@ public void run(RunConfiguration config) throws AppEngineException {
...
+    Map<String,String> environmentVariables = Maps.newHashMap();
     if (!Strings.isNullOrEmpty(config.getJavaHomeDir())) {
-      env.put("JAVA_HOME", config.getJavaHomeDir());
+      environmentVariables.put("JAVA_HOME", config.getJavaHomeDir());
     }
 
+    List<String> environment = config.getEnvironmentVariables();
+    if (environment != null) {
+      for (String nameValue : environment) {
+        String[] pair = nameValue.split("\\=");
```
chanseokoh: Is escaping necessary in `"\\="`? Before that, I wonder if we can't get a `Map<String, String>` from the first place to avoid splitting.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer.java
```diff
@@ -23,6 +23,7 @@
 import com.google.cloud.tools.appengine.cloudsdk.internal.args.DevAppServerArgs;
 import com.google.cloud.tools.appengine.cloudsdk.internal.process.ProcessRunnerException;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Splitter;
```
chanseokoh: I don't think this is used.

chanseokoh:
@meltsufin @loosebazooka With this, lib clients can provide any env variable. Cool?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer.java
elharo: done

elharo: done

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer.java
elharo: done

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer.java
```diff
@@ -64,11 +64,19 @@ public void run(RunConfiguration config) throws AppEngineException {
...
+    Map<String,String> environmentVariables = Maps.newHashMap();
     if (!Strings.isNullOrEmpty(config.getJavaHomeDir())) {
-      env.put("JAVA_HOME", config.getJavaHomeDir());
+      environmentVariables.put("JAVA_HOME", config.getJavaHomeDir());
     }
 
+    List<String> environment = config.getEnvironmentVariables();
+    if (environment != null) {
+      for (String nameValue : environment) {
+        String[] pair = nameValue.split("=");
```
briandealwis: An environment variable value is allowed to have `=`. I agree with @chanseokoh — push the onus of splitting to the library consumers.

elharo:
PTAL @chanseokoh @briandealwis converted to a map instead of a list; added more verification and tests.

src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
```diff
@@ -254,4 +257,29 @@ public String getJavaHomeDir() {
...
+   * Makes an immutable copy of the specified map. 
+   * 
+   * @param environment the environment variables to store.
+   * @throws NullPointerException if any key or value in environment is null
+   * @throws IllegalArgumentException if any key contains the NUL or = character 
+   */
+  public void setEnvironmentVariables(Map<String, String> environment) {
+    for (String key : environment.keySet()) {
+      if (key.indexOf('\0') > -1) {
+        throw new IllegalArgumentException("Environment variable name " + key + " contains a NUL");
```
chanseokoh: Your preference, but you can use
```java
Preconditions.checkArgument(key.indexOf('\0') == -1,
    "Environment variable name %s contains a NUL", key);
```

src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
```diff
@@ -254,4 +257,29 @@ public String getJavaHomeDir() {
...
+      }
+      if (key.indexOf('\0') > -1 || key.indexOf('=') > -1) {
```
chanseokoh: I think `key.indexOf('\0') > -1` doesn't go with the exception message below.

src/test/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfigurationTest.java
```diff
@@ -0,0 +1,64 @@
+package com.google.cloud.tools.appengine.api.devserver;
+
+import java.util.HashMap;
+import java.util.Map;
+
```
chanseokoh: Blank line to be removed per Google style.

src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
```diff
@@ -254,4 +257,29 @@ public String getJavaHomeDir() {
...
+  }
+ 
+  /**
+   * Makes an immutable copy of the specified map. 
+   * 
+   * @param environment the environment variables to store.
+   * @throws NullPointerException if any key or value in environment is null
+   * @throws IllegalArgumentException if any key contains the NUL or = character 
+   */
+  public void setEnvironmentVariables(Map<String, String> environment) {
```
chanseokoh: `addEnvironmentVariable(s)` suddenly came up in my mind, but `setEnvironmentVariables` LGTM.

elharo: set > add because it replaces the entire map, not simply adds variables to the existing map

src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
elharo: 1. &gt; sometimes more reliable than ==
2. I know some devs like Preconditions, but &lt;redacted&gt;talk to me offline if you're curious&lt;/redacted>

src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
elharo: oops, done

src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
```diff
@@ -254,4 +257,28 @@ public String getJavaHomeDir() {
...
+    for (String key : environment.keySet()) {
+      if (key.indexOf('\0') > -1) {
```
briandealwis: What's the motivation for this NUL check?

src/main/java/com/google/cloud/tools/appengine/api/devserver/RunConfiguration.java
```diff
@@ -53,6 +54,13 @@
 
   List<String> getJvmFlags();
 
+  /**
+   * Return an immutable map of environment variable strings.
+   * 
+   * @return the environment variables
+   */
+  ImmutableMap<String, String> getEnvironmentVariables();
```
briandealwis: Seems strange to see `Immutable` as part of the signature.

src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
briandealwis: Should this allow `null` to pass through the current environment?

elharo: No, Immutable map doesn't allow nulls and I can't see a reason we need them.

src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
elharo: It's one of two characters not allowed in Unix environment variable names, the other being =

src/main/java/com/google/cloud/tools/appengine/api/devserver/RunConfiguration.java
elharo: What's strange is that ImmutableMap is a subclass of Map rather than the other way around. It's a violation of the Liskov substitution principal, but that's Java for you. :-(

joaoandremartins: Since ImmutableMap is a subclass of Map, this should be Map, no?

joaoandremartins:
What is the motivation for dynamically adding new environment variables?

What env vars do we want to add? If it's just a couple, and the rest of the tools could use them as well, should we just add them statically?

src/main/java/com/google/cloud/tools/appengine/api/devserver/RunConfiguration.java
elharo: No, it shouldn't. An ImmutableMap is not a Map and cannot be used as one since it doesn't implement the write operations. This is the violation of the Liskov substitution principal I mentioned above. IMHO this is a design flaw in the Java Collections API, but it is not one we can fix here.

patflynn:
Does it make sense to allow the user to set environment variables this way for the dev appserver when they can't do so in prod?  Typically appengine-web.xml is the way they go about doing it.

briandealwis:
It can be useful to set env vars for debugging/testing.  There's a danger here, however, as I doubt these values will override those specified in the `appengine-web.xml`.

elharo:
I think @patflynn's point is convincing. I'm inclined to revert this PR, and remove the environment tab from the eclipse side.

joaoandremartins:
There's also a similar tab in IntelliJ where I think we're currently ignoring all environment flags (except for the debug stuff) for the devappserver. We should also probably look into removing it.

PR #286, opened 2017-01-02
Fix CloudSdkNotFoundException

frantello:
Update path resolver for look at /usr/bin directory. This patch fix CloudSdkNotFoundException related to official https://cloud.google.com/sdk/downloads Ubuntu/Debian apt-get install method. After installation, binaries are located at common /usr/bin/ directory.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
@@ -50,6 +50,8 @@ public class PathResolver implements CloudSdkResolver {
     } else {
       // home directory
       possiblePaths.add(System.getProperty("user.home") + "/google-cloud-sdk");
+      // usr directory
+      possiblePaths.add("/usr");
       // try devshell VM
       possiblePaths.add("/google/google-cloud-sdk");
       // try bitnami Jenkins VM:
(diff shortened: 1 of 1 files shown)


frantello:
I signed it!

src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
```diff
@@ -50,6 +50,8 @@ public Path getCloudSdkPath() {
     } else {
       // home directory
       possiblePaths.add(System.getProperty("user.home") + "/google-cloud-sdk");
+      // usr directory
+      possiblePaths.add("/usr");
```
elharo: I think what you want here is /usr/bin, not /usr. And that said what we need here is the path to the actual google-cloud-sdk directory such as /usr/local/google-cloud-sdk, not simply a path to symlinks the installer may set up elsewhere. These paths are used for more than just the binaries. 

I don't have a Debian system handy to test this with. Where is the installer putting the google-cloud-sdk directory on your system?

frantello: Thanks @elharo 

Checked with my Ubuntu 14.04.5 installation, link points to:
/usr/bin/gcloud -> /usr/lib/google-cloud-sdk/bin/gcloud

I added just "/usr" because "/bin" is added after programmatically. Path must match SDK directory, and not the bin directory.

As you said, correct possible path should be /usr/lib/google-cloud-sdk/

PR #287, opened 2017-01-02
add exception javadoc

### LINKED ISSUE #279: Update Javadoc about CloudSdkNotFoundException and CloudSdkOutOfDateException
There are several public-facing methods that (directly or indirectly) call `validateCloudSdk()`:

`runAppCommand()`
`runSourceCommand()`
`getComponents()`
`runDevAppServerCommand(List)`
`runDevAppServerCommand(List, Map<>)`

The last two methods document that they throw `CloudSdkNotFoundException`:

`@throws CloudSdkNotFoundException when the Cloud SDK is not installed where expected`

However, the others don't document it. Moreover, after #276, all of them will additionally throw `CloudSdkOutOfDateException`, so this should be documented too.

elharo:
~~fix #279~~ @chanseokoh

UPDATE(@chanseokoh): helps with #279.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -36,7 +36,6 @@ import com.google.common.collect.Maps;
 import com.google.gson.JsonSyntaxException;
 
 import java.io.File;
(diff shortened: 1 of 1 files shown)


chanseokoh:
I updated your original PR comment, so merging will not auto-close #279.

PR #294, opened 2017-01-05
Add shlex utility

elharo:
@chanseokoh @briandealwis 

Please suggest bugs or suggestion for additional features in the form of test cases not currently handled. E.g.

foo='bar baz'
or
foo="bar\"baz"

I'm not sure how far we need to go with this for version 1.

### DIFF
--- a/src/main/java/com/google/cloud/tools/shell/CommandLine.java
+++ b/src/main/java/com/google/cloud/tools/shell/CommandLine.java
@@ -15,21 +15,24 @@ package com.google.cloud.tools.shell;
 
 import java.util.ArrayList;
 
+import com.google.common.annotations.Beta;
+
 /**
  * Utilities for working with command line arguments.
  */
+@Beta
 public class CommandLine {
 
   /**
    * Split a full command line into an array of individual arguments,
-   * similar to the shlex function in Python.
+   * similar to the shlex function in Python according to POSIX rules. 
+   * All input characters are preserved except for separating whitespace. 
+   * This function tokenizes the input string, but does not attempt to parse it.
    * 
    * @param line the input line
    * @return a non-null but possibly empty array of arguments
    */
-  public static String[] split(String line) {
-    String[] empty = new String[0];
-    
+  public static String[] split(String line) {    
     char quote = '"';
     boolean quoted = false;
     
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/cloud/tools/shell/CommandLine.java
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * Split a full command line into an array of individual arguments,
+   * similar to the shlex function in Python according to POSIX rules. 
+   * All input characters are preserved except for separating whitespace. 
+   * This function tokenizes the input string, but does not attempt to parse it.
+   * For example, it recognizes that a space escaped with a backslash does not
+   * start a new token, but it does not replace the "\ " with a single space. 
+   * 
+   * @param line the input line
```
chanseokoh: Add "cannot be null"?

src/main/java/com/google/cloud/tools/shell/CommandLine.java
```diff
@@ -0,0 +1,80 @@
...
+   * @return a non-null but possibly empty array of arguments
+   */
+  public static String[] split(String line) {    
+    char quote = '"';
```
chanseokoh: I think initializing this to a particular quote adds confusion. The value is not used and will just be overwritten, right?

elharo: It needs to be initialized to something though:

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.5.1:compile (default-compile) on project appengine-plugins-core: Compilation failure
[ERROR] /Users/elharo/appengine-plugins-core/src/main/java/com/google/cloud/tools/shell/CommandLine.java:[52,35] variable quote might not have been initialized
[ERROR] -> [Help 1]
[ERROR]

src/main/java/com/google/cloud/tools/shell/CommandLine.java
elharo: added @throws

elharo:
Ping @chanseokoh

src/test/java/com/google/cloud/tools/shell/CommandLineTest.java
```diff
@@ -0,0 +1,99 @@
+/* Copyright 2017 Google Inc.
```
chanseokoh: Ditto.

briandealwis:
The original motivation for this work was https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1139. As there is an alternative in Eclipse, is this necessary?

loosebazooka:
reopen if you still intend on doing this.

Issue #267, opened 2016-12-02
If the gcloud is on the path, the plugin should use that to find the SDK path


patflynn:
See https://github.com/GoogleCloudPlatform/app-maven-plugin/issues/118

patflynn:
I believe we were already doing something like this in intellij using 'which' or 'where'.  I assumed this was happening in the core lib but I guess it's not.

chanseokoh:
FWIW, the library searches `$PATH`. But, yeah, it doesn't actually invoke `gcloud` but only looks for `"google-cloud-sdk/bin" in `$PATH`.

https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/master/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java#L40
https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/master/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java#L74

I guess it'd be much safer to actually invoke `gcloud` rather than relying on [some Linux commands](http://unix.stackexchange.com/questions/85249/why-not-use-which-what-to-use-then).

briandealwis:
See also https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/131#issuecomment-227783960

loosebazooka:
fixed by #295

PR #295, opened 2017-01-10
Follow symlinks on gcloud to cloudsdk

loosebazooka:
What do you guys think of something like this for path resolution for the case

```
~/bin/gcloud -> ~/my-custom-software-location/google-cloud-sdk/bin/gcloud
```

This is a rough PR for #267 , not expected to submit in this form.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
@@ -103,8 +103,8 @@ public class PathResolver implements CloudSdkResolver {
     }
   }
 
-  @VisibleForTesting
   // resolve symlinks to a path that could be the bin directory of the cloud sdk
+  @VisibleForTesting
   static void getLocationsFromLink(List<String> possiblePaths, Path link) {
     try {
       Path resolvedLink = link.toRealPath();
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
@@ -16,6 +16,8 @@
 
 package com.google.cloud.tools.appengine.cloudsdk;
 
+import com.google.common.annotations.VisibleForTesting;
+
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
@@ -101,8 +103,9 @@ public class PathResolver implements CloudSdkResolver {
     }
   }
 
+  @VisibleForTesting
   // resolve symlinks to a path that could be the bin directory of the cloud sdk
-  private static void getLocationsFromLink(List<String> possiblePaths, Path link) {
+  static void getLocationsFromLink(List<String> possiblePaths, Path link) {
     try {
       Path resolvedLink = link.toRealPath();
       Path possibleBinDir = resolvedLink.getParent();
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
@@ -17,16 +17,21 @@
 package com.google.cloud.tools.appengine.cloudsdk;
 
 import java.io.File;
+import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * Resolve paths with CloudSdk and Python defaults.
  */
 public class PathResolver implements CloudSdkResolver {
+
+  private static final Logger logger = Logger.getLogger(PathResolver.class.getName());
 
   /**
    * Attempts to find the path to Google Cloud SDK.
@@ -58,6 +63,8 @@ public class PathResolver implements CloudSdkResolver {
       possiblePaths.add("/usr/local/share/google/google-cloud-sdk");
     }
 
+    Path finalPath = searchPaths(possiblePaths);
+    logger.log(Level.FINE, "Resolved SDK path : " + finalPath.toString());
     return searchPaths(possiblePaths);
   }
 
@@ -85,8 +92,32 @@ public class PathResolver implements CloudSdkResolver {
         if (path.endsWith("google-cloud-sdk" + File.separator + "bin")) {
           possiblePaths.add(path.substring(0, path.length() - 4));
         }
+
+        Path possibleLink = Paths.get(path, "gcloud");
+        if (Files.isSymbolicLink(possibleLink)) {
+          getLocationsFromLink(possiblePaths, possibleLink);
+        }
       }
     }
+  }
+
+  // resolve symlinks to a path that
+  private static void getLocationsFromLink(List<String> possiblePaths, Path link) {
+    try {
+      Path resolvedLink = link.toRealPath();
+      Path possibleBinDir = resolvedLink.getParent();
+      // check if the parent is "bin", we actually depend on that for other resolution
+      if (possibleBinDir != null && possibleBinDir.getFileName().toString().equals("bin")) {
+        Path possibleCloudSdkHome = possibleBinDir.getParent();
+        if (possibleCloudSdkHome != null && Files.exists(possibleCloudSdkHome)) {
+          possiblePaths.add(possibleCloudSdkHome.toString());
+        }
+      }
+    } catch (IOException ioe) {
+      // intentionally ignore exception
+      logger.log(Level.FINE, "Non-critical exception when searching for cloud-sdk", ioe);
+    }
+
   }
 
   private static String getProgramFilesLocation() {
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
@@ -101,7 +101,7 @@ public class PathResolver implements CloudSdkResolver {
     }
   }
 
-  // resolve symlinks to a path that
+  // resolve symlinks to a path that could be the bin directory of the cloud sdk
   private static void getLocationsFromLink(List<String> possiblePaths, Path link) {
     try {
       Path resolvedLink = link.toRealPath();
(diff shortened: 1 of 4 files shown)


loosebazooka:
@patflynn @briandealwis @chanseokoh

patflynn:
Wasn't that gcloud command the suggested solution?

https://github.com/GoogleCloudPlatform/app-maven-plugin/issues/118

loosebazooka:
Oh I didn't see that. Do we really want to be running gcloud to see where it is though?

patflynn:
I think IJ tends to use 'which/where' to solve this problem as well which
seemed effective.

loosebazooka:
even on windows?

patflynn:
Yeah. 'which' = unix, 'where'=windows.

loosebazooka:
oh right... :man_facepalming:

chanseokoh:
I think using `which`-like commands could be problematic, since it does not follow symlinks AFAIK. It is particularly problematic if, e.g., users [install gcloud using `apt-get` on Ubuntu/Debian](https://cloud.google.com/sdk/downloads#apt-get). When installed through `apt-get`, [a gcloud symlink will be created at /usr/bin/gcloud, but the actual SDK home will be at /usr/lib/google-cloud-sdk](https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/286#discussion_r94313335).

src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
```diff
@@ -85,6 +86,16 @@ private static void getLocationsFromPath(List<String> possiblePaths) {
         if (path.endsWith("google-cloud-sdk" + File.separator + "bin")) {
           possiblePaths.add(path.substring(0, path.length() - 4));
         }
+
+        Path possibleLink = Paths.get(path, "gcloud");
+        if (Files.exists(possibleLink) && Files.isSymbolicLink(possibleLink)) {
```
chanseokoh: On my Linux, the `gcloud` executable is not a symlink. (I just unzipped the SDK zip.) So I think `isSymbolicLink()` is always false for me.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
```diff
@@ -85,6 +86,16 @@ private static void getLocationsFromPath(List<String> possiblePaths) {
...
+          try {
+            Path destination = Files.readSymbolicLink(possibleLink);
```
chanseokoh: - Following a symlink seems a nice idea, but why once? Maybe we should iterate until it is not a symlink.
- But then, if the SDK is updated in the future in a way that `gcloud` becomes a symlink and the actual binary is placed in some weird sub-folder, it might break. (Maybe this is stretched too far.)
- Might be a good idea to check existence of `destination`?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
briandealwis: `File.getCanonicalPath()` resolves symlinks.

loosebazooka:
if anyone wants to take a look at this new version which follows the symlink to its original file.

loosebazooka:
@chanseokoh @briandealwis anyone interested in taking another look at this?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
```diff
@@ -85,8 +94,33 @@ private static void getLocationsFromPath(List<String> possiblePaths) {
...
+      }
+    }
+  }
+
+  @VisibleForTesting
+  // resolve symlinks to a path that could be the bin directory of the cloud sdk
+  static void getLocationsFromLink(List<String> possiblePaths, Path link) {
+    try {
+      Path resolvedLink = link.toRealPath();
+      Path possibleBinDir = resolvedLink.getParent();
```
chanseokoh: It may be possible that `getParent()` doesn't work as you expected: http://stackoverflow.com/a/1099370/1701388. Probably it's better to do `getCanonicalPath()` before it.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
```diff
@@ -85,8 +94,33 @@ private static void getLocationsFromPath(List<String> possiblePaths) {
...
+        }
+      }
+    }
+  }
+
+  @VisibleForTesting
+  // resolve symlinks to a path that could be the bin directory of the cloud sdk
+  static void getLocationsFromLink(List<String> possiblePaths, Path link) {
+    try {
+      Path resolvedLink = link.toRealPath();
```
chanseokoh: Theoretically it's possible that `resolvedLink` is again a symlink?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
```diff
@@ -85,8 +94,33 @@ private static void getLocationsFromPath(List<String> possiblePaths) {
...
+        Path possibleLink = Paths.get(path, "gcloud");
+        if (Files.isSymbolicLink(possibleLink)) {
+          getLocationsFromLink(possiblePaths, possibleLink);
+        }
+      }
+    }
+  }
+
+  @VisibleForTesting
+  // resolve symlinks to a path that could be the bin directory of the cloud sdk
```
chanseokoh: I guess the comment should go above `@VisibleFotTesting`.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
chanseokoh: But then, the last portion of `resolvedLink` is always `gcloud`, so I guess this is OK.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
briandealwis: The javadocs return result says:
> _an absolute path represent the <em>real</em> path of the file located by this object_

It suggests that symlinks should be fully resolved.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
briandealwis: Javadocs suggest that `..` should be collapsed where its supported by the underlying file system.

chanseokoh: But my test below gives me `/a/b`, not `/` on Linux.

```java
        String pathStr = "/a/b/../";
        Path path = Paths.get(pathStr);
        System.out.println(path.getParent());
```

chanseokoh: > But then, the last portion of resolvedLink is always gcloud, so I guess this is OK.

Actually, it could matter if the path was like `/home/google-cloud-sdk/bin/./gcloud`. `possibleBinDir.getFileName()` will get you `.` instead of `bin`.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
chanseokoh: Oh, yeah, I verified it fully resolves a symlink, at least on Linux.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
loosebazooka: @chanseokoh, I think the fact that we call toRealPath() first takes care of the issue.

Path.resolve("/home/usr/google-cloud-sdk/bin/./gcloud").toRealPath().getFileName() = "bin"

chanseokoh: Oh, I see. Seems like `toRealPath()` returns a canonical path.

briandealwis: Sorry, I should have been clear that I was referring to `toRealPath()`.

src/test/java/com/google/cloud/tools/appengine/cloudsdk/PathResolverTest.java
```diff
@@ -31,6 +46,51 @@ public void testResolve() {
...
-  }  
-  
+  }
+
+  @Test
+  public void testGetLocationFromLink_valid() throws IOException {
+    Path sdkHome = temporaryFolder.newFolder().toPath();
+    Path bin = Files.createDirectory(sdkHome.resolve("bin"));
+    Path gcloud =  Files.createFile(bin.resolve("gcloud"));
+    Files.createSymbolicLink(temporaryFolder.getRoot().toPath().resolve("gcloud"), gcloud);
```
loosebazooka: So these tests seem to work if you run under Administrator in windows server 2016. I've updated the test assumptions on these symlink test classes.

From windows testing, PathResolver will skip 3 tests if windows(non-admin), run all tests in admin mode.

src/test/java/com/google/cloud/tools/appengine/cloudsdk/PathResolverTest.java
```diff
@@ -31,6 +46,51 @@ public void testResolve() {
...
+    PathResolver.getLocationsFromLink(possiblePaths, gcloud);
+
+    Assert.assertEquals(1, possiblePaths.size());
+    Assert.assertEquals(gcloud.getParent().getParent().toString(), possiblePaths.get(0));
+  }
+
+  @Test
+  public void testGetLocationFromLink_notValid() throws IOException {
+    Path invalidPath = temporaryFolder.newFolder().toPath();
+    Files.createSymbolicLink(temporaryFolder.getRoot().toPath().resolve("gcloud"), invalidPath);
```
loosebazooka: updated.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
loosebazooka: done

loosebazooka:
@chanseokoh @briandealwis made some updates to testing. PTAL, I know you've already approved, but just in case.

PR #306, opened 2017-01-22
Single exception for version file problems

elharo:
@aslo @meltsufin Looking at this code for other reasons I realized that CT4E wasn't handling two of three possible failure modes because every mode has a different, unrelated, runtime exception type. The safest solution is to use  checked exceptions which would have turned our problems into compile time errors that were easily fixed. However as a stop gap use a single subclass of AppEngineException instead.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkVersionFileNotFoundException.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkVersionFileNotFoundException.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2016 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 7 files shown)


patflynn:
@elharo @aslo I think IJ would have issues with this change.  I'm all for using checked exceptions here for things that we want to react to, but we need to be able to differentiate between error causes without parsing error messages.

elharo:
In that case we should use different exception subclasses, not arbitrary runtime exceptions. The existing code already caused user facing problems in Eclipse. I.e. we received an external bug report from a non-Google user that was hard to classify and debug because of this code that didn't report a missing VERSION file.

patflynn:
@elharo I agree, but stuff like CloudSdk#validateCloudSdkVersion() needs to still work with this change.

aslo:
Can we not just change to using checked exceptions while we're in the neighborhood?

elharo:
I'd like to change AppEngineException to be a checked exception. That's going to be a big change that should probably be in its own PR though, and I've been holding off until we have agreement that's the way to go.

elharo:
Ping. I'd like to get these changes in and push out a new release of appengine-plugins-core with this week's CT4E beta if possible.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkVersionFileNotFoundException.java
```diff
@@ -16,22 +16,10 @@
 
 package com.google.cloud.tools.appengine.cloudsdk;
 
-import com.google.cloud.tools.appengine.api.AppEngineException;
-
-/**
- * The Cloud SDK's version file could not be found where expected.
- */
-public class CloudSdkVersionFileNotFoundException extends AppEngineException {
+public class CloudSdkVersionFileNotFoundException extends CloudSdkVersionFileException {
```
patflynn: you're missing javadoc here.

PR #307, opened 2017-01-22
Add AppEngineDescriptor class

elharo:
@chanseokoh Copied from CT4E

Also suppresses outdated import order rules

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
+++ b/src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
@@ -85,8 +85,7 @@ public class AppEngineDescriptor {
     return getTopLevelValue(document, "appengine-web-app", "module");
   }
 
-  private static String getTopLevelValue(Document doc, String parentTagName, String childTagName)
-      throws AppEngineException {
+  private static String getTopLevelValue(Document doc, String parentTagName, String childTagName) {
     try {
       NodeList parentElements = doc.getElementsByTagNameNS(APP_ENGINE_NAMESPACE, parentTagName);
       if (parentElements.getLength() > 0) {
--- a/src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
+++ b/src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
@@ -60,7 +60,6 @@ public class AppEngineDescriptor {
   /**
    * @return project ID from the &lt;application&gt; element of the appengine-web.xml or null
    *         if it is missing
-   * @throws AppEngineException if parsing the value fails
    */
   public String getProjectId()  {
     return getTopLevelValue(document, "appengine-web-app", "application");
@@ -69,7 +68,6 @@ public class AppEngineDescriptor {
   /**
    * @return project version from the &lt;version&gt; element of the appengine-web.xml or
    *         null if it is missing
-   * @throws AppEngineException if parsing the value fails
    */
   public String getProjectVersion() {
     return getTopLevelValue(document, "appengine-web-app", "version");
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
```diff
@@ -0,0 +1,112 @@
...
+import org.w3c.dom.NodeList;
+import org.xml.sax.SAXException;
+
+/**
+ * Utilities to obtain information from appengine-web.xml.
+ */
+public class AppEngineDescriptor {
+
+  private static final String WEB_XML_NS_URI = "http://appengine.google.com/ns/1.0";
+  private Document document;
```
chanseokoh: How about initializing this in the constructor and make it `final`? Seems doable.

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
```diff
@@ -0,0 +1,112 @@
...
+   *         null if it is missing
+   * @throws AppEngineException if parsing the value fails
+   */
+  public String getProjectVersion() {
+    return getTopLevelValue(document, "appengine-web-app", "version");
+  }
+
+  /**
+   * @return service ID from the &lt;service&gt; element of the appengine-web.xml, or
+   *         null if it is missing. Will also look at module ID.
```
chanseokoh: Like the other methods that use `getTopLevelValue()`, add `@throws AppEngineException if parsing the value fails` to the javadoc?

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
elharo: done

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
```diff
@@ -0,0 +1,111 @@
...
+            if (child.getNodeName().equals(childTagName)) {
+              return child.getTextContent();
+            }
+          }
+        }
+      }
+      return null;
+    } catch (DOMException ex) {
+      // this shouldn't happen barring a very funky DOM implementation
+      throw new AppEngineException(ex);
```
patflynn: throw new RuntimException(ex)?

patflynn:
@elharo I don't see any use-cases for IJ or Maven/Gradle for this class at the moment.

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
elharo: Since this is a "can't happen" case in an already Nullable method, I decided to go with returning null here instead of throwing an exception.

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
elharo: decided to just return null instead

elharo:
Ping

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
```diff
@@ -0,0 +1,109 @@
...
+  public String getServiceId() {
+    String serviceId = getTopLevelValue(document, "appengine-web-app", "service");
+    if (serviceId != null) {
+      return serviceId;
+    }
+    return getTopLevelValue(document, "appengine-web-app", "module");
+  }
+
+  private static String getTopLevelValue(Document doc, String parentTagName, String childTagName)
+      throws AppEngineException {
```
chanseokoh: Now we don't throw this, right?

elharo: right, removed

PR #310, opened 2017-01-24
Add clearDatastore flag

loosebazooka:
Also fix some minor checkstyle issue

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
@@ -47,6 +47,7 @@ public class DefaultRunConfiguration implements RunConfiguration {
   private Boolean skipSdkUpdateCheck;
   private String defaultGcsBucketName;
   private String javaHomeDir;
+  private Boolean clearDatastore;
 
   @Override
   public List<File> getAppYamls() {
@@ -253,5 +254,14 @@ public class DefaultRunConfiguration implements RunConfiguration {
 
   public void setJavaHomeDir(String javaHomeDir) {
     this.javaHomeDir = javaHomeDir;
+  }
+
+  @Override
+  public Boolean getClearDatastore() {
+    return clearDatastore;
+  }
+
+  public void setClearDatastore(Boolean clearDatastore) {
+    this.clearDatastore = clearDatastore;
   }
 }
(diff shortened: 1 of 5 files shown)


loosebazooka:
Fixes #303

src/main/java/com/google/cloud/tools/appengine/api/devserver/RunConfiguration.java
```diff
@@ -72,4 +72,6 @@
   String getDefaultGcsBucketName();
 
   String getJavaHomeDir();
+
+  Boolean getClearDatastore();
```
elharo: 1. Needs Javadoc.
2. Effective Java recommends using boolean primitive

src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
```diff
@@ -254,4 +255,13 @@ public String getJavaHomeDir() {
   public void setJavaHomeDir(String javaHomeDir) {
     this.javaHomeDir = javaHomeDir;
   }
+
+  @Override
+  public Boolean getClearDatastore() {
+    return clearDatastore;
+  }
+
+  public void setClearDatastore(Boolean clearDatastore) {
```
elharo: needs Javadoc

src/main/java/com/google/cloud/tools/appengine/api/devserver/RunConfiguration.java
loosebazooka: We want default behavior to be null, so we always delegate to gcloud defaults (which can be true or false)

patflynn:
@loosebazooka 
Are you thinking of introducing the configuration property independently of plumbing it into the run command?

loosebazooka:
no, looks like I forgot to add a file in my rush to eat lunch... :open_mouth:

src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
loosebazooka: We haven't been documenting value class getters/setters like this in this project. Do you think we should?

elharo: Yes. This needs multiple things clearly stated:

1. What this flag does. i.e. mostly copy the gcloud docs on this
2. What the default is and what null means.

PR #314, opened 2017-01-28
copy DeleteAllVisitor to appengine-plugins-core and add tests

elharo:
@loosebazooka

### DIFF
--- a/src/main/java/com/google/cloud/tools/io/DeleteAllVisitor.java
+++ b/src/main/java/com/google/cloud/tools/io/DeleteAllVisitor.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2016 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


elharo:
Ping

elharo:
Ping ping ping

loosebazooka:
where is this used?

elharo:
In Cloud Tools for Eclipse

loosebazooka:
I wonder if we should be extending the appengine-plugin library to implement utilities that are not appengine related. Obviously this is a philosophical issue about the purpose of the library, and there is likely to be a difference in how people see it, but personally I'd rather not see generic file operations implemented in the library for use outside of it.

For example in Java9 I would like FileUtil to be not exported out of the module.

patflynn:
+1. I agree with this mainly because the different target client platforms
diverge signicantly (maven vs gradle vs intellij vs eclipse) in terms of
Java language level, available libraries, and dependency requirements
(things like IJ standardizing on a version of Guava).  The core app-engine
is so important across those plugins that I want to keep it's scope as
narrow as possible.

PR #320, opened 2017-02-06
Remove findbugs, add errorprone

loosebazooka:


patflynn:
@loosebazooka Would this be a problem for Eclipse.Maven/Gradle that are on Java 7?

loosebazooka:
@patflynn the target and source are set to java 7, so this should be okay.

PR #322, opened 2017-02-07
Make cobertuna run verify instead of just test

loosebazooka:
I think we had this misconfigured. Currently travis is only running `test` from the `verify` stages. This enables all `verify` steps.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,11 +25,14 @@ script:
     else
       mvn clean install -B -U
     fi
-    mvn cobertura:cobertura verify
+  - mvn cobertura:cobertura
 
 env:
   global:
     - secure: "da45ryIB/m37Gkpon/Uy6PPl1AhsUMdxgG7aeQIM1alh3Np4JwqKHobdP8tC0Qc/vb5fqfisEZEGPukh9Kxw95pRATO/QrQrsxYISMXUUc+Dlvo8WLeDq5F2LdUSn3933H9p5Mk8bKrZql+Jb+Il5S+B3Ib8uO+e3L4itrGKu5dw6i8TAxMggUjK8L6RuRYZeXMNiw3iaLjlHhNVEZ7F7RQ/gsHT2LhzybY6gfVJU+8AHhwEv+Tuapz5QCYTah6pwKqP+EQPhFlfrow9zdBfQm7m4h+uU+TB67VzZi46pAx+drC1quW+HZllutMHKM+cR13HsTET9qbFmV00cr2gZ3UXMVPX+PG1johZj0gs8s/S0+MVh6IQ0QLMWSqgJH/ULlyoU9PLE+PMs8A2qprO8NALuS3GXgvMQ8tHGuAzUVht/CxH5WTxW8nGFv8lwCIrT+m0hWi26gXWpNItC4N5GawoJQp+eWW0dO2Uko34kC3CIkLppRGxgzQWQQHkR+Hh+Yi3AzZsUzz30YRezF+G1RuQ68Va9yZ9qgei3h0Ppx2OIZ6fyiGg6Z8MHzOA8AULoe/Sz7CsfCiOyjTWmccRSl79wc0kpZwrF2qLO46LD3DgpGcfcKEB/cqcdwj/SA4QJoRcwUPSqy2eXb6ILhHA9UGlSBwluPZ2tSXDh2zxCL0="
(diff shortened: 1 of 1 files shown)


loosebazooka:
um, this isn't right... hang on

.travis.yml
```diff
@@ -25,7 +25,7 @@ script:
     else
       mvn clean install -B -U
     fi
-    mvn cobertura:cobertura
+    mvn cobertura:cobertuna verify
```
elharo: cobertura, not cobertuna

loosebazooka: right, oops

elharo:
I'm unclear on what this is doing. What does this run it didn't run before?

loosebazooka:
It didn't run a verify (which is where checkstyle is registered to). See http://www.mojohaus.org/cobertura-maven-plugin/cobertura-mojo.html

elharo:
so the effect is that it isn't running checkstyle? Honestly, I'm OK with that. The external checkstyle rules are not very up to date, and seem to cause as many problems as they solve.

loosebazooka:
or anything in the verify block.

loosebazooka:
if we don't want to run checkstyle, we should remove it.

loosebazooka:
After looking at this some more it looks like verify was running but the cobertura result is what determines the pass/fail result of the build which seems bad. See https://docs.travis-ci.com/user/customizing-the-build

I'll update it so cobertura runs as a second check.

.travis.yml
```diff
@@ -25,11 +25,14 @@ script:
...
     - secure: "Z/Haq0MHdfBpJKntdYZFdK3uxUUk+jjiNzuuOt2hLa7P5XxpoHsboRykWSY41tWD5NnXfXoJvcHPASwhz2OesQcyNCCEHxsikkNRk+XUgFjyu8I3Ohm8Fya1k9gn+Xlz9uDBtoGN1e2SHhfEY+a9a5nGvv4aLVblBewt9J14su8fc4WSyDYggoZnUjB69DzIJrBL/ol0VOqlo8/vrcbWRbB/hlJ3QwKsFkehIhx5//GVdEJGqM2CZNJqG/bVLaSIoqQPqd6v3eyemnS2O5bXNIoNEq9yUWGSm1PfnagBYMzkHcqNrg5D1aYI+vFFeZe4PAsBaE6Xw/trCMtdRr5uDKXdNW8s8zwMKfOYru3zY9KTq2N5Fl/HgazGfimpmAjXqpH74+WaITmQmCZ6LuLb7hgltF+f2OY71bHJI7lGeuCraiwb6ECFIZiF9+FA6m0DXU6fCsajoThUtkRSpTMxDTAMzlrH/Kw2SzwEfuGk3evsb937pLUkL1kNxSc6GXhdyRrLTSrYgoqavjtNUN5S7v7MqrmO4R6UF1ZtJOAxWJ1a4W/kQP6S/36u8dZhCWbds78EocOH/+fsJ8vlCmTAnFEbtZkSpmFHjGrpPu0gVYdG+u96eAtKv90p2IF8bwCzFQRh1U25OlcxHKZfDgGzplR1Cy8DdJoNcH3KJKQd1Cw="
+
 after_success:
+  - unset SONATYPE_USERNAME
```
elharo: I don't see how this determines success/failure

loosebazooka: this is to remove the environment variable before letting the codecov.io script run on travis.

Success failure is in the "script" block, as described in this section : https://docs.travis-ci.com/user/customizing-the-build#Customizing-the-Build-Step

elharo: Sounds reasonable. But it looks like there's no other significant change in this PR' i.e. it's leaving verification and success-failure as they were previously. Or am I missing something?

loosebazooka: commented on the changed line

.travis.yml
```diff
@@ -25,11 +25,14 @@ script:
     else
       mvn clean install -B -U
     fi
-    mvn cobertura:cobertura
+  - mvn cobertura:cobertura
```
loosebazooka: @elharo see the addition of a "-" here.

elharo: I guess the previous minus sign is just a git diff marker?

loosebazooka: haha yeah.

.travis.yml
loosebazooka: ack

PR #327, opened 2017-02-10
Allowing access to the runtime configuration.

ludoch:


src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
```diff
@@ -65,6 +65,13 @@ public String getProjectId()  {
   }
 
   /**
+   * @return runtime from the &lt;application&gt; element of the appengine-web.xml or null
```
elharo: runtime element; application element contains the project ID

ludoch:
PTAL

PR #332, opened 2017-02-16
Fix command-line generation for devappserver1

briandealwis:
A few bug fixes uncovered when trying to get the devappserver1 to work with CT4E.
  - jvm flags need to be passed directly to the invoked JVM
  - `--allow_remote_shutdown` isn't supported in Cloud SDK 144.0
  - `--server` is to specify an update server; use `--address` and `--port`

PTAL @ludoch

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
@@ -27,7 +27,6 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.collect.Maps;
 import com.google.common.io.ByteStreams;
-
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
@@ -87,11 +86,13 @@ public class CloudSdkAppEngineDevServer1 implements AppEngineDevServer {
               .getAbsolutePath();
       jvmArguments.add("-javaagent:" + appengineAgentJar);
     } 
-    arguments.addAll(DevAppServerArgs.get("server", config.getHost()));
-    arguments.addAll(DevAppServerArgs.get("address", config.getPort()));
-    arguments.addAll(DevAppServerArgs.get("jvm_flag", config.getJvmFlags()));
+    arguments.addAll(DevAppServerArgs.get("address", config.getHost()));
+    arguments.addAll(DevAppServerArgs.get("port", config.getPort()));
+    if (config.getJvmFlags() != null) {
+      jvmArguments.addAll(config.getJvmFlags());
+    }
 
-    arguments.add("--allow_remote_shutdown");
+    // arguments.add("--allow_remote_shutdown");
     arguments.add("--disable_update_check");
     if (appengineWeb.isJava8()) {
       arguments.add("--no_java_agent");
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
@@ -92,7 +92,7 @@ public class CloudSdkAppEngineDevServer1 implements AppEngineDevServer {
       jvmArguments.addAll(config.getJvmFlags());
     }
 
-    // arguments.add("--allow_remote_shutdown");
+    arguments.add("--allow_remote_shutdown");
     arguments.add("--disable_update_check");
     if (appengineWeb.isJava8()) {
       arguments.add("--no_java_agent");
--- a/src/test/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1Test.java
+++ b/src/test/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1Test.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright 2016 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.tools.appengine.cloudsdk;
+
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+import com.google.cloud.tools.appengine.api.AppEngineException;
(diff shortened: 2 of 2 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -87,11 +86,13 @@ public void run(RunConfiguration config) throws AppEngineException {
...
-    arguments.addAll(DevAppServerArgs.get("address", config.getPort()));
-    arguments.addAll(DevAppServerArgs.get("jvm_flag", config.getJvmFlags()));
+    arguments.addAll(DevAppServerArgs.get("address", config.getHost()));
+    arguments.addAll(DevAppServerArgs.get("port", config.getPort()));
+    if (config.getJvmFlags() != null) {
+      jvmArguments.addAll(config.getJvmFlags());
+    }
 
-    arguments.add("--allow_remote_shutdown");
+    // arguments.add("--allow_remote_shutdown");
```
elharo: please delete if we don't need this

loosebazooka: From what I remember, this has always been part of the kickstart mechansim. It might still be needed?

briandealwis: It's not listed by `DevAppServerMain` as a supported argument, and its presence causes failure:
```sh
$ java -cp $GOOGLE_CLOUD_SDK_HOME/platform/google_appengine/google/appengine/tools/java/lib/appengine-tools-api.jar com.google.appengine.tools.development.DevAppServerMain
Usage: <dev-appserver> [options] <app directory>

Options:
 --help, -h                 Show this help message and exit.
 --sdk_root=DIR             Overrides where the SDK is located.
 --runtime=runtime_id       (java7 or java8).
 --server=SERVER            The server to use to determine the latest
  -s SERVER                   SDK version.
 --address=ADDRESS          The address of the interface on the local machine
  -a ADDRESS                  to bind to (or 0.0.0.0 for all interfaces).
 --port=PORT                The port number to bind to on the local machine.
  -p PORT
 --disable_update_check     Disable the check for newer SDK versions.
 --generated_dir=DIR        Set the directory where generated files are created.
 --default_gcs_bucket=NAME  Set the default Google Cloud Storage bucket name.
 --jvm_flag=FLAG            Pass FLAG as a JVM argument. May be repeated to
                              supply multiple flags.
```
The `—jvm_flag` argument isn't supported either.

loosebazooka: right, jvm_flag is a dev_appserver2 thing. But we have access directly to the jvm here, so that's okay.

Does stopping the server remotely still work even without allow_remote_shutdown?

loosebazooka: @ludoch do you know what's going on here?

ludoch: allow_remote_shutdown is needed...

ludoch: Link to failures with --allow_remote_shutdown?

ludoch:
Only the Maven plugin has some e2e tests involving async start and stop of local servers, and allow_remote_shutdown is needed...

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
loosebazooka: I believe it's a hidden flag.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -87,11 +86,13 @@ public void run(RunConfiguration config) throws AppEngineException {
               .getAbsolutePath();
       jvmArguments.add("-javaagent:" + appengineAgentJar);
     } 
-    arguments.addAll(DevAppServerArgs.get("server", config.getHost()));
-    arguments.addAll(DevAppServerArgs.get("address", config.getPort()));
-    arguments.addAll(DevAppServerArgs.get("jvm_flag", config.getJvmFlags()));
```
loosebazooka: I think this line is okay, DevAppServerArgs should do the right thing.

loosebazooka: for jvm_flag

briandealwis: It definitely doesn't.  The easy way to test is to run:
```
$ java -cp $GOOGLE_CLOUD_SDK_HOME/platform/google_appengine/google/appengine/tools/java/lib/appengine-tools-api.jar com.google.appengine.tools.development.DevAppServerMain [flags] /tmp
```
For flags that aren't accepted, the command will show the help text.  For flags that are accepted, the command will run and spit out some error messages of some kind:
```
$ java -cp $GOOGLE_CLOUD_SDK_HOME/platform/google_appengine/google/appengine/tools/java/lib/appengine-tools-api.jar com.google.appengine.tools.development.DevAppServerMain /tmp
********************************************************
There is a new version of the SDK available.
-----------
Latest SDK:
Release: 1.9.49
Timestamp: Wed Jan 18 12:33:47 EST 2017
API versions: [1.0]
[…]
```

Running with `-jvm_flag=foo` results in the help text:
```
$ java -cp $GOOGLE_CLOUD_SDK_HOME/platform/google_appengine/google/appengine/tools/java/lib/appengine-tools-api.jar com.google.appengine.tools.development.DevAppServerMain -jvm_flag=foo /tmp
Usage: <dev-appserver> [options] <app directory>

Options:
 —help, -h                 Show this help message and exit.
[…]
```

briandealwis: Just to be clear, the actual `—jvm_arg=XXX` quoting is fine, but devappserver1 itself doesn't understand them.

briandealwis:
> @elharo: You might be able to test this with a mock CloudSdk.

Indeed there are some tests but they aren't being executed as it's called `CloudSdkAppEngineDevServerTest1`.  I've renamed the test and fixed it up.

ludoch:
Brian: thanks for the update which looks good for me.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
loosebazooka: I'm still not convinced this is true, running the script from the command line 
```
 ./dev_appserver.sh --jvm_flag=-Xdebug --
jvm_flag=-Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=5005 --allow_remote_shutdown /path/to/some/app
```
will work and start the server in debug.

I will try from the library and see though.

loosebazooka: Oh, KickStart is handling the jvm_flags... never mind.

PR #339, opened 2017-03-04
add test and remove unused code

elharo:
@loosebazooka 

git's diff algorithm appears to be a bit confused by this.
The deleted file and the new file are not related.


src/main/java/com/google/cloud/tools/appengine/api/services/DefaultServicesSelectionConfiguration.java
```diff
@@ -1,40 +0,0 @@
...
-package com.google.cloud.tools.appengine.api.services;
-
-import com.google.cloud.tools.appengine.api.DefaultConfiguration;
-
-import java.util.Collection;
-
-/**
- * Plain Java bean implementation of {@link ServicesSelectionConfiguration}.
- */
-public class DefaultServicesSelectionConfiguration
```
joaoandremartins: Should we split this PR in two? I'm not convinced this code needs to be deleted. I think it should probably stay, in case we're ever interested in getting every service for an app (which is what I think this config helps doing?)

elharo: https://martinfowler.com/bliki/Yagni.html

PR #340, opened 2017-03-05
add BUILD status badge

elharo:
@joaoandremartins

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,9 @@
 [![experimental](http://badges.github.io/stability-badges/dist/experimental.svg)](http://github.com/badges/stability-badges)
+[![Build Status](https://travis-ci.org/GoogleCloudPlatform/appengine-plugins-core.svg?branch=master)](https://travis-ci.org/GoogleCloudPlatform/appengine-plugins-core)
+
 # Google App Engine Plugins Core Library
 
-This is the client Java library to manage App Engine Java applications. It should be used by any application that performs App Engine Java application management. For example, the Maven, Gradle and Eclipse App Engine plugins, custom user tools, etc.
+This is the client Java library to manage App Engine Java applications. It should be used by any Java application that performs App Engine Java application management. For example, the Maven, Gradle and Eclipse App Engine plugins, custom user tools, etc.
 
 # Requirements
 


README.md
```diff
@@ -1,7 +1,9 @@
 [![experimental](http://badges.github.io/stability-badges/dist/experimental.svg)](http://github.com/badges/stability-badges)
+[![Build Status](https://travis-ci.org/GoogleCloudPlatform/appengine-plugins-core.svg?branch=master)](https://travis-ci.org/GoogleCloudPlatform/appengine-plugins-core)
+
 # Google App Engine Plugins Core Library
 
-This is the client Java library to manage App Engine Java applications. It should be used by any application that performs App Engine Java application management. For example, the Maven, Gradle and Eclipse App Engine plugins, custom user tools, etc.
+This is the client Java library to manage App Engine Java applications. It should be used by any Java application that performs App Engine Java application management. For example, the Maven, Gradle and Eclipse App Engine plugins, custom user tools, etc.
```
patflynn: This wording slipped by me originally. 

This library should not be used by anybody but our plugins. We make no commitment to its API, featureset, or future existence.

joaoandremartins: I'm not so sure that was the original plan. I don't see how using this API externally is mutually exclusive with a stable API. Users can just use it at their own risk. If, at any point, there are any breaking changes, they can always resort to an older stable version.

Not that I think the plugin will see much external usage...

patflynn: If external users choose to use this library we certainly can't stop them.  But we want to disabuse them of the notion that this is an official supported Google admin API for App Engine.  The naming of the repo was designed to do that, but this wording isn't.  We just need to fix it.

rdayal: I think you need to be clear about what the purpose of this library is.

If your plan is to support feature requests and breaks reported by other users that are using this library for their own purposes, then it's public. If that's not the plan, then it's private, and you should not make/imply any support guarantees.

I'd recommend leaving this as private. Unless it is clear how/why this is valuable for other users and unless it's requested, I'd opt not to expose an API, because then you have to support it.

joaoandremartins: I've always thought of this as the Cloud SDK-based App Engine Java SDK. But that direction might have changed while I was out. You could argue that there is already the App Engine Admin API Java client lib which does the same things that this one does. Since we're also not planning to implement staging and all other goodies that gcloud does here, it's probably OK to reword.

elharo: Right now we've marked this repo as stability:experimental. I think that sends the correct message to anyone who depends on it. We version it in Maven so it's easy for any client to   upgrade with breaking changes on their own schedule. CT4E depends on this.

rdayal: What versioning are you using in Maven?

patflynn:
@elharo yes I'm glad we have the experimental badge on it. But we should be clear that this is not 'the' general purpose App Engine admin API and external users are discouraged from depending on it.

patflynn:
I'll send a PR. It doesn't have to block this change.

patflynn:
nm. looks like this is merged already anyway.

PR #342, opened 2017-03-06
External users are not supported

patflynn:
update our README to inform external users that this is not a general purpose App Engine admin library.


rdayal:
LGTM.

joaoandremartins:
It could be interesting to provide this kind of API to external users in the future.

patflynn:
Ah. Totally forgot to merge this.

PR #343, opened 2017-03-16
Add param verifier to tests

loosebazooka:
- This helps us ensure we're using all the params we create at
  testing time or we handle them with warnings
- Update Mockito to a modern version
- Remove DevServerTest1 which already exists as DevServer1Test
  and wasn't getting run anyway
- Remove stubs that Mockito complained about


src/test/java/com/google/cloud/tools/test/utils/SpyVerifier.java
```diff
@@ -0,0 +1,64 @@
...
+
+  public void verifyDeclaredGetters() throws Exception {
+    verifyDeclaredGetters(Collections.<String, Integer>emptyMap());
+  }
+
+  /**
+   * Verify getters were called once or if in the exception map, called 'count' times
+   * @param exceptions Exceptions of the form ["getter name" : count]
+   */
+  public void verifyDeclaredGetters(Map<String, Integer> exceptions) throws Exception {
```
akerekes: Is this throwing `Exception` to avoid declaring 4 different exceptions?

loosebazooka: It's also just at the test level, but yeah. If the test fails due to an exception, then it's something we should fix, not catch.

loosebazooka: What I mean is that the exception here is not of any significant value.

joaoandremartins: Why did you call the parameter "exceptions"? Isn't it the amount of times that a getter gets called?

src/test/java/com/google/cloud/tools/test/utils/SpyVerifier.java
```diff
@@ -0,0 +1,64 @@
...
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.Collections;
+import java.util.Map;
+
+public class SpyVerifier {
+  private final Object objectToInspect;
+  private final Class classToInspectAs;
+
+  public static SpyVerifier newVerifier(Object objectUnderInspection) {
```
joaoandremartins: Just curious, any specific reason you couldn't just have gone for a simple constructor instead of the factory method?

src/test/java/com/google/cloud/tools/test/utils/SpyVerifier.java
loosebazooka: exceptions to the 1 count rule. Yeah this probably could have been named better.

src/test/java/com/google/cloud/tools/test/utils/SpyVerifier.java
loosebazooka: Not really, it just a style thing (kinda guava like)

PR #344, opened 2017-03-16
Add deployment options for individual yaml files

loosebazooka:


src/main/java/com/google/cloud/tools/appengine/api/deploy/AppEngineDeployment.java
```diff
@@ -31,4 +31,28 @@
    */
   void deploy(DeployConfiguration config) throws AppEngineException;
 
+  /**
+   * Deploy cron configuration to App Engine.
+   */
+  void deployCron(DeployCronConfiguration config) throws AppEngineException;
+
+  /**
+   * Deploy dos configuration to App Engine.
```
elharo: dos? disk operating system? :-)

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDeployment.java
```diff
@@ -93,4 +98,99 @@ public void deploy(DeployConfiguration config) throws AppEngineException {
...
+    } catch (ProcessRunnerException e) {
+      throw new AppEngineException(e);
+    }
+  }
+
+  @Override
+  public void deployQueue(DeployQueueConfiguration config) throws AppEngineException {
+    Preconditions.checkNotNull(config);
+    Preconditions.checkNotNull(config.getQueueYaml());
+    Preconditions.checkArgument(config.getQueueYaml().getName().equals("queue.yaml"),
```
chanseokoh: Are we forcing users to use a fixed name for these files? Is this what the way it works?

loosebazooka: yeah, to deply a cron.yaml and update a project's cron config

```
gcloud app deploy cron.yaml
```

The only name you can change is `app.yaml`

loosebazooka:
@meltsufin yeah this feels a little redundant, but I believe the method you suggest would be less strict in how we enforce naming. The explicit definition could potentially lead to fewer user errors. Unless I misunderstand what you are suggesting.

Also when defining the configuration in plugins, I suspect we don't want to reuse configuration names. If we going to do cronYaml anyway (in gradle or maven for example) and map it to a generic yaml name when configuring, we're moving that work out of the library and into the client.

meltsufin:
@loosebazooka I see what you're saying. In that case, maybe create a `DeployYamlConfiguration` and then subclass it? This way you can still just have one method in `CloudSdkAppEngineDeployment` called `deployYaml`?

loosebazooka:
@meltsufin I want to make the names explicit for two reasons

1. so that in the maven/gradle config they are explicit
```
deploy {
  cronYaml = "path/to/cron.yaml"
}
```
and the handler in the client (build tool) just passes that on directly.

2. we're doing checks on the naming to see if they are deploying a "cron.yaml", for instance if they did
deploy {
  cronYaml = "path/to/app.yaml"
}
and we would explicitly fail here in the preconditions.

loosebazooka:
@meltsufin moved some code around for better re-use, also rebased against master. ptal

meltsufin:
@loosebazooka looks good - definitely a reduction in code duplication.

loosebazooka:
I might take another look here, do it more like Mike suggested as a generic method

loosebazooka:
@meltsufin I completely rewrote this after having a hard time convincing myself to submit it. PTAL.

src/main/java/com/google/cloud/tools/appengine/api/deploy/DeployProjectConfigurationConfiguration.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package com.google.cloud.tools.appengine.api.deploy;
+
+import com.google.cloud.tools.appengine.api.Configuration;
+
+import java.io.File;
+
+/**
+ * Created by appu on 3/21/17.
```
meltsufin: You need this?

src/main/java/com/google/cloud/tools/appengine/api/deploy/DefaultDeployProjectConfigurationConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.appengine.api.deploy;
+
+import com.google.cloud.tools.appengine.api.DefaultConfiguration;
+
+import java.io.File;
+
+public class DefaultDeployProjectConfigurationConfiguration extends DefaultConfiguration implements
```
meltsufin: `ConfigurationConfiguration`.. really? :)

src/main/java/com/google/cloud/tools/appengine/api/deploy/DefaultDeployProjectConfigurationConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+    DeployProjectConfigurationConfiguration {
+
+  private File projectConfigurationDirectory;
+
+  @Override
+  public File getProjectConfigurationDirectory() {
```
meltsufin: maybe just use `appEngineDirectory`?

src/main/java/com/google/cloud/tools/appengine/api/deploy/DefaultDeployProjectConfigurationConfiguration.java
loosebazooka: Sometimes you gotta do what you gotta do.

src/main/java/com/google/cloud/tools/appengine/api/deploy/DeployProjectConfigurationConfiguration.java
loosebazooka: done

src/main/java/com/google/cloud/tools/appengine/api/deploy/DefaultDeployProjectConfigurationConfiguration.java
loosebazooka: yeah, that works

PR #345, opened 2017-03-17
More mockito helpers to ensure betters tests

loosebazooka:


src/test/java/com/google/cloud/tools/test/utils/SpyVerifier.java
```diff
@@ -41,24 +45,68 @@ private SpyVerifier(Object objectUnderInspection) {
...
+   * This method is to verify that all setters in a configuration were called. The motivation is
+   * to ensure that configurations are fully built and we can ensure our handling is being
+   * properly tested.
+   */
+  public SpyVerifier verifyDeclaredSetters() throws Exception {
+    // extract all invocations of getters by inspecting the spy
+    List<Method> knownSetters = Arrays.asList(classToInspectAs.getDeclaredMethods());
+
+    Map<Method, Integer> methodInvocationCount = new HashMap<>();
+    for (Invocation invocation : Mockito.mockingDetails(objectToInspect).getInvocations()) {
```
joaoandremartins: Eventually, we should enumerate the Java 8 blockers for the common library, so we can start using streams.

loosebazooka: Yeah, life would be so nice with some java8 :D

PR #346, opened 2017-03-17
Ensure proper handling of RunConfiguration for DevAppserver v1

loosebazooka:
- handle multiple modules
  - establish rules and handle java8/java7 detection
- warning messages for ignored parameters (that are only v2)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -145,4 +161,33 @@ public void stop(StopConfiguration configuration) throws AppEngineException {
       }
     }
   }
+
+  @VisibleForTesting
+  void checkAndWarnIgnored(Object valueToIgnore, String propertyName) {
```
akerekes: Minor: I'd put the property name first, like how in other situations usually the name/key is first and the value is second.

src/test/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1Test.java
```diff
@@ -118,4 +190,80 @@ public void testPrepareCommand_noFlags() throws AppEngineException, ProcessRunne
...
+        "--disable_update_check", pathToJava7Service);
+    Map<String, String> expectedEnv = ImmutableMap.of();
+
+    List<String> expectedJvmArgs = ImmutableList
+        .of("-javaagent:" + fakeJavaSdkHome.resolve("agent/appengine-agent.jar").toAbsolutePath()
+            .toString());
+
+    devServer.run(configuration);
+
+    verify(sdk, times(1)).runDevAppServer1Command(eq(expectedJvmArgs), eq(expectedFlags), eq(expectedEnv));
```
akerekes: If you only have non-mock arguments, `eq()` is not necessary.

src/test/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1Test.java
```diff
@@ -118,4 +190,80 @@ public void testPrepareCommand_noFlags() throws AppEngineException, ProcessRunne
...
+    List<String> expectedFlags = ImmutableList.of("--allow_remote_shutdown",
+        "--disable_update_check", "--no_java_agent", pathToJava7Service, pathToJava8Service);
+    Map<String, String> expectedEnv = ImmutableMap.of();
+
+    List<String> expectedJvmArgs = ImmutableList.of("-Duse_jetty9_runtime=true",
+        "-D--enable_all_permissions=true");
+
+    devServer.run(configuration);
+
+    verify(sdk, times(1)).runDevAppServer1Command(eq(expectedJvmArgs), eq(expectedFlags), eq(expectedEnv));
```
akerekes: If you only have non-mock arguments, `eq()` is not necessary.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -76,26 +72,46 @@ public void run(RunConfiguration config) throws AppEngineException {
...
+    checkAndWarnIgnored(config.getClearDatastore(), "clearDatastore");
+    checkAndWarnIgnored(config.getDevAppserverLogLevel(), "devAppserverLogLevel");
+    checkAndWarnIgnored(config.getSkipSdkUpdateCheck(), "skipSdkUpdateCheck");
 
     arguments.add("--allow_remote_shutdown");
     arguments.add("--disable_update_check");
-    if (appengineWeb.isJava8()) {
+    if (determineJavaRuntimeVersion(config.getServices()).equals("java8")) {
+      jvmArguments.add("-Duse_jetty9_runtime=true");
+      jvmArguments.add("-D--enable_all_permissions=true");
```
akerekes: this `--` looks a bit weird

src/test/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1Test.java
```diff
@@ -118,4 +190,80 @@ public void testPrepareCommand_noFlags() throws AppEngineException, ProcessRunne
...
+  @Test
+  public void testCheckAndWarnIgnored_withUnsetValue() {
+    LogStoringHandler testHandler = LogStoringHandler.getForLogger(CloudSdkAppEngineDevServer1.class.getName());
+    devServer.checkAndWarnIgnored(null, "testName");
+
+    Assert.assertEquals(0, testHandler.getLogs().size());
+  }
+
+  @Test
+  public void testDetermineJavaRuntime() {
```
akerekes: I'd split up these into separate test methods.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -76,26 +72,46 @@ public void run(RunConfiguration config) throws AppEngineException {
...
     arguments.addAll(DevAppServerArgs.get("address", config.getHost()));
     arguments.addAll(DevAppServerArgs.get("port", config.getPort()));
     if (config.getJvmFlags() != null) {
       jvmArguments.addAll(config.getJvmFlags());
     }
+    arguments.addAll(DevAppServerArgs.get("default_gcs_bucket", config.getDefaultGcsBucketName()));
+    // TODO: It's not really clear what the behavior here is when this is specified
+    arguments.addAll(DevAppServerArgs.get("runtime", config.getRuntime()));
+
+    // Arguments ignore by dev appserver 1
```
elharo: ignored

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -145,4 +161,33 @@ public void stop(StopConfiguration configuration) throws AppEngineException {
...
+      log.warning(propertyName + " will be ignored by Dev Appserver v1");
+    }
+  }
+
+  @VisibleForTesting
+  String determineJavaRuntimeVersion(List<File> services) {
+    boolean java8Detected = false;
+    boolean java7Detected = false;
+    for (File serviceDirectory : services) {
+      File appengineWebXml = new File(serviceDirectory, "WEB-INF/appengine-web.xml");
```
joaoandremartins: Use Paths.get().resolve?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -145,4 +161,33 @@ public void stop(StopConfiguration configuration) throws AppEngineException {
...
+
+  @VisibleForTesting
+  void checkAndWarnIgnored(Object valueToIgnore, String propertyName) {
+    if (valueToIgnore != null) {
+      log.warning(propertyName + " will be ignored by Dev Appserver v1");
+    }
+  }
+
+  @VisibleForTesting
+  String determineJavaRuntimeVersion(List<File> services) {
```
joaoandremartins: Why isn't this a method of AppEngineDescriptor?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -76,26 +72,46 @@ public void run(RunConfiguration config) throws AppEngineException {
...
-      jvmArguments.add("-javaagent:" + appengineAgentJar);
-    } 
     arguments.addAll(DevAppServerArgs.get("address", config.getHost()));
     arguments.addAll(DevAppServerArgs.get("port", config.getPort()));
     if (config.getJvmFlags() != null) {
       jvmArguments.addAll(config.getJvmFlags());
     }
+    arguments.addAll(DevAppServerArgs.get("default_gcs_bucket", config.getDefaultGcsBucketName()));
+    // TODO: It's not really clear what the behavior here is when this is specified
+    arguments.addAll(DevAppServerArgs.get("runtime", config.getRuntime()));
```
patflynn: I didn't even know there was such a parameter. What does it so?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -76,26 +72,46 @@ public void run(RunConfiguration config) throws AppEngineException {
...
+    checkAndWarnIgnored(config.getMaxModuleInstances(), "maxModuleInstances");
+    checkAndWarnIgnored(config.getUseMtimeFileWatcher(), "useMtimeFileWatcher");
+    checkAndWarnIgnored(config.getThreadsafeOverride(), "threadsafeOverride");
+    checkAndWarnIgnored(config.getPythonStartupScript(), "pythonStartupScript");
+    checkAndWarnIgnored(config.getPythonStartupArgs(), "pythonStartupArgs");
+    checkAndWarnIgnored(config.getCustomEntrypoint(), "customEntrypoint");
+    checkAndWarnIgnored(config.getAllowSkippedFiles(), "allowSkippedFiles");
+    checkAndWarnIgnored(config.getApiPort(), "apiPort");
+    checkAndWarnIgnored(config.getAutomaticRestart(), "automaticRestart");
+    checkAndWarnIgnored(config.getClearDatastore(), "clearDatastore");
```
patflynn: Isn't there a way to configure the old devappserver to do this?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
loosebazooka: yeah, it's something that was added before me, I just moved it around

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -76,26 +72,46 @@ public void run(RunConfiguration config) throws AppEngineException {
...
+    checkAndWarnIgnored(config.getDevAppserverLogLevel(), "devAppserverLogLevel");
+    checkAndWarnIgnored(config.getSkipSdkUpdateCheck(), "skipSdkUpdateCheck");
```
patflynn: there's a disable update check in the devappserver 1 flags too.

src/test/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1Test.java
loosebazooka: oh yeah I just moved this code around, I never noticed it, but I can change it.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -145,4 +161,33 @@ public void stop(StopConfiguration configuration) throws AppEngineException {
       }
     }
   }
+
+  @VisibleForTesting
+  void checkAndWarnIgnored(Object valueToIgnore, String propertyName) {
+    if (valueToIgnore != null) {
+      log.warning(propertyName + " will be ignored by Dev Appserver v1");
```
patflynn: it'd be nice to explain why you're ignoring the flag here.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
loosebazooka: It's actually testing more than one service to determine what runtime we should assume. I think I should add a comment on top of this.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
loosebazooka: will do.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -76,26 +72,46 @@ public void run(RunConfiguration config) throws AppEngineException {
...
+    checkAndWarnIgnored(config.getApiPort(), "apiPort");
+    checkAndWarnIgnored(config.getAutomaticRestart(), "automaticRestart");
+    checkAndWarnIgnored(config.getClearDatastore(), "clearDatastore");
+    checkAndWarnIgnored(config.getDevAppserverLogLevel(), "devAppserverLogLevel");
+    checkAndWarnIgnored(config.getSkipSdkUpdateCheck(), "skipSdkUpdateCheck");
 
     arguments.add("--allow_remote_shutdown");
     arguments.add("--disable_update_check");
-    if (appengineWeb.isJava8()) {
+    if (determineJavaRuntimeVersion(config.getServices()).equals("java8")) {
```
patflynn: since this is the only place that `determineJavaRuntimeVersion` is called would it make more sense to push all the logic into the method and have it just return a boolean? e.g. `bool requiresJava8(services)`

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -76,26 +72,46 @@ public void run(RunConfiguration config) throws AppEngineException {
...
+      jvmArguments.add("-Duse_jetty9_runtime=true");
```
patflynn: how did you know to configure the flags like this? Is this documented somewhere?

loosebazooka: @ludoch ?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
loosebazooka: yeah I talked to @ludoch about this. That nag mentions the old SDK, it's not clear whether we want that messaging. we just disable the update check by default here.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
loosebazooka: well then the devappserver1 will have to know about devappserver2 flags?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
loosebazooka: @ludoch ?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
patflynn: I see. makes sense.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
patflynn: I was thinking we can just do the mapping (if there is one) here.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
loosebazooka: yup, done.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
loosebazooka: done

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
loosebazooka: It does some internal stuff in the devappserver. It seems "kickstart" code and devappserver code kind of overlap is some way that I don't understand. @ludoch ?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
loosebazooka: @ludoch ?

loosebazooka: I think I misunderstood what you meant. I thought you were talking about all the flags. I don't know if clearDatstore has a corresponding value in devappserver1 or how we would emulate that behavior. @ludoch ?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
loosebazooka: done

src/test/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1Test.java
loosebazooka: done

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
loosebazooka: I think the property name is just a "hint" here for the error message. The subject of the method is valueToIgnore. I mean I'd like to keep it like this for now, but open to discussing it.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
ludoch: for devappserver1, the local datastore is in the web-inf/ generated area, so when you do a mvn/gradle clean, you reset the datastore.
There is maybe a flag to control this location. can't remember.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
ludoch: Not all flags are documented as they are for our internal tooling purpose.
Consider yourself are an internal user, and use these hidden flags.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
ludoch: move it maybe, but keep it, it disable the local sandbox security code, not needed for java8

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
loosebazooka: Oh right : 
```
jvmFlags = ['-Ddatastore.backing_store="/path/to/datastore/file/location/local_db.bin']
```

patflynn:
Let's talk about how we want to support the clear datastore flag for
devappserverv1 when I get back.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
patflynn: @ludoch do you know where devappserverv2 stores the local datastore file?

patflynn: Does it use that datastore environment variable?

ludoch: no,  I dnk...

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
ludoch: Which parameter? runtime? Not clear.

joaoandremartins:
By the way, with devappserver1, do we also want to add another condition to validateCloudSdkLocation() that probes the location of devappserver1?

joaoandremartins:
DevAppServerMain is in appengine-tools-api.jar, so the current check for the app-engine-java component should suffice.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
briandealwis: I see `$TMP/appengine.<app>.$USER/datastore.db`, where `<app>` is `None` if there's no `<application>` element in the `appengine-web.xml`.

patflynn: @loosebazooka 
@briandealwis 
@ludoch 

do you guys think it would be too hacky to simulate the clearDatastore flag on v1 by putting the file under tmp if the flag is set to false?

briandealwis: Simulating the datastore options does sound like a good idea.  But the devappserver2's `—clear-datastore` flag works with the `—storage_path` argument (and `—datastore-path` too, which defaults to `—storage-path`), and we should respect those if we can.

And seems to me that putting the datastore under `WEB-INF` won't play well with multiple services?

patflynn:
@loosebazooka 

I'd love to merge this today to get a release out that includes this and Joao's JDK fix.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
loosebazooka: Sounds, good, I'll use --datastore-path and --clear-datastore to emulate the behavior

loosebazooka:
take another look!

PR #348, opened 2017-03-22
kokoro build files

akerekes:


build/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
loosebazooka: Just wondering:
build is a directory that gradle generates also and we tend to ignore it, is this a convention or can it be anything?

kokoro/ubuntu/continuous.cfg
```diff
@@ -0,0 +1,4 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# Location of the continuous bash script in repository.
+build_file: "appengine-plugins-core/kokoro/ubuntu/continuous.sh"
```
loosebazooka: So for windows/mac we just add in more buildfiles here?

build/build.sh
loosebazooka: I mean this repo is maven build, but if we transition others.

kokoro/ubuntu/continuous.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
loosebazooka: What's the difference between build and continuous?

build/build.sh
akerekes: We can call this directory something else than `build`, e.g. `build-scripts`?

kokoro/ubuntu/continuous.cfg
akerekes: We'll create separate directories for Win and Mac under the `kokoro` directory and they'll refer their respective files.

kokoro/ubuntu/continuous.sh
akerekes: Actually we can get rid of `build.sh` and add the same logic into `continuous.sh`, and move `continuous.sh` up to the `kokoro` directory (and share it between Ubuntu and Mac)

kokoro/ubuntu/continuous.cfg
```diff
@@ -0,0 +1,4 @@
+# Format: //devtools/kokoro/config/proto/build.proto
+
+# Location of the continuous bash script in repository.
```
elharo: continuous build bash

PR #350, opened 2017-03-22
More fixes to make the kokoro build successful

akerekes:


kokoro/continuous.sh
```diff
@@ -6,4 +6,6 @@ set -e
 set -x
 
 cd github/appengine-plugins-core
+sudo /opt/google-cloud-sdk/bin/gcloud components update
```
loosebazooka: kokoro already has gcloud?

akerekes: yes, but currently it's on version 141

PR #352, opened 2017-03-27
Makes Java SDK be provided at CloudSdk instantiation time.

joaoandremartins:
Removes Java SDK from RunConfiguration.

This fix is instrumental in fixing a bigger issue in the IJ plugin, https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/1316, so your quick review is appreciated!

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -280,7 +280,7 @@ public class CloudSdk {
 
     List<String> command = new ArrayList<>();
 
-    command.add(getJavaExecutablePath().toAbsolutePath().toString());
+    command.add(getJavaExecutablePath().toString());
 
     command.addAll(jvmArgs);
     command.add("-Dappengine.sdk.root=" + getJavaAppEngineSdkPath().getParent().toString());
@@ -317,7 +317,7 @@ public class CloudSdk {
     System.setProperty("appengine.sdk.root", getJavaAppEngineSdkPath().toString());
 
     List<String> command = new ArrayList<>();
-    command.add(getJavaExecutablePath().toAbsolutePath().toString());
+    command.add(getJavaExecutablePath().toString());
     command.add("-cp");
     command.add(jarLocations.get(JAVA_TOOLS_JAR).toString());
     command.add("com.google.appengine.tools.admin.AppCfg");
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -280,7 +280,7 @@ public class CloudSdk {
 
     List<String> command = new ArrayList<>();
 
-    command.add(getJavaBinaryPath().toAbsolutePath().toString());
+    command.add(getJavaExecutablePath().toAbsolutePath().toString());
 
     command.addAll(jvmArgs);
     command.add("-Dappengine.sdk.root=" + getJavaAppEngineSdkPath().getParent().toString());
@@ -317,7 +317,7 @@ public class CloudSdk {
     System.setProperty("appengine.sdk.root", getJavaAppEngineSdkPath().toString());
 
     List<String> command = new ArrayList<>();
-    command.add(getJavaBinaryPath().toAbsolutePath().toString());
+    command.add(getJavaExecutablePath().toAbsolutePath().toString());
     command.add("-cp");
     command.add(jarLocations.get(JAVA_TOOLS_JAR).toString());
     command.add("com.google.appengine.tools.admin.AppCfg");
@@ -406,8 +406,9 @@ public class CloudSdk {
   }
 
   @VisibleForTesting
-  Path getJavaBinaryPath() {
-    return javaHomePath.resolve("bin/java");
+  Path getJavaExecutablePath() {
+    return javaHomePath.toAbsolutePath().resolve(
+        System.getProperty("os.name").contains("Windows") ? "bin/java.exe" : "bin/java");
   }
 
   // https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/189
@@ -488,9 +489,9 @@ public class CloudSdk {
   }
 
   private void validateJdk() {
-    if (!Files.exists(getJavaBinaryPath())) {
+    if (!Files.exists(getJavaExecutablePath())) {
       throw new InvalidJavaSdkException(
-          "Invalid Java SDK. " + getJavaBinaryPath().toString() + " does not exist.");
+          "Invalid Java SDK. " + getJavaExecutablePath().toString() + " does not exist.");
     }
   }
 
@@ -653,8 +654,8 @@ public class CloudSdk {
     /**
      * Sets the desired Java SDK path, used in devappserver runs and App Engine standard staging.
      */
-    public Builder javaSdkPath(Path javaSdkPath) {
-      this.javaHomePath = javaSdkPath;
+    public Builder javaHome(Path javaHomePath) {
+      this.javaHomePath = javaHomePath;
       return this;
     }
 
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -181,7 +181,7 @@ public class CloudSdk {
     // This is to ensure IDE credentials get correctly passed to the gcloud commands, in Windows.
     // It's a temporary workaround until a fix is released.
     // https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/985
-    if (System.getProperty("os.name").contains("Windows")) {
+    if (IS_WINDOWS) {
       environment.put("CLOUDSDK_APP_NUM_FILE_UPLOAD_PROCESSES", "1");
     }
 
@@ -407,8 +407,7 @@ public class CloudSdk {
 
   @VisibleForTesting
   Path getJavaExecutablePath() {
-    return javaHomePath.toAbsolutePath().resolve(
-        System.getProperty("os.name").contains("Windows") ? "bin/java.exe" : "bin/java");
+    return javaHomePath.toAbsolutePath().resolve(IS_WINDOWS ? "bin/java.exe" : "bin/java");
   }
 
   // https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/189
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -31,7 +31,6 @@ import com.google.cloud.tools.appengine.cloudsdk.serialization.CloudSdkComponent
 import com.google.cloud.tools.appengine.cloudsdk.serialization.CloudSdkVersion;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
-import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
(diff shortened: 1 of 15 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -86,11 +87,13 @@
   private final String appCommandOutputFormat;
   private final WaitingProcessOutputLineListener runDevAppServerWaitListener;
 
-  private CloudSdk(Path sdkPath, String appCommandMetricsEnvironment,
-      String appCommandMetricsEnvironmentVersion, @Nullable File appCommandCredentialFile,
-      String appCommandOutputFormat, ProcessRunner processRunner,
-      WaitingProcessOutputLineListener runDevAppServerWaitListener) {
+  private CloudSdk(Path sdkPath, Path javaBinaryPath, String appCommandMetricsEnvironment,
```
elharo: javaBinaryPath doesn't feel like the usual terminology for this. The common name for this directory is javaHome or jdkHome

joaoandremartins: The constructor isn't part of the API. The API in Builder only surfaces a javaSdkHome() method.

joaoandremartins:
@elharo why is it very strange? The library was making that decision by itself until now, using System.getProperty("java.home") (or RunConfiguration.getJavaSdkHome() for devappserver). We're giving the option to specify the JDK and only have a single place to configure it, while still assuming the default of System.getProperty("java.home").

This is useful, for example, in IntelliJ, when we want to use the Project JDK instead of the bundled JRE.

joaoandremartins:
@elharo I don't recall there being an auto-discovery for Java SDKs in the common library. You might be confusing with auto-discovering the Cloud SDK path, or the Python binary. For Java, we've always assumed System.getProperty("java.home"), until we had to add a Java SDK location to RunConfiguration, to specify a JRE to run the devappserver.

And this code is perfectly backwards compatible. If you don't call Builder().javaSdkPath(), it will assume the default that used to be.

patflynn:
I'm still a little unsure why the caller of the client library can't simply
set java.home or JAVA_HOME.

joaoandremartins:
I think it can be a bit intrusive to be setting java.home on the process that's running the library client (e.g., IJ or Maven). Users may not want that to be changed, and the library call is a synchronous Java library call, so it's made in the same process.

Also, there would be an API for selecting the JRE.

elharo:
We may need to back up and talk through this. My first question is why does a specific JDK need to be chosen at all? What in this library is dependent on a specific JDK?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -308,6 +292,9 @@ public void runDevAppServer1Command(List<String> jvmArgs, List<String> args,
     command.addAll(args);
 
     logCommand(command);
+
+    Map<String, String> environment = Maps.newHashMap();
+    environment.put("JAVA_HOME", javaBinaryPath.getParent().getParent().toString());
```
meltsufin: I don't think these 2 lines are needed.

joaoandremartins:
@elharo I'm wfhing today. @patflynn might be able to explain why this is needed in person. If not, I'm happy to jump into a hangout with you at any time.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
joaoandremartins: You mean devappserver1 doesn't internally use JAVA_HOME? I was under the impression it did, after talking with @loosebazooka and @ludoch ?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -536,6 +531,7 @@ WaitingProcessOutputLineListener getRunDevAppServerWaitListener() {
     private List<CloudSdkResolver> resolvers;
     private int runDevAppServerWaitSeconds;
     private boolean inheritProcessOutput;
+    private Path javaBinaryPath = Paths.get(System.getProperty("java.home"));
```
patflynn: perhaps we should just store the java home, since that's what most tools present as the path of interest to the user.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
patflynn: how would this deal with symlinks?

patflynn:
@joaoandremartins overall after discussion with Mike and Alex I think this is the right approach. It does seem like it could be bad to set the environment in the parent, and it seems reasonable to have a means of configuring the jdk in the API.

patflynn:
@elharo both running the dev appserver and staging should be done using the JDK of the project.

elharo:
dev appserver I can sort of see, though that ideally shouldn't matter too much.

Why does the JDK used for staging matter?

patflynn:
jsp compilation, quickstart annotation processing, ... (who knows what
else?)

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
joaoandremartins: It would attempt to invoke Java using its symlink.
I can't think of any issues that could arise from that.

chanseokoh: For example, if the `java` binary is under `/usr/bin/java` via a symlink, you'd end up putting the root to `JAVA_HOME`?

joaoandremartins: That would fail on validateJdk(), so we wouldn't even get to this point.
But PTAL

joaoandremartins:
@patflynn PTAL, I think your comments were addressed

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -419,6 +405,11 @@ public Path getJavaAppEngineSdkPath() {
     return getSdkPath().resolve(JAVA_APPENGINE_SDK_PATH);
   }
 
+  @VisibleForTesting
+  Path getJavaBinaryPath() {
+    return javaHomePath.resolve("bin/java");
```
patflynn: should this just return javaHomePath.resolve("bin/java").getRealPath()?  It seems like you're always canonicalize it from the caller.

briandealwis: Will this work on Windows?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -495,6 +487,13 @@ private void validateCloudSdkLocation() {
     }
   }
 
+  private void validateJdk() {
+    if (!Files.exists(getJavaBinaryPath())) {
```
briandealwis: This won't work on Windows with the current `getJavaBinaryPath()`

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
briandealwis: Perhaps this should be `getJavaExecutable()`. It should do an is-windows test and check for the java.exe.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -651,6 +651,14 @@ public Builder inheritProcessOutput(boolean inheritProcessOutput) {
     }
 
     /**
+     * Sets the desired Java SDK path, used in devappserver runs and App Engine standard staging.
+     */
+    public Builder javaSdkPath(Path javaSdkPath) {
```
briandealwis: This is an odd name as there is no Java SDK: there's a JRE and a JDK, and JDKs usually include the JRE in `jre/`.  `javaHome()` or `javaInstallRoot()`?

<hr>
I hesitate in suggesting the following…

It would be more flexible if we could instead provide a JVM runner/launcher object.  In the Eclipse JDT world, the user can configure multiple JREs, each of which can have additional VM arguments and libraries to form the bootclasspath, and there's already infrastructure for auto-finding JVMs and determining these flags.
Perhaps something like:
```
interface JvmRunner {
   void validate() throws AppEngineException;
   void run(String mainClass, String[] args, String[] jvmArgs, String[] classpath);
}
```
and move much of the changes above into a implementation `StandardJavaRunner`?  Which we would ignore and instead create a wrapper around the JDT `IVMRunner`.

My concern is that what we execute may not match what the user has configured in their JRE description

briandealwis: On the flip side, JAVA_HOME is exactly what's being done by devappserver2.  So we'll have this issue anyways.

joaoandremartins: Hum, I always thought there was a Java System Development Kit, but turns out there's only a Java Development Kit. Thanks for the tip!

Your suggested solution sounds too engineered for the problem at hand, and it binds us too much to there being a separate process being ran by a specific JDK which, in an ideal world, would just take the form of a Java API call.

It would also require a serious overhaul of the current ProcessRunner architecture, which I'm not convinced is worth revising.

I've also had ideas regarding validators for appengine-plugins-core for a year now, that I was never able to get around. But I wouldn't want to bundle them with other entities, like runners.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
joaoandremartins: Good catch.

I kind of assumed it would work because, if you look at the previously existing code, it never makes Windows checks and the devappserver somehow still works...

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
patflynn: @briandealwis could you file an issue for this?  I think your suggestion makes sense and we should spend some time thinking it through for both devappservers, but I don't want to block this change on it as we need to get a fix out soon.

patflynn: @joaoandremartins let me know once you've addressed Brian's other issues and I'll approve.

briandealwis: #353

src/test/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkTest.java
```diff
@@ -213,4 +213,10 @@ public void testResolverCascading() {
     CloudSdk sdk = builder.build();
     assertEquals("r1 should not resolve", r2.getCloudSdkPath(), sdk.getSdkPath());
   }
+
+  @Test
+  public void testGetJavaBinary() {
```
patflynn: would this test fail on windows?

joaoandremartins: Fixed. On the topic of failing tests in Windows, I noticed there's a bunch of others failing.
@meltsufin maybe we could look into turning on AppVeyor for appengine-plugins-core too?

joaoandremartins:
@patflynn PTAL

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -280,24 +268,19 @@ public void runDevAppServerCommand(List<String> args, Map<String, String> enviro
...
 
     List<String> command = new ArrayList<>();
-    String javaHome = environment.get("JAVA_HOME");
-    if (javaHome == null) {
-      javaHome = System.getProperty("java.home");
-    }
-    command.add(
-            Paths.get(javaHome).resolve("bin/java").toAbsolutePath().toString());
+
+    command.add(getJavaExecutablePath().toAbsolutePath().toString());
```
joaoandremartins: Since getJavaExecutablePath() is already calling toAbsolutePath(), need to remove this from here.

PR #355, opened 2017-03-29
Avoid possible NPE

briandealwis:
`finalPath` can be null.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
@@ -17,7 +17,6 @@
 package com.google.cloud.tools.appengine.cloudsdk;
 
 import com.google.common.annotations.VisibleForTesting;
-
 import java.io.File;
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
```diff
@@ -66,7 +65,7 @@ public Path getCloudSdkPath() {
     }
 
     Path finalPath = searchPaths(possiblePaths);
-    logger.log(Level.FINE, "Resolved SDK path : " + finalPath.toString());
+    logger.log(Level.FINE, "Resolved SDK path : " + finalPath);
     return searchPaths(possiblePaths);
```
chanseokoh: Odd. Shouldn't it just return `finalPath`?

briandealwis: Well-spotted optimization :)

loosebazooka:
I think @akerekes has this fix in #356

PR #356, opened 2017-03-29
Win build fixes

akerekes:
test fixes for Windows build failures

### DIFF
--- a/src/test/java/com/google/cloud/tools/io/FilePermissionsTest.java
+++ b/src/test/java/com/google/cloud/tools/io/FilePermissionsTest.java
@@ -48,7 +48,7 @@ public class FilePermissionsTest {
   
   @Test // Unix only
   public void testSubDirectoryCannotBeCreatedInDevNull()  {
-    Assume.assumeTrue(!System.getProperty("os.name").startsWith( "Windows"));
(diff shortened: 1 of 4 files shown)


src/test/java/com/google/cloud/tools/io/FilePermissionsTest.java
```diff
@@ -79,8 +79,9 @@ public void testSubDirectoryCannotBeCreatedDueToPreexistingFile() throws IOExcep
     }
   }
   
-  @Test
+  @Test // Unix only
   public void testDirectoryCannotBeCreatedDueToUnwritableParent() throws IOException {
```
loosebazooka: what makes these unix only?

briandealwis: Apparently [Windows doesn't actually support read-only directories](http://stackoverflow.com/questions/25034068/how-to-make-directory-read-only-in-windows-through-java), and `setWritable(false) == false` (i.e., operation failed).  So I think this change, and the other similar changes, would be better encoded as:
```
Path dir = Files.createDirectory(Paths.get(parent.toString(), "child"));
Assume.assumeTrue(dir.toFile().setWritable(false));
```

PR #363, opened 2017-03-30
Fixes #361 - Flex Staging should only copy app.yaml out of appengineDirectory

### LINKED ISSUE #361: Flex Staging should only copy app.yaml out of appengineDirectory
We don't need to copy the other yaml configs into the destination folder anymore.

akerekes:


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
```diff
@@ -114,13 +110,14 @@ static void copyDockerContext(StageFlexibleConfiguration config, CopyService cop
     }
   }
 
+  @VisibleForTesting
   static void copyAppEngineContext(StageFlexibleConfiguration config, CopyService copyService)
       throws IOException {
     // Copy app.yaml and other App Engine config files to staging
```
loosebazooka: can you update this comment

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
```diff
@@ -114,13 +110,14 @@ static void copyDockerContext(StageFlexibleConfiguration config, CopyService cop
...
     String[] appEngineConfigFiles = config.getAppEngineDirectory().list();
     if (appEngineConfigFiles != null) {
       for (String configFile : appEngineConfigFiles) {
-        if (APP_ENGINE_CONFIG_FILES_WHITELIST.contains(configFile)) {
+        if (APP_YAML.equals(configFile)) {
           copyService.copyFileAndReplace(
               config.getAppEngineDirectory().toPath().resolve(configFile),
               config.getStagingDirectory().toPath().resolve(configFile));
```
loosebazooka: line 128 needs to be modified, we're don't fail if we find "unexpected" configuration files. We just copy app.yaml from that directory.

I think this whole loop might need to change, no more loop?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
```diff
@@ -113,27 +112,15 @@ static void copyDockerContext(StageFlexibleConfiguration config, CopyService cop
...
+    if (!appYaml.toFile().exists()) {
+      throw new AppEngineException(APP_YAML + " not found in the App Engine directory.");
     }
+    copyService.copyFileAndReplace(appYaml, 
+        config.getStagingDirectory().toPath().resolve(APP_YAML));
   }
 
-  @VisibleForTesting
-  static void copyArtifact(StageFlexibleConfiguration config, CopyService copyService)
+  private static void copyArtifact(StageFlexibleConfiguration config, CopyService copyService)
```
loosebazooka: Maybe this should have been tested? Did you remove it because it wasn't used? I think when I refactored this to test the other stuff, I didn't add the test in. We probably still want a test for it.

akerekes: Yes, this and the other methods weren't called from anywhere but this class, so I've reduced their visibility. I'll create a new PR to test them.

PR #366, opened 2017-03-30
Update snapshot version to 0.3.0-SNAPSHOT

loosebazooka:


loosebazooka:
waiting on the actual release to happen...

elharo:
should we maybe declare 0.3 rather than using double digit minor versions?

PR #368, opened 2017-03-31
Fixes #367 - appYamls vs Services

### LINKED ISSUE #367: appYamls vs Services
We should remove appYamls from devAppServer2 and just use services?

akerekes:
Remove `RunConfiguration.appYamls` and use `RunConfiguration.services` instead.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
@@ -17,7 +17,6 @@
 package com.google.cloud.tools.appengine.api.devserver;
 
 import java.io.File;
-import java.nio.file.Path;
 import java.util.List;
 
 /**
@@ -25,7 +24,7 @@ import java.util.List;
  */
 public class DefaultRunConfiguration implements RunConfiguration {
 
-  private List<Path> services;
+  private List<File> services;
   private String host;
   private Integer port;
   private String adminHost;
@@ -51,11 +50,11 @@ public class DefaultRunConfiguration implements RunConfiguration {
   private File datastorePath;
 
   @Override
-  public List<Path> getServices() {
+  public List<File> getServices() {
     return services;
   }
 
-  public void setServices(List<Path> services) {
+  public void setServices(List<File> services) {
     this.services = services;
   }
 
--- a/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
@@ -24,7 +24,6 @@ import java.util.List;
  */
 public class DefaultRunConfiguration implements RunConfiguration {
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
```diff
@@ -51,16 +50,6 @@
...
-    return appYamls;
-  }
-
-  public void setAppYamls(List<File> appYamls) {
-    this.appYamls = appYamls;
-  }
-
-
-  @Override
   public List<File> getServices() {
```
joaoandremartins: Might want to add some Javadoc here that mentions these should be directories for flexible and app.yaml's for standard (or vice-versa)?

joaoandremartins: Might also be a good opportunity to turn these from File's (which they should have never been) into Path's.

akerekes: There is some javadoc on the method in `RunConfiguration`. Since devappserver doesn't support flex, and local runs are executed from the build directory (not stage where app.yamls are), I think the javadoc is correct, but feel free to correct me.

akerekes: I'll look into it

joaoandremartins: Sorry, didn't realise I was looking at the implementation.

src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
```diff
@@ -17,15 +17,15 @@
...
 import java.util.List;
 
 /**
  * Plain Java bean implementation of {@link RunConfiguration}.
  */
 public class DefaultRunConfiguration implements RunConfiguration {
 
-  private List<File> appYamls;
-  private List<File> services;
+  private List<Path> services;
```
loosebazooka: We don't want to use Path just yet as it messed up the gradle configuration unfortunately.

loosebazooka: I probably need to a better job there, lemme see if I can figure that out and we can do a larger scale migration to Path in a future PR

joaoandremartins: How did it mess it up?
If this is being annotated on the Gradle side for the value to be automatically initialized, I'm wondering if we could add another field to Gradle's configuration and then translate those Files into Paths at some point in the run workflow?

loosebazooka: No, it's the auto generation of configuration closures in the build.gradle that requires files at the moment.

PR #369, opened 2017-04-01
Move StringBuilderProcessOutputLineListener out of internal 

elharo:
so Eclipse can use it

### DIFF
--- a/src/test/java/com/google/cloud/tools/appengine/cloudsdk/StringBuilderProcessOutputLineListenerTest.java
+++ b/src/test/java/com/google/cloud/tools/appengine/cloudsdk/StringBuilderProcessOutputLineListenerTest.java
@@ -14,19 +14,18 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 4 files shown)


src/test/java/com/google/cloud/tools/appengine/cloudsdk/StringBuilderProcessOutputLineListenerTest.java
```diff
@@ -14,19 +14,18 @@
  * limitations under the License.
  */
 
-package com.google.cloud.tools.appengine.cloudsdk.internal.process;
+package com.google.cloud.tools.appengine.cloudsdk;
```
chanseokoh: Shouldn't this be `c.g.c.t.a.cloudsdk.process`? That'll also remove the `import` below.

elharo: you're right; done

PR #371, opened 2017-04-04
allow non default admin port

### LINKED ISSUE #370: Cannot stop dev appserver1 if running on different port than the default 8080
The admin port (which is just same as the server port in dev appserver1) in `CloudSdkAppEngineDevServer1.stop()` is fixed to be 8080.

https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/v0.2.11/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java#L136

I think it should be like
```
configuration.getAdminPort() != null ? configuration.getAdminPort() : DEFAULT_PORT,
```

elharo:
@chanseokoh @loosebazooka fix #370

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
@@ -125,13 +125,13 @@ public class CloudSdkAppEngineDevServer1 implements AppEngineDevServer {
   public void stop(StopConfiguration configuration) throws AppEngineException {
     Preconditions.checkNotNull(configuration);
     HttpURLConnection connection = null;
+    String adminHost =
+        configuration.getAdminHost() != null ? configuration.getAdminHost() : DEFAULT_HOST;
+    int adminPort =
+        configuration.getAdminPort() != null ? configuration.getAdminPort() : DEFAULT_PORT;
+    URL adminServerUrl = null;
     try {
-      URL adminServerUrl = new URL(
-              "http",
-              configuration.getAdminHost() != null
-              ? configuration.getAdminHost() : DEFAULT_HOST,
-              DEFAULT_PORT,
-              "/_ah/admin/quit");
+      adminServerUrl = new URL("http", adminHost, adminPort, "/_ah/admin/quit");
       connection = (HttpURLConnection) adminServerUrl.openConnection();
       connection.setDoOutput(true);
       connection.setDoInput(true);
@@ -142,10 +142,10 @@ public class CloudSdkAppEngineDevServer1 implements AppEngineDevServer {
       int responseCode = connection.getResponseCode();
       if (responseCode < 200 || responseCode > 299) {
         throw new AppEngineException(
-                "The development server responded with " + connection.getResponseMessage() + ".");
+            adminServerUrl + " responded with " + connection.getResponseMessage() + ".");
       }
-    } catch (IOException e) {
-      throw new AppEngineException(e);
+    } catch (IOException ex) {
+      throw new AppEngineException("Error connecting to " + adminServerUrl, ex);
     } finally {
       if (connection != null) {
         try {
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -125,13 +125,13 @@ public void run(RunConfiguration config) throws AppEngineException {
   public void stop(StopConfiguration configuration) throws AppEngineException {
     Preconditions.checkNotNull(configuration);
     HttpURLConnection connection = null;
+    String adminHost =
+        configuration.getAdminHost() != null ? configuration.getAdminHost() : DEFAULT_HOST;
```
patflynn: are these even valid parameters to devappserverv1?

chanseokoh: At least the port param is valid in the sense that it is the only way to shutdown the server if it starts on a port other than 8080.

chanseokoh: Ah, you are saying the term "admin port" does not even apply to dev appserver1?

patflynn:
but there is no separation between the server port and the admin port
right? Shouldn't it just reuse the port specified for the server?

patflynn:
~ $ ./sdks/appengine-java-sdk-1.9.46/bin/dev_appserver.sh
Usage: <dev-appserver> [options] <app directory>

Options:
 --help, -h                 Show this help message and exit.
 --sdk_root=DIR             Overrides where the SDK is located.
 --server=SERVER            The server to use to determine the latest
  -s SERVER                   SDK version.
 --address=ADDRESS          The address of the interface on the local
machine
  -a ADDRESS                  to bind to (or 0.0.0.0 for all interfaces).
 --port=PORT                The port number to bind to on the local machine.
  -p PORT
 --disable_update_check     Disable the check for newer SDK versions.
 --generated_dir=DIR        Set the directory where generated files are
created.
 --default_gcs_bucket=NAME  Set the default Google Cloud Storage bucket
name.
 --jvm_flag=FLAG            Pass FLAG as a JVM argument. May be repeated to
                              supply multiple flags.

chanseokoh:
- We could do the same thing that we did for `RunConfiguration`: https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/v0.2.11/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java#L85
Then, I guess this requires us to extend the `StopConfiguration` interface with two additional methods.

- Alternatively, we could just rename `getAdminHost()` and `getAdminPort()` to something else. I don't like `getHost()` and `getPort()`, as they could confuse clients of the library, particularly when we use dev appserver2 eventually. Maybe `getStopPort()`?

patflynn:
I think you should just remove those fields from the UI.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
loosebazooka: @chanseokoh that is correct. Dev Appserver 1 does not have the concept of admin port. The "stop" listens on the port the app server was launched from.

loosebazooka:
Is there something going wrong with devappserver1 that it is not stopping?

chanseokoh:
Eclipse allows users to run the local server on an arbitrary port:

selection_007

Users can even set the port to 0, and the Eclipse plugin lets devappserver1 to choose a random port:

selection_008

So, to stop the server, we should send the HTTP request to the right port.

> I think you should just remove those fields from the UI.

I'm confused. You're suggesting to remove the users' ability to configure the server port, and to always use 8080?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
elharo: But we do need to specify the port in the URL. It isn't listening on 80.

loosebazooka:
Sorry, yeah I think we just need to remove "stopConfiguration" and use the runner configuration to stop the server.

chanseokoh:
> Sorry, yeah I think we just need to remove "stopConfiguration" and use the runner configuration to stop the server.

Not sure I like re-using `RunConfiguration`, for a few reasons. (For example, we may do `runConfig.setAdminPort(0)`or `runConfig.setPort(0)`, and we'd have to mutate the `runConfig` after making a copy to call `runConfig.setAdminPort(actualPortUsed)`. I'd just prefer `stop(String endpointUrl, int endpointPort)`.

PR #378, opened 2017-04-19
Extract environment settings from appengine-web.xml and set them for devappserver1 runs

etanshaul:
What do you think of this approach?

If ok, I'll go ahead and add unit tests.

I tested this with devappserver1 and IntelliJ.


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -194,4 +199,22 @@ boolean isJava8(List<File> services) {
     }
     return java8Detected;
   }
+
+  @VisibleForTesting
+  Map<String, String> getAllAppEngineEnvironment(List<File> services) {
+    Map<String, String> allAppEngineEnvironment = Maps.newHashMap();
+    for (File serviceDirectory : services) {
+      Path appengineWebXml = serviceDirectory.toPath().resolve("WEB-INF/appengine-web.xml");
```
etanshaul: I know theres some duplication here with isJava8; looking at refactoring a little.

patflynn:
The fact that the environment variables can clash between services looks
like a problem for the devappserver1 multi-module model.

etanshaul:
That's true. However isn't that a problem for devappserver2 as well? I wonder how it handles it.

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
```diff
@@ -99,16 +105,55 @@ public boolean isJava8() {
...
+
+  private static String getText(Node node) {
+    if (node != null) {
+      return node.getTextContent();
+    }
+
+    return null;
+  }
+
+  private static Map<String, String> getAttributeMap(Node node, String key, String value) {
```
etanshaul: I realized I need to update this to also search for a target child node by name. E.g.:

```
<env-variables>
        <env-var name="ENDPOINTS_SERVICE_NAME" value="some_value" />
 </env-variables>
```
The logic should probably search for the tag 'env-var'.

patflynn:
devappserver2 uses a different jvm for each service.

In any case this seems better than what we have now. Even if there's
potential for a clash.

etanshaul:
oh, I see. yeah the effect of the clash here is simply to use the value from the last service in the list.

patflynn:
that's probably worth warning on.

etanshaul:
@loosebazooka @patflynn and anyone else that's interested, please review. This change affects all tooling updated to use devappserverv1, allowing it to properly set environment variables configured in appengine-web.xml.

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
```diff
@@ -99,24 +105,88 @@ public boolean isJava8() {
...
+    Node environmentParentNode = getNode(document, "appengine-web-app", "env-variables");
+    return getAttributeMap(environmentParentNode, "env-var", "name", "value");
+  }
+
+  private static String getText(Node node) {
+    if (node != null) {
+      try {
+        return node.getTextContent();
+      } catch (DOMException ex) {
+        // this shouldn't happen barring a very funky DOM implementation
```
patflynn: shouldn't we throw runtime exception?

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
```diff
@@ -99,24 +105,88 @@ public boolean isJava8() {
...
+      }
+    }
+
+    return null;
+  }
+
+  /**
+   * @return a map formed from the attributes of the nodes contained within the parent node.
```
patflynn: I don't think this is google java format.  You have to have full @javadoc annotations for a method if you have any. If you want to ommit them completely just change this to:
```
/** 
 * Returns a map..
 **/ 
```

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
```diff
@@ -99,24 +105,88 @@ public boolean isJava8() {
...
+
+          if (nodeName.equals(child.getNodeName()) && attributeMap != null) {
+            Node keyNode = attributeMap.getNamedItem(keyAttributeName);
+
+            if (keyNode != null) {
+              Node valueNode = attributeMap.getNamedItem(valueAttributeName);
+              try {
+                nameValueAttributeMap.put(keyNode.getNodeValue(), valueNode.getNodeValue());
+              } catch (DOMException ex) {
+                // this shouldn't happen barring a very funky DOM implementation
```
patflynn: swallowing exceptions seems risky.

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
```diff
@@ -99,24 +105,88 @@ public boolean isJava8() {
...
-      return null;
+
+      return nameValueAttributeMap;
+    }
+
+    return null;
+  }
+
+  /**
+   * @return first node found matching the given name contained within the parent node.
```
patflynn: same comment as above

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
etanshaul: maybe. the question is, if the parsing fails, do we want to throw an exception, or fall back on default values (how it was working before)? I'm in favor of runtime exception.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -112,8 +115,11 @@ public void run(RunConfiguration config) throws AppEngineException {
     for (File service : config.getServices()) {
       arguments.add(service.toPath().toString());
     }
+
+    Map<String, String> appEngineEnvironment = getAllAppEngineEnvironment(config.getServices());
```
patflynn: would it makes sense to perform this logic in the `runDevAppServer1Command` implementation and not pass it as a parameter? It seems like we would always do it.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -194,4 +200,26 @@ boolean isJava8(List<File> services) {
     }
     return java8Detected;
   }
+
+  @VisibleForTesting
+  Map<String, String> getAllAppEngineEnvironment(List<File> services) {
```
patflynn: method name is a little weird.

`getAllAppEngineWebXmlEnvironmentVariables`?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -194,4 +200,26 @@ boolean isJava8(List<File> services) {
...
+    Map<String, String> allAppEngineEnvironment = Maps.newHashMap();
+    for (File serviceDirectory : services) {
+      Path appengineWebXml = serviceDirectory.toPath().resolve("WEB-INF/appengine-web.xml");
+      try (InputStream is = Files.newInputStream(appengineWebXml)) {
+        Map<String, String> appEngineEnvironment = AppEngineDescriptor.parse(is).getEnvironment();
+        if (appEngineEnvironment != null) {
+          if (!Collections.disjoint(
+              allAppEngineEnvironment.keySet(), appEngineEnvironment.keySet())) {
+            log.warning("Found duplicated environment keys between appengine-web.xml files.");
```
patflynn: we should print out which keys and in which services.

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
etanshaul: updated

src/test/resources/projects/Standard8ProjectEnvironmentVariables/WEB-INF/appengine-web.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
```
patflynn: it'd be nice to have a test that ran multiple services.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
etanshaul: done

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
etanshaul: I'm not sure about this. If you look at the isJava8 method in CloudSdkAppEngineDevServer1 its doing something similar (parsing the appengine-web.xml, and it does this every time).

I think of this preparation of the env vars from appengine-web.xml as similar to the preparation of the arguments to the command (which happens in this class).

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
etanshaul: Yeah this was on my radar too. I'll add it.

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
etanshaul: done

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
etanshaul: done

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
etanshaul: done

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
```diff
@@ -16,16 +16,23 @@
 
 package com.google.cloud.tools.appengine;
```
loosebazooka: was there ever a test for this file?

etanshaul: no, it was never directly tested. I'll if I can add something.

etanshaul:
@loosebazooka added unit tests for AppEngineDescriptor

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -273,7 +273,8 @@ public void runDevAppServerCommand(List<String> args)
    * @throws CloudSdkOutOfDateException when the installed Cloud SDK is too old 
    * @throws AppEngineException when dev appserver cannot be found
    */
-  public void runDevAppServer1Command(List<String> jvmArgs, List<String> args)
+  public void runDevAppServer1Command(List<String> jvmArgs, List<String> args,
```
elharo: would it make sense to overload this method so we don't break existing clients?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -194,4 +201,35 @@ boolean isJava8(List<File> services) {
...
+    }
+    return allAppEngineEnvironment;
+  }
+
+  private void checkAndWarnDuplicateEnvironmentVariables(Map<String, String> newEnvironment,
+                                                        Map<String, String> existingEnvironment,
+                                                        String service) {
+    for (String key : newEnvironment.keySet()) {
+      if (existingEnvironment.containsKey(key)) {
+        log.warning(String.format("Found duplicated environment variable key '%s' across "
```
elharo: duplicated --> duplicate

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
etanshaul: actually clients don't invoke it this way, they call CloudSdkAppEngineDevServer1#run. I'm not even sure the method is public, unless other clients that I'm unaware of invoke it this way? @loosebazooka

etanshaul: within the core lib, its only called by the unit test, so maybe we can make it package-private so its not leaked out in the API

chanseokoh: +1

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
etanshaul: fixed

etanshaul:
@elharo since you previously commented, are you good with this now?

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
```diff
@@ -99,24 +106,90 @@ public boolean isJava8() {
...
+   */
+  private static Map<String, String> getAttributeMap(Node parent,
+                                                     String nodeName,
+                                                     String keyAttributeName,
+                                                     String valueAttributeName) {
+    if (parent != null) {
+      Map<String, String> nameValueAttributeMap = Maps.newHashMap();
+
+      if (parent.hasChildNodes()) {
+        for (int i = 0; i < parent.getChildNodes().getLength(); ++i) {
```
elharo: i++ is more common

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
```diff
@@ -99,24 +106,90 @@ public boolean isJava8() {
...
+        throw new AppEngineException("Failed to parse text content from node "
+            + node.getNodeName());
+      }
+    }
+
+    return null;
+  }
+
+  /**
+   * Returns a map formed from the attributes of the nodes contained within the parent node.
```
elharo: what if two child elements have attributes with the same name?

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
etanshaul: yea, not sure why its like that, Ill update

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
etanshaul: then the last one found will be added to the map (according to the contract of Map#put)

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
briandealwis: There is: `src/test/java/com/google/cloud/tools/appengine/AppEngineDescriptorTest.java`

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -292,7 +293,6 @@ public void runDevAppServer1Command(List<String> jvmArgs, List<String> args)
 
     logCommand(command);
 
-    Map<String, String> environment = Maps.newHashMap();
     environment.put("JAVA_HOME", javaHomePath.toAbsolutePath().toString());
```
briandealwis: Shouldn't we be copying the provided environment and avoid side-effects?

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
etanshaul: woa how did I miss that. Thanks!

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
etanshaul: good point

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
loosebazooka: Kinda disappointing that we can't use the shortened for loop syntax here :disappointed:

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
etanshaul: fixed

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
etanshaul: updated

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
etanshaul: updated

etanshaul:
@elharo @briandealwis please take a look to see if I've addressed all your comments.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -194,4 +201,35 @@ boolean isJava8(List<File> services) {
...
+      }
+    }
+    return allAppEngineEnvironment;
+  }
+
+  private void checkAndWarnDuplicateEnvironmentVariables(Map<String, String> newEnvironment,
+                                                        Map<String, String> existingEnvironment,
+                                                        String service) {
+    for (String key : newEnvironment.keySet()) {
+      if (existingEnvironment.containsKey(key)) {
```
briandealwis: Downside to this detection approach is that it doesn't provide the source of the original definitions.  But this can be left as a todo for later.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -194,4 +201,35 @@ boolean isJava8(List<File> services) {
...
+
+  private Map<String, String> getAllAppEngineWebXmlEnvironmentVariables(List<File> services) {
+    Map<String, String> allAppEngineEnvironment = Maps.newHashMap();
+    for (File serviceDirectory : services) {
+      Path appengineWebXml = serviceDirectory.toPath().resolve("WEB-INF/appengine-web.xml");
+      try (InputStream is = Files.newInputStream(appengineWebXml)) {
+        Map<String, String> appEngineEnvironment = AppEngineDescriptor.parse(is).getEnvironment();
+        if (appEngineEnvironment != null) {
+          checkAndWarnDuplicateEnvironmentVariables(
+              appEngineEnvironment, allAppEngineEnvironment, serviceDirectory.getName());
```
briandealwis: Since you are parsing the `appengine-web.xml`, why not report the real service name rather than the directory?  The directory name may not correspond to the service name.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
etanshaul: yea, I was aware of this when implementing it. I didn't think it was that essential to justify the extra logic to deliver this (knowing one of the offending files to me was adequate). But yea, can be a todo.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
etanshaul: @briandealwis updated

etanshaul:
@briandealwis / @elharo do I have the ok? I'd love to merge this in today if you have no further comments.

patflynn:
@etanshaul 
@elharo 
@briandealwis 

What do you guys think of a log.info message tell the user we're setting these environment variables?

briandealwis:
+1: that would help the developer figure out where values are coming from.

etanshaul:
@patflynn, added a logging statement ca8678e69677ceb65a768d6923855087d5ffae6c

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -292,9 +293,14 @@ public void runDevAppServer1Command(List<String> jvmArgs, List<String> args)
 
     logCommand(command);
 
-    Map<String, String> environment = Maps.newHashMap();
-    environment.put("JAVA_HOME", javaHomePath.toAbsolutePath().toString());
-    processRunner.setEnvironment(environment);
+    Map<String, String> devServerEnvironment = Maps.newHashMap(environment);
+    if (!devServerEnvironment.isEmpty()) {
+      logger.info("Setting user supplied environment variables: "
```
patflynn: nit: I think you need to let them know this is coming from appengine-web.xml :)

etanshaul: done

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -292,9 +293,14 @@ public void runDevAppServer1Command(List<String> jvmArgs, List<String> args)
 
     logCommand(command);
 
-    Map<String, String> environment = Maps.newHashMap();
-    environment.put("JAVA_HOME", javaHomePath.toAbsolutePath().toString());
-    processRunner.setEnvironment(environment);
+    Map<String, String> devServerEnvironment = Maps.newHashMap(environment);
+    if (!devServerEnvironment.isEmpty()) {
+      logger.info("Setting appengine-web.xml configured environment variables: "
```
patflynn: OK now I am being a total PITA but I worry that some other entrypoint might pass environment that's not based off of appengine-web.xml. Perhaps this logging statement belongs closer to CloudSdkAppEngineDevServer1:120?

Feel free to throw darts at an effigy of me.

etanshaul: no, you're right. I was planning on improving this in the other PR (that allows client supplied env vars) since we'll have to differentiate between them. But yea, I think maybe moving them to the other class is the right approach.

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
```diff
@@ -99,24 +106,90 @@ public boolean isJava8() {
...
+    }
+
+    return null;
+  }
+
+  /**
+   * Returns the first node found matching the given name contained within the parent node.
+   */
+  private static Node getNode(Document doc, String parentNodeName, String targetNodeName) {
+    NodeList parentElements = doc.getElementsByTagNameNS(APP_ENGINE_NAMESPACE, parentNodeName);
```
elharo: I'm not sure why you're using the parentNodeName here. You could jump straight to the relevant nodes.

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
```diff
@@ -99,24 +106,90 @@ public boolean isJava8() {
...
+    if (parentElements.getLength() > 0) {
+      Node parent = parentElements.item(0);
+      if (parent.hasChildNodes()) {
+        for (int i = 0; i < parent.getChildNodes().getLength(); i++) {
+          Node child = parent.getChildNodes().item(i);
+          if (child.getNodeName().equals(targetNodeName)) {
```
elharo: Theoretically you should be checking namesapces here too. Probalby doesn't matter much in practice. However there are a lot of ways this navigtion could go wrong is the appen gine-web.xml file is not as required.

src/test/resources/projects/Standard8Project1EnvironmentVariables/WEB-INF/appengine-web.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
```
elharo: You can make these test data file smaller if all they're used for is to test the environment variables.

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
etanshaul: I figured this helps to ensure the xml for the env vars is properly formed: e.g. `env-var` elements contained within a parent `env-variables`. If I skipped straight to `env-vars` then wouldn't it match against invalid parent elements?

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
etanshaul: could you be more specific regarding your last sentence? e.g. what could go wrong and do you think I should be handling it here?

src/test/resources/projects/Standard8Project1EnvironmentVariables/WEB-INF/appengine-web.xml
etanshaul: done

src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
etanshaul: Also, this method is used by all the extractors (service, version etc. Not just for environment). Its pretty much a direct refactor of what was there before (`getTopLevelValue` as it was previously called).

etanshaul:
discussed with @elharo offline and we are good to go with this. will merge

Issue #377, opened 2017-04-17
Set environment variables from appengine-web.xml for local run


etanshaul:
see https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/1364 for some background.

The environment variables defined in appengine-web.xml are not being set on the process running devappserver1. 

One possible solution could be to parse appengine-web.xml for `<env-variables>` and set them when doing devappserver1 runs

patflynn:
also https://github.com/GoogleCloudPlatform/app-maven-plugin/issues/161

PR #381, opened 2017-04-21
Allow clients to pass in environment for devappserver runs

### LINKED ISSUE #379: Allow clients to pass in arbitrary environment variables
See https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/1364 for more context

As an example, IntelliJ has a UI for configuring environment variables to be passed into the local dev server run. We'd like these to be honored and supplied to devappserver.

Also, there are certain configs (like google authentication path for using client api's) that need to be set as environment variables (and is documented as such) that currently require the user to set them manually in their environment.

etanshaul:
fixes #379 

Note the this depends on #378 so the target branch is the branch of that PR. I'll update the target branch once #378 is merged in.

I added the environment just to the config of the devappserver. I didn't want to add it to the CloudSdk configuration (as its possible that the object is reused for multiple commands and we don't necessarily want to share these env variables across commands). Also, this is closing a gap we have with the devappserver so I wanted to limit it to that.


etanshaul:
For consistency, I think I should pass these env vars into DevAppServer2 runs as well. Will update.

etanshaul:
@patflynn (and whoever else has time) ptal. #377 should be approved first since this depends on that.

etanshaul:
@elharo since you previously commented, are you good with this now?

src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
```diff
@@ -48,6 +49,7 @@
   private String defaultGcsBucketName;
   private Boolean clearDatastore;
   private File datastorePath;
+  private Map<String, String> environment;
```
loosebazooka: I'm not sure how maven handles maps as inputs. I need to check it out, but how does the command line handle this for dev_appserver2? which actually has an --env_var parameter.

etanshaul: my python is rusty, but it looks like it just joins it by a ',', so like this I think:
`--env_var=key1,val2,key2, val2...`

etanshaul: yea after discussing with @patflynn I realized what you were getting at. I updated devappserver1 to set these env vars onto the running process. However for devappserver2 it needs to be updated to split them up and passed into the command for --env_var. I'll update.

etanshaul: @loosebazooka what are your thoughts, for devappserver2, on what env vars should be set on the process, or passed in as args to --env_var?

We currently set a couple vars on the process itself (JAVA_HOME, CLOUDSDK_CORE_DISABLE_PROMPT). Do you think it makes sense to set all client supplied env vars as --env_var args instead? I'm not sure if there is a difference (and if not, why not just set them all on the process).

loosebazooka: I don't actually know the behavior, but I believe the env_var flag was specifically to do something that would allow them to be readable by the servers. Perhaps @davidjn might know though.

etanshaul: just to correct myself, the format for env_var is actually as follows:

`--evn_var key1=val1 --env_var key2=val2 ...`

etanshaul: @patflynn, @loosebazooka - I updated this:

- For devappserver1, the client supplied env vars are set on the running process
- For devappserver2, the client supplied env vars are passed into the --env_var argument in the format of my previous comment

please take a look

etanshaul:
updated target branch to master since #378 has now been merged.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/DevAppServerArgs.java
```diff
@@ -71,4 +72,12 @@
     }
     return Collections.emptyList();
   }
+
+  /**
+   * @return {@code [--name, key1=val1, --name, key2=val2, ...]} or {@code []} if
```
patflynn: You can have a method comment and omit the @return, but you can't do the converse. 
https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment

src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
```diff
@@ -264,4 +266,13 @@ public File getDatastorePath() {
...
     this.datastorePath = datastorePath;
   }
+
+  @Override
+  public Map<String, String> getEnvironment() {
+    return environment;
+  }
+
+  public void setEnvironment(Map<String, String> environment) {
+    this.environment = environment;
```
patflynn: you could conceivably do the defensive copy to an immutable map here and then you don't have to worry about it elsewhere.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer.java
```diff
@@ -30,6 +30,7 @@
 import java.net.URL;
 import java.nio.file.InvalidPathException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 
 /**
```
patflynn: it would be nice to rename this class to CloudSdkAppEngineDevServer2 in a subsequent PR

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/Args.java
```diff
@@ -126,13 +125,27 @@
    * @return [key1=value1,key2=value2,...], [] if keyValueMapping=empty/null
    */
   public static List<String> keyValues(Map<?, ?> keyValueMapping) {
+    return namedKeyValues("", keyValueMapping);
+  }
+
+  /**
+   * Produces a named key/value pair list given a name and a {@link Map}.
+   *
+   * @return [--name, key1=value1, --name, key2=value2,...], [] if keyValueMapping=empty/null
```
patflynn: @code

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/Args.java
```diff
@@ -126,13 +125,27 @@
    * @return [key1=value1,key2=value2,...], [] if keyValueMapping=empty/null
```
patflynn: It'd be a tad clearer with "" around the array entries.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/Args.java
```diff
@@ -126,13 +125,27 @@
...
+    return namedKeyValues("", keyValueMapping);
+  }
+
+  /**
+   * Produces a named key/value pair list given a name and a {@link Map}.
+   *
+   * @return [--name, key1=value1, --name, key2=value2,...], [] if keyValueMapping=empty/null
+   */
+  public static List<String> namedKeyValues(String name, Map<?, ?> keyValueMapping) {
     List<String> result = Lists.newArrayList();
```
patflynn: could we invert this logic and have namedKeyValues call keyValues and zip in the name?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/DevAppServerArgs.java
```diff
@@ -71,4 +72,12 @@
...
+   *      keyValues=empty/null.
+   */
+  public static List<String> get(String name, Map<String, String> keyValues) {
+    return Args.namedKeyValues(name, keyValues);
```
patflynn: actually I'd put the zip logic right here.

etanshaul: I was thinking the Args class is the right place for this. I know its just a pass through for this class, but this seems to be the pattern we are following (plus its possible that a gcloud arg may need this in the future).

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/Args.java
etanshaul: maybe this comment is obsolete given your other one, but if we invert this then the zipping would require an extra iteration over the list that has already been formed. This way, its done 'in place' during the iteration.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/DevAppServerArgs.java
etanshaul: yeah. unfortunately we are doing this everywhere else in this class too. I don't want to fix those and leak stuff into this PR, but I guess I could kick off using the right format here (instead of continuing the incorrect approach).

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/DevAppServerArgs.java
patflynn: fair point. I'm wondering if namedKeyValues is too generic. At first read I think 'what's the difference between a key and a name?'. What do you think of something like flagPerEntry(flagName, map)

src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
etanshaul: done. I also did this for the jvmArgs field (a List), for consistency. I don't think this allows to get rid of any other code though (since clients could presumable invoke CloudSdk#runDevAppServer1Command directly, in which case we still need a defensive copy here).

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/DevAppServerArgs.java
etanshaul: screw it, I'll just fix all the other @return statements too in this PR. Its not a lot of changes.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/DevAppServerArgs.java
etanshaul: `flagPerEntry` isn't bad; but what would you suggest for the sister method that returns the entry without the flag? `namedKeyValues` is a bit confusing I agree, but when paired with `keyValues` it because a little clearer.

patflynn:
Aha, the penny has dropped a little further.  This GcloudArgs library is
inconsistent in who is reponsible for including the '--flagName', with the
map support leaving it to the client to prepend.

How about just 'flaggedKeyValues", and relying on the javadoc to resolve
confusion?

I'm flumoxed by the keyValues implementation. It seems to always return a
single entry list.. Why doesn't it just return string?  The lack of "" in
the javadoc return comment compounds the confusion.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/Args.java
patflynn: I'm really not a fan of having namedKeyValues method called from the non-named use-case and branching on null/empty name. 

I've just seen that zip is not in jdk8 (learn something new every day).  This looks like a tidy use-case for streams and mapping, but the weirdness of keyValues stymies elegance. what do you tihnk of just ignoring keyValues to its confusion and implement flaggedKeyValues with something as simple as: 
```
keyValues().entrySet().stream()
        .map(e -> "--flag " + e.getKey() + '=' + e.getValue())
        .collect(Collectors.joining(" , "));
```

is that crazy?

etanshaul: aren't we stuck with java7 though for this lib? even if so we can use guava to do stuff like this, although it will be much uglier.

src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
```diff
@@ -172,7 +177,7 @@ public void setPythonStartupArgs(String pythonStartupArgs) {
   }
 
   public void setJvmFlags(List<String> jvmFlags) {
-    this.jvmFlags = jvmFlags;
+    this.jvmFlags = jvmFlags != null ? ImmutableList.copyOf(jvmFlags) : null;
```
loosebazooka: I wonder if we should do safer copying of lists everywhere?

It's a good idea, but is the expectation that a user would change their environment after setting this?

src/test/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineServicesTest.java
```diff
@@ -61,7 +61,7 @@ public void setTrafficTest() throws ProcessRunnerException {
     appEngineService.setTraffic(configuration);
 
     List<String> args =
-        Arrays.asList("services", "set-traffic", "myService", "--splits", "v1=0.3,v2=0.7",
+        Arrays.asList("services", "set-traffic", "myService", "--splits", "v1=0.3", "v2=0.7",
```
loosebazooka: I dont understand how this would work.

loosebazooka: does splits accept multiple parameters after itself?

src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
etanshaul: For the first question - yeah probably. I only touched this field unrelated to this PR because it's in same class. We can deal with the rest of the cases in a different PR.

No I don't think it's expected that they change it. It's just to protect against them unknowingly (or knowingly even but for other reasons) changing it and getting unexpected side effects. Maybe I misunderstood your point.

src/test/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineServicesTest.java
etanshaul: Yeah I think so. Or did at least when I first implemented this. I'll check the current state.

etanshaul: ohh. good eye! I think my change may have subtly broke this. It accepts the following format:

--splits v2=.5,v1=.5

With my update it will be space separated. I'll revisit this.

etanshaul:
@loosebazooka fixed: cb0ecc0

@patflynn turns out you were right to be suspicious that the updates to the keyValue method were client-breaking.

Thanks for catching.

etanshaul:
@patflynn & @loosebazooka, I know you approved already but do you mind taking a quick last look given the last change I had to make (see the last couple of comments).

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/Args.java
```diff
@@ -121,11 +121,11 @@
...
 
   /**
-   * Produces a key/value pair list from a {@link Map}.
+   * Produces a single element list with a key/value pair comma separated string from a {@link Map}.
    *
-   * @return [key1=value1,key2=value2,...], [] if keyValueMapping=empty/null
+   * @return {@code ["key1=value1,key2=value2,..."]} or {@code []} if keyValueMapping=empty/null
    */
-  public static List<String> keyValues(Map<?, ?> keyValueMapping) {
+  static List<String> keyValueString(Map<?, ?> keyValueMapping) {
```
patflynn: can you safely make this non public?

etanshaul: what do you mean? There were no non-package private usages of it.

etanshaul: also the class itself is package private

patflynn: ok

loosebazooka:
lgtm :surfing_man:

PR #383, opened 2017-04-26
Update to 0.3.3-SNAPSHOT

etanshaul:


patflynn:
what happened to 0.3.2?

etanshaul:
Shouldn't the snapshot be one version up from the release? E.g we just
released 0.3.2 so the snapshot should be 0.3.3. Previously the snapshot was
set to the same version as the release which is why I had to skip one.

PR #387, opened 2017-04-29
distinguish between I/O errors and XML parsing errors in appengine-we…

elharo:
@briandealwis @chanseokoh Eclipse needs to distinguish I/O errors (log) and XML parsing errors (don't log)

see https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1885


elharo:
My gut says no. The specific failure mode here is a syntax level XML parsing failure, not a semantic failure such as an unrecognized runtime. `SAXException` is a pretty standard means of signaling this, and is part of the JRE core library, so there's no extra dependency here.

PR #392, opened 2017-05-03
Support staging optional config files for flex

### LINKED ISSUE #391: Add support for staging optional configuration files along with `app.yaml` in flex staging
This is the opposite of https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/347. The current flex staging copies only `app.yaml`, but it'd be nice to have an additional option to include other config files when staging flex.

I know we have separate goals for deploying individual config files, but until the Eclipse plugin comes up with a working UI to enable that, we need a temporary support for copying all config files along with `app.yaml`. (https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1895)

chanseokoh:
Fixes #391.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/api/deploy/DefaultStageFlexibleConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/appengine/api/deploy/DefaultStageFlexibleConfiguration.java
@@ -24,6 +24,7 @@ import java.io.File;
 public class DefaultStageFlexibleConfiguration implements StageFlexibleConfiguration {
 
   private File appEngineDirectory;
+  private boolean includeOptionalConfigurationFiles;
   private File dockerDirectory;
   private File artifact;
   private File stagingDirectory;
@@ -35,6 +36,15 @@ public class DefaultStageFlexibleConfiguration implements StageFlexibleConfigura
 
   public void setAppEngineDirectory(File appEngineDirectory) {
     this.appEngineDirectory = appEngineDirectory;
+  }
+
+  @Override
+  public boolean getIncludeOptionalConfigurationFiles() {
+    return includeOptionalConfigurationFiles;
+  }
+
+  public void setIncludeOptionalConfigurationFiles(boolean includeOptionalConfigurationFiles) {
+    this.includeOptionalConfigurationFiles = includeOptionalConfigurationFiles;
   }
 
   @Override
--- a/src/main/java/com/google/cloud/tools/appengine/api/deploy/StageFlexibleConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/appengine/api/deploy/StageFlexibleConfiguration.java
@@ -24,9 +24,17 @@ import java.io.File;
 public interface StageFlexibleConfiguration {
 
   /**
-   * Directory containing {@code app.yaml}.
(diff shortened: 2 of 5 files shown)


src/main/java/com/google/cloud/tools/appengine/api/deploy/StageFlexibleConfiguration.java
```diff
@@ -39,8 +47,8 @@
   File getArtifact();
 
   /**
-   * Directory where {@code app.yaml}, files in docker directory, and the artifact to deploy will
-   * be copied for deploying.
+   * Directory where {@code app.yaml}, optional configuration files, files in docker directory, and
```
elharo: is "files in docker directory" just the docker directory itself?

src/test/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineFlexibleStagingTest.java
```diff
@@ -198,8 +202,98 @@ public void testCopyAppEngineContext_appYamlInAppEngineDirectory() throws Except
...
+  public void testCopyAppEngineContext_includeQueueYaml() throws Exception {
+    verifyOptionalConfigurationFileStaged("queue.yaml", true /* includeSet */);
+  }
+
+  @Test
+  public void testCopyAppEngineContext_doNotIncludeQueueYaml() throws Exception {
+    verifyOptionalConfigurationFileStaged("queue.yaml", false /* includeSet */);
+  }
+
+  private void verifyOptionalConfigurationFileStaged(String filename, boolean includeSet)
```
elharo: I did not initially understand the variable name "includeSet". I thought this referred to a java.util.Set or some other kind of set. Maybe just `include`?

src/main/java/com/google/cloud/tools/appengine/api/deploy/DefaultStageFlexibleConfiguration.java
```diff
@@ -38,6 +39,15 @@ public void setAppEngineDirectory(File appEngineDirectory) {
   }
 
   @Override
+  public boolean getIncludeOptionalConfigurationFiles() {
+    return includeOptionalConfigurationFiles;
+  }
+
+  public void setIncludeOptionalConfigurationFiles(boolean includeOptionalConfigurationFiles) {
```
elharo: It's a bit strange to think of this as a boolean. Also maybe not too extensible. Would it make sense to ask clients to provide the list of files they want to include?

patflynn: @elharo that's already supported through deployables, which is how Intellij supports this for standard.   
@chanseokoh I'm not convinced we need this at all, but additionally we shouldn't be staging these files anywhere. Staging is only required for applications.

On a tangent: I think the 'optional configuration files' terminology used here and in the Eclipse UI is a bit poor.

chanseokoh:
@patflynn Gosh, what I have been thinking? I recently made the Eclipse plugin list the configuration files as deployables, so staging these files is unnecessary. I think I was briefly confused.

patflynn:
:).  It's a very easy mistake to make.  In fact we used to stage those files automatically on every standard build.

PR #399, opened 2017-05-05
Close file with try-with-resources / use Files.newInputStream() instead of FileReader()

### LINKED ISSUE #398: AppYaml parsing app.yaml does not close open file descriptor
https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/v0.3.2/src/main/java/com/google/cloud/tools/project/AppYaml.java#L49

This might be the reason that Eclipse test builds on Windows fail a lot: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1896

The SnakeYAML 1.17 Javadoc does not say it will close the reader: http://www.atetric.com/atetric/javadoc/org.yaml/snakeyaml/1.17/org/yaml/snakeyaml/Yaml.html#load-java.io.Reader-

chanseokoh:
Fixes #398.

### DIFF
--- a/src/main/java/com/google/cloud/tools/project/AppYaml.java
+++ b/src/main/java/com/google/cloud/tools/project/AppYaml.java
@@ -21,7 +21,9 @@ import org.yaml.snakeyaml.Yaml;
 import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.Reader;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/cloud/tools/project/AppYaml.java
```diff
@@ -33,18 +35,13 @@
   private static final String RUNTIME_KEY = "runtime";
 
   /**
-   * @param appyaml A valid existing app.yaml
-   * @throws FileNotFoundException if app.yaml doesn't exist
+   * @param appYaml A valid existing app.yaml
```
elharo: the app.yaml file

src/main/java/com/google/cloud/tools/project/AppYaml.java
```diff
@@ -33,18 +35,13 @@
...
-    Yaml yaml = new Yaml();
-
-    // Standard snake yaml parsing.
-    @SuppressWarnings("unchecked")
-    Map<String, Object> parserResult =
-        (Map<String, Object>) yaml.load(new FileReader(appyaml.toFile()));
-
-    yamlMap = parserResult;
+  public AppYaml(Path appYaml) throws IOException {
+    try (Reader reader = new FileReader(appYaml.toFile())) {
```
elharo: FileReader is discouraged because it uses the platform default encoding. Hook an InputStreamReader with a specified encoding to a FileInputStream instead.

elharo: Actually, we should use an InputStream here, not a reader at all:

https://stackoverflow.com/questions/488694/how-to-set-the-character-encoding-in-a-yaml-file

chanseokoh: That's a good point.

PR #402, opened 2017-05-08
better Java 8 checking

elharo:
@chanseokoh @ludoch


src/main/java/com/google/cloud/tools/appengine/AppEngineDescriptor.java
```diff
@@ -100,11 +100,11 @@ public String getServiceId() {
   }
 
   /**
-   * @return true if the runtime specified by the user is Java8.
+   * @return true if the runtime read from appengine-web.xml is Java8
    */
   public boolean isJava8() {
     String runtime = getRuntime();
-    return runtime != null && runtime.startsWith("java8");
+    return "java8".equals(runtime) || "java8g".equals(runtime);
```
briandealwis: Would it be useful to expose this as a constant/enum?  Especially since there seem to be other valid runtimes (`python27`, etc.)

elharo: There are no other currently valid runtimes that use appengine-web.xml (instead of app.yaml) though. And so far we don't seem to need that.

briandealwis: Good point.  The other implication here is that CT4E will turn a change to `java8h` (so `!isJava8()`) into downgrading the project to Java 7.  But I'd guess these `java8[a-z]` runtime identifiers are internal and won't be seen by developers in practice?

elharo: There are no other identifiers currently. There were never java8[a-f]

briandealwis: You had a test for `java8h`, so I wondered.  LGTM.

PR #404, opened 2017-05-09
Update AppYaml constructor Javadoc

chanseokoh:


chanseokoh:
I see the `kokoro-macos` build has a configuration issue. Please disregard that.

chanseokoh:
Turns out that it can also throw `ParserException` in case of malformed YAML.

PR #407, opened 2017-05-09
Fix NPE in getRuntime() when empty app.yaml

### LINKED ISSUE #405: NullPointerException from AppYaml.getRuntime()
`yamlMap` can be null [here](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/v0.3.3/src/main/java/com/google/cloud/tools/project/AppYaml.java). The problem is that `Yaml.load()` may return null [in the constructor](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/v0.3.3/src/main/java/com/google/cloud/tools/project/AppYaml.java#L45), when the `app.yaml` is completely empty for example. [The SnakeYAML Javadoc](http://javadox.com/org.yaml/snakeyaml/1.8/org/yaml/snakeyaml/Yaml.html#load(java.io.InputStream)) was not friendly enough to mention it can return null.

chanseokoh:
Fixes #405.

### DIFF
--- a/src/test/java/com/google/cloud/tools/project/AppYamlTest.java
+++ b/src/test/java/com/google/cloud/tools/project/AppYamlTest.java
@@ -23,6 +23,7 @@ import org.junit.rules.TemporaryFolder;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/cloud/tools/project/AppYamlTest.java
```diff
@@ -52,9 +52,16 @@ public void testGetRuntime_failureBecauseNotPresent() throws IOException {
     Assert.assertNull(new AppYaml(appYaml).getRuntime());
   }
 
+  // https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/405
+  @Test
+  public void testGetRuntime_noNullPointerExceptionIfEmptyAppYaml() throws IOException {
```
elharo: this name is a bit long. Maybe just
testGetRuntime_emptyAppYaml

PR #408, opened 2017-05-09
Wrap malformed app.yaml parsing exception into AppEngineException

### LINKED ISSUE #406: CloudSdkAppEngineFlexibleStaging.stageFlexible() fails if app.yaml is malformed
[Parsing malformed app.yaml here](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/v0.3.3/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineFlexibleStaging.java#L82) using SnakeYAML will throw `ScannerException` and `ParserException` that are `RuntimeException`, as noted [here](https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/404/files).

I think the staging logic should just log a warning and continue staging rather than exploding, very much like it currently does with `IOException` on `app.yaml`.

chanseokoh:
Fixes #406.


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
```diff
@@ -73,16 +75,17 @@ public void stageFlexible(StageFlexibleConfiguration config) throws AppEngineExc
...
     // verification for app.yaml that contains runtime:java
     Path appYaml = config.getAppEngineDirectory().toPath().resolve(APP_YAML);
     String runtime = null;
     try {
       if (Files.isRegularFile(appYaml)) {
         runtime = new AppYaml(appYaml).getRuntime();
       }
-    } catch (IOException e) {
-      log.warning("Unable to determine runtime: error parsing app.yaml : " + e.getMessage());
+    } catch (IOException | ScannerException | ParserException ex) {
```
patflynn: I can imagine an IDE scenario where we would want to know details why we couldn't find a runtime.

chanseokoh: Note that `findRuntime()` is an internal method. From the perspective of clients calling `stageFlexible()`, I think it makes sense for `stageFlexible()` to just stage `app.yaml` regardless of the parsing failure.

If we do want the method to propagate the parsing failure, I think we will want to make `stageFlexible()` throw `AppEngineException` as other usual public methods do, but I don't really feel that's needed. Thoughts?

patflynn: I see.  It think it makes more sense to me for this method to throw  a checked InvalidAppYamlException that wraps the cause and lets the calling staging code log it and carry on.

chanseokoh: You mean let `findRuntime()` throw these exceptions (for that matter, I think we don't really need a new `InvalidAppYamlException`) and let `copyAppEngineContext()` (which is the caller) to swallow it, right? Or, do you propagate such an exception to clients of appengine-plugins-core, meaning staging failure?

elharo: Why should staging proceed at all here? If app.yaml is malformed, I think we'd want to cancel the operation and alert the user of the problem in their code.

chanseokoh:
I changed the code to wrap `ScannerException` and `ParserException` into `AppEngineException` (all of which are `RuntimeException`) and put a proper error message. I also removed catching `IOException` in `findRuntime()`, as it will be handled (and be re-thrown) by the outer caller.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
```diff
@@ -73,33 +73,30 @@ public void stageFlexible(StageFlexibleConfiguration config) throws AppEngineExc
...
-    return runtime;
   }
 
   @VisibleForTesting
   static void copyDockerContext(StageFlexibleConfiguration config, CopyService copyService,
       String runtime) throws IOException {
     if (config.getDockerDirectory() != null && config.getDockerDirectory().exists()) {
-      if (runtime != null && runtime.equals("java")) {
-        log.warning("WARNING: 'runtime 'java' detected, any docker configuration in "
+      if ("java".equals(runtime)) {
```
elharo: FYI, "java" means Java 6 if this is the same as in appengine-web.xml. Is this really what we want to check for here?

chanseokoh: It is different from appengine-web.xml. It is always "java" for Java. You can check this in the official app.yaml reference doc.

patflynn: You may want to check with @aslo

aslo: For flex, in app.yaml I believe @chanseokoh is right that "java" is currently the only accepted value. However, that will likely change in the near future (details are still being worked out).

PR #411, opened 2017-05-11
Set working directory for dev_appserver1

### LINKED ISSUE #385: Working directory for spawned devappservers
http://stackoverflow.com/a/43639428/1701388

We're not setting the working directory when launching the devappserver.

chanseokoh:
Fixes #385. For dev_appserver1, I think the best we can do for a multi-service run is to set the directory working to be the root of a representative (i.e., "default") service.

Potentially appengine-plugins-core could be extended to customize a working directory, but this will cover most use-cases.


patflynn:
@ludoch any thoughts on what the working directory should be for multi-service scenarios?
How does the working directory get used by App Engine Standard apps?

patflynn:
@chanseokoh I'm wonder if step 1, should just be to allow clients to set the working directory, and defaults to the service root path, if there is just one service and working directory is not specified.

 For our new Gradle and Maven plugins the working directory should potentially be the module in which the run configuration is configured, and only the tool would know that.

chanseokoh:
@patflynn No problem. I anticipate that the step 1 you mentioned will be a fairly trivial change, so I can do it in this PR.

chanseokoh:
The last commit is the "step 1", i.e., fallbacking conservatively: setting the working directory to the root of an app only if clients did not explicitly set a working directory and there is only one project.

I did not touch dev_appserver2, as it is not clear if there exists an issue when not setting a working directory at the current state.

chanseokoh:
> For our new Gradle and Maven plugins the working directory should potentially be the module in which the run configuration is configured, and only the tool would know that.

Not sure if I understood the "module" here, but if it refers to a directory where a "pom.xml" exists (for Maven), this directory shouldn't be the working directory. It should be the root of the servlet root context (where for example `index.html` to be served resides), so that [things like this](http://stackoverflow.com/q/43618900/1701388) will work.

patflynn:
yes. What I meant is that it should be the exploded-app directory for that
module.  However, I suspect that for multi-module deployments using the
working directory can never be made to work sensibly on devappserverv1.

patflynn:
@ludoch do you have any thoughts on how working directory should work?

src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
```diff
@@ -278,4 +278,13 @@ public void setDatastorePath(File datastorePath) {
...
     this.environment = environment != null ? ImmutableMap.copyOf(environment) : null;
   }
+
+  @Override
+  public File getWorkingDirectory() {
+    return workingDirectory;
+  }
+
+  public void setWorkingDirectory(File workingDirectory) {
+    this.workingDirectory = workingDirectory;
```
elharo: perhaps throw an exception if this isn't a directory?
should we use Path instead of File here? I don't know.

src/main/java/com/google/cloud/tools/appengine/api/devserver/RunConfiguration.java
```diff
@@ -81,4 +81,9 @@
   File getDatastorePath();
 
   Map<String, String> getEnvironment();
+
+  /**
+   * Valid only for dev_appserver1.
+   */
+  File getWorkingDirectory();
```
elharo: Hmm, guess we're already using File everywhere. Probably best to stay consistent.

src/main/java/com/google/cloud/tools/appengine/api/devserver/RunConfiguration.java
```diff
@@ -81,4 +81,9 @@
   File getDatastorePath();
 
   Map<String, String> getEnvironment();
+
+  /**
+   * Valid only for dev_appserver1.
```
elharo: This suggests an abstraction failure. It really shouldn't care 1 or 2. That's why the interface exists. And what does it do for dev_appserver2? Throw an UnsupportedOperationException?

Also, nees JavaDoc

src/main/java/com/google/cloud/tools/appengine/api/devserver/DefaultRunConfiguration.java
chanseokoh: This class doesn't do any validation by its design (a simple bean), so if we were to check it, we should do it in other places. But then, `java.lang.Process` will ultimately throw `IOException`, so I'm not inclined to add an extra check.

src/main/java/com/google/cloud/tools/appengine/api/devserver/RunConfiguration.java
chanseokoh: Makes sense. I'll make dev_appserver2 throw UnsupportedOperationException.

patflynn:
Let's keep @loosebazooka on core library PRs.

chanseokoh:
Had a chat with @patflynn and @loosebazooka, and we agreed we don't want/need to expose `setWorkingDirectory()`. I'll update the PR.

lesv:
Just found the email for this - not sure that understand @patflynn comment about multi-module deployments on DevAppServer1, but it would be nice to debug mutli-module things locally using DevAppServer1.

loosebazooka:
Multi modules will work and debug locally.

This is about a weird message that pops up when the working directory of
the dev_appserver is not the root of the web-application:

WARNING: Your working directory, (/path/to/my/project/src/root) is not
equal to your
web application root
(/path/to/my/project/src/root/target/exploded-<project>)
You will not be able to access files from your working directory on the
production server.

PR #417, opened 2017-05-14
small improvement to exceptions

elharo:
@loosebazooka nest original exception

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -662,6 +662,8 @@ public class CloudSdk {
     /**
      * Create a new instance of {@link CloudSdk}.
      * If {@code sdkPath} is not set, this method looks for the SDK in known install locations.
+     * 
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -723,6 +725,8 @@ private Path discoverSdkPath() {
           // prevent interference from exceptions in other resolvers
           logger.log(Level.SEVERE, resolver.getClass().getName()
               + ": exception thrown when searching for Google Cloud SDK", ex);
+          throw new CloudSdkNotFoundException(
```
chanseokoh: This breaks the outer `for` loop. I can we can recover from individual resolver exceptions and continue trying other resolvers. Also it doesn't seem right to throw the "not-found" exception when one of the resolvers just exploded.

loosebazooka: yeah we probably want to make the comment more clear above the logger to indicate more clearly that we don't want to fail hard here.

PR #420, opened 2017-05-15
better exception JavaDoc

### LINKED ISSUE #279: Update Javadoc about CloudSdkNotFoundException and CloudSdkOutOfDateException

elharo:
@chanseokoh fix #279


patflynn:
Make sure to keep @loosebazooka and core library PRs.

PR #421, opened 2017-05-15
fail earlier

### LINKED ISSUE #313: CloudSdk should fail earlier with inheritProcessOutput
In CloudSdk build we have

    public CloudSdk build() {

      // Default SDK path
      if (sdkPath == null) {
        sdkPath = discoverSdkPath();
      }

      // Verify there aren't listeners if subprocess inherits output.
      // If output is inherited, then listeners won't receive anything.
      if (inheritProcessOutput
          && (stdOutLineListeners.size() > 0 || stdErrLineListeners.size() > 0)) {
        throw new AppEngineException(
            "You cannot specify subprocess output inheritance and output listeners.");
      }

This is the wrong place to throw this exception. It should be an IllegalArgument or perhaps IllegalStateException, and it should be thrown as soon as a line listener is added when inheritProcessOutput is true. No reason to wait until build is called.

elharo:
@chanseokoh fix #313

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -597,6 +597,12 @@ public class CloudSdk {
      * parent process.
      */
     public Builder addStdOutLineListener(ProcessOutputLineListener stdOutLineListener) {
+      // Verify there aren't listeners if subprocess inherits output.
+      // If output is inherited, then listeners won't receive anything.
+      if (inheritProcessOutput) {
+        throw new IllegalStateException(
+            "You cannot specify subprocess output inheritance and output listeners.");
+      }
       this.stdOutLineListeners.add(stdOutLineListener);
       return this;
     }
@@ -606,6 +612,12 @@ public class CloudSdk {
      * process.
      */
     public Builder addStdErrLineListener(ProcessOutputLineListener stdErrLineListener) {
+      // Verify there aren't listeners if subprocess inherits output.
+      // If output is inherited, then listeners won't receive anything.
+      if (inheritProcessOutput) {
+        throw new IllegalStateException(
+            "You cannot specify subprocess output inheritance and output listeners.");
+      }
       this.stdErrLineListeners.add(stdErrLineListener);
       return this;
     }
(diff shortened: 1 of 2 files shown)


patflynn:
@loosebazooka owns the core library, so make sure you include him on PRs as assignee or reviewers.

src/test/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkTest.java
```diff
@@ -171,14 +170,31 @@ public void testNewCloudSdk_errListener() {
     assertEquals(1, builder.getExitListeners().size());
   }
 
-  @Test(expected = AppEngineException.class)
   public void testNewCloudSdk_inheritOutputAndOutListener() {
-    builder.inheritProcessOutput(true).addStdOutLineListener(outputListener).build();
+    try {
+      builder.inheritProcessOutput(true).addStdOutLineListener(outputListener).build();
```
elharo: done

src/test/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkTest.java
```diff
@@ -171,14 +170,31 @@ public void testNewCloudSdk_errListener() {
...
+    } catch (IllegalStateException ex) {
+      assertNotNull(ex.getMessage());
+    }
   }
 
-  @Test(expected = AppEngineException.class)
   public void testNewCloudSdk_inheritOutputAndErrListener() {
-    builder.inheritProcessOutput(true).addStdErrLineListener(outputListener).build();
+    try {
+      builder.inheritProcessOutput(true).addStdErrLineListener(outputListener).build();
```
elharo: done

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -647,6 +659,13 @@ public Builder runDevAppServerWait(int runDevAppServerWaitSeconds) {
      * @param inheritProcessOutput if true, stdout and stderr are redirected to the parent process
      */
     public Builder inheritProcessOutput(boolean inheritProcessOutput) {
+      // Verify there aren't listeners if subprocess inherits output.
+      // If output is inherited, then listeners won't receive anything.
+      if (inheritProcessOutput
+          && (stdOutLineListeners.size() > 0 || stdErrLineListeners.size() > 0)) {
```
elharo: what we really want is isNotEmpty()
With the extra ! this starts to get confusing

loosebazooka: Would `!= 0` be more obvious in its intent than `> 0`?

PR #426, opened 2017-06-21
Adds CloudSdk parameter for setting structured logs value

### LINKED ISSUE #425: Support gcloud structured logging
Release 159 added support for structured output logging. 

The library should add support so that clients can consume and parse these as needed.

etanshaul:
fixes #425

This new parameter mimics the new gcloud configuration for providing the structured error logs environment variable setting `CLOUDSDK_CORE_SHOW_STRUCTURED_LOGS`.

gcloud accepts a string value for this: `terminal`, `log`, `always`, `never`

Setting this to terminal, or always, produces json output for messages coming from the cloudsdk according to the verbosity value currently set. 

**Example (consumed from IntelliJ):**

Without structured logs set (defaulted to `never`):
image

Same error, with verbosity  defaulted to `error`, and the structure logs parameter set to `log`:
image

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -84,7 +84,8 @@ public class CloudSdk {
   private final String appCommandShowStructuredLogs;
   private final WaitingProcessOutputLineListener runDevAppServerWaitListener;
 
-  private CloudSdk(Path sdkPath, Path javaHomePath,
+  private CloudSdk(Path sdkPath,
+                   @Nullable Path javaHomePath,
                    @Nullable String appCommandMetricsEnvironment,
                    @Nullable String appCommandMetricsEnvironmentVersion,
                    @Nullable File appCommandCredentialFile,
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -181,7 +181,7 @@ public class CloudSdk {
       environment.put("CLOUDSDK_METRICS_ENVIRONMENT_VERSION", appCommandMetricsEnvironmentVersion);
     }
     if (appCommandShowStructuredLogs != null) {
-      environment.put("CLOUDSDK_SHOW_STRUCTURED_LOGS", appCommandShowStructuredLogs);
+      environment.put("CLOUDSDK_CORE_SHOW_STRUCTURED_LOGS", appCommandShowStructuredLogs);
     }
     // This is to ensure IDE credentials get correctly passed to the gcloud commands, in Windows.
     // It's a temporary workaround until a fix is released.
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -59,7 +59,7 @@ import javax.annotation.Nullable;
  */
 public class CloudSdk {
 
-  public static final CloudSdkVersion MINIMUM_VERSION = new CloudSdkVersion("145.0.0");
+  public static final CloudSdkVersion MINIMUM_VERSION = new CloudSdkVersion("159.0.0");
 
   private static final Logger logger = Logger.getLogger(CloudSdk.class.getName());
   private static final Joiner WHITESPACE_JOINER = Joiner.on(" ");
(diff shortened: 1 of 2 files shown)


patflynn:
this environment variable has always been present in cloud sdk?

etanshaul:
No it was added as part of the structured error changes. It made the 159 release.

patflynn:
Do we need to update the minimum cloud sdk version?

etanshaul:
No, not necessarily. Setting this env var on an earlier version of the SDK will have no effect.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -585,6 +592,15 @@ public Builder appCommandOutputFormat(String appCommandOutputFormat) {
     }
 
     /**
+     * Sets structured json logs for the stderr output. Supported values include 'never' (default),
+     * 'always', 'terminal', etc.
+     */
+    public Builder appCommandShowStructuredLogs(String appCommandShowStructuredLogs) {
```
elharo: Name is unwieldy. Maybe just "useStructuredLogs"?

etanshaul: the name mimics the gcloud config var name `showStructuredLogs`. This is the patter we've been using here for the other parameters. But of course this can be debated.

patflynn:
hmm.  If a client wants to show structured logs it's likely because they want to parse them.  I'm not really a big fan of the parameter silently being ignored dependent on the user's Cloud SDK version.   We have a pretty good notification UI for when the Cloud SDK needs to be updated for our plugins to work, I'd be OK with forcing them to come up to the version we need.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
etanshaul: I agree the name isn't the best though; `useStructuredLogs` doesn't help much though because if you look at the values it accepts, the name should be something like `setStructuredLogsScope` or something like that

elharo: Given that, I agree that "show" beats "use". However, I still think "appCommand" is unnecessary.

elharo:
I agree we should update the minimum SDK version.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
etanshaul: the `appCommand` prefix, while ugly, is to show clearly that the scope of the parameter is for gcloud command execution (e.g. `CloudSdk#runAppCommand` as opposed to a dev server command or appcfg etc.

I'm not denying though that this whole thing could be refactored to avoid this, but I don't think this PR is the place for it

etanshaul:
@patflynn @elharo, bumped the min sdk version to 159

etanshaul:
I'm looking into how I can add unit tests for this.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
etanshaul: actually it looks @loosebazooka has some code in the experimental package that addresses all of this.

etanshaul:
@elharo @patflynn see 1a46a29 for unit tests

etanshaul:
@patflynn @elharo, any other changes you'd like for this?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -79,22 +79,26 @@
...
-  private CloudSdk(Path sdkPath, Path javaHomePath, String appCommandMetricsEnvironment,
-                   String appCommandMetricsEnvironmentVersion,
+  private CloudSdk(Path sdkPath, Path javaHomePath,
+                   @Nullable String appCommandMetricsEnvironment,
+                   @Nullable String appCommandMetricsEnvironmentVersion,
                    @Nullable File appCommandCredentialFile,
-                   String appCommandOutputFormat, ProcessRunner processRunner,
+                   @Nullable String appCommandOutputFormat,
+                   @Nullable String appCommandShowStructuredLogs,
+                   ProcessRunner processRunner,
```
elharo: Is this Nullable? I don't see a null check in the constructor. Ditto sdkPath and javaHomePath.

etanshaul: they are not nullable

etanshaul: actually they may be nullable depending on which command is invoked..

etanshaul: @elharo. PTAL. I added an @Nullable to javaHomePath as that is the only one of those that is truely nullable.

etanshaul:
@patflynn or @elharo can I please get another review / approval? my last commit auto dismissed the last approval

PR #429, opened 2017-06-30
Better handling of invalid paths

elharo:
@loosebazooka @lesv and fix up the owners for the commit log


loosebazooka:
Looks like it might be checkstyle. Use `mvn verify` instead of `mvn test` ?

loosebazooka:
Lemme check the macos configs to see what we're doing. Maybe they updated the macos images and we're not configured correctly anymore.

loosebazooka:
Hey, I'm gone all next week, perhaps @chanseokoh can do the follow ups on this PR if any are required.

elharo:
OK, I expect at least one more PR on this issue so I'll send that to @chanseokoh

PR #431, opened 2017-07-01
unquote paths on Windows

elharo:
@lesv @chanseokoh @loosebazooka

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
@@ -31,11 +31,12 @@ import java.util.logging.Logger;
 import javax.annotation.Nullable;
 
 /**
- * Resolve paths with CloudSdk and Python defaults.
+ * Resolve paths with Google Cloud SDK and Python defaults.
  */
 public class PathResolver implements CloudSdkResolver {
 
   private static final Logger logger = Logger.getLogger(PathResolver.class.getName());
+  private static final boolean IS_WINDOWS = System.getProperty("os.name").contains("Windows");
 
   /**
    * Attempts to find the path to Google Cloud SDK.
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
```diff
@@ -111,6 +117,17 @@ private static String getLocalAppDataLocation() {
     return possiblePaths;
   }
 
+  @VisibleForTesting
+  static String unquote(String path) {
+    if (path.startsWith("\"")) {
```
chanseokoh: Isn't it safer to combine the two conditions?

`if (path.startsWith("\"") && path.endsWith("\""))`

src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
```diff
@@ -84,12 +85,17 @@ private static String getLocalAppDataLocation() {
...
-  private static List<String> getLocationsFromPath() {
+  @VisibleForTesting
+  static List<String> getLocationsFromPath(String pathEnv) {
     List<String> possiblePaths = new ArrayList<>();
-    String pathEnv = System.getenv("PATH");
 
     if (pathEnv != null) {
       for (String path : pathEnv.split(File.pathSeparator)) {
+        // Windows sometimes quotes paths so we need to strip these. 
+        // However quotes are legal in Unix paths.
```
loosebazooka: Not sure if you mean *illegal* in Unix paths

elharo: Nope, legal is correct. " is an acceptable character in Unix file names.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java
elharo: Absent normative documentation, I don't know that it is. The " is always illegal in paths on Windows, even if there's only one.

PR #433, opened 2017-07-07
Add support to specify arbitrary arguments to devappserver

### LINKED ISSUE #432: Add support to specify arbitrary arguments to devappserver
The devappservers occasionally add new command-line flags which may be necessary for some users, but may not make sense to have custom UI created for their use.  Rather than require users to run the devappservers outside of the IDEs, it would be useful to support specifying additional arguments to be provided to the devappserver command line.

briandealwis:
Fixes #432


briandealwis:
The behaviour is up to the devappserver.  The additional arguments are appended at the end, before the app directories / service files.

I imagine this will only be used to add items that we don't have any UI to manage.

briandealwis:
Any concerns @loosebazooka?

loosebazooka:
Oh oops, I was supposed to look at this. I'll do it right now.

briandealwis:
ping @loosebazooka ?

loosebazooka:
Oh sorry again.

PR #436, opened 2017-07-20
strawman API metadata format

elharo:
@patflynn @loosebazooka @briandealwis @chanseokoh Here's a straw man sample of what generic API client library metadata might look like.

### DIFF
--- a/src/main/java/com/google/cloud/tools/libraries/libraries.json
+++ b/src/main/java/com/google/cloud/tools/libraries/libraries.json
@@ -4,7 +4,7 @@
     "documentation": "https://cloud.google.com/bigquery/docs/reference/libraries",
     "description":"Google's fully managed, petabyte scale, serverless data warehouse for analytics",
     "transport": "http",
-    "icon":"http://google.tieto.com/images/services/google-cloud-platform/BigQuery_500px-1f2fdb56.png",
+    "icon":"https://cloud.google.com/_static/images/cloud/products/logos/svg/bigquery.svg",
     "clients": [
       {
       "language":"java",
--- a/src/main/java/com/google/cloud/tools/libraries/libraries.json
+++ b/src/main/java/com/google/cloud/tools/libraries/libraries.json
@@ -19,7 +19,7 @@
          "artifactId" : "google-cloud-bigquery",
          "version" : "0.22.0-beta"
       },
-      "platforms" : [
+      "compatible" : [
           "Android",
           "AES-7",
           "AES-8",
--- a/src/main/java/com/google/cloud/tools/libraries/libraries.json
+++ b/src/main/java/com/google/cloud/tools/libraries/libraries.json
@@ -1,22 +1,25 @@
 [
   {
-    "name": "BigQuery API v2",
-    "definition": "https://www.googleapis.com/discovery/v1/apis/bigquery/v2/rest",
+    "name": "Google BigQuery API",
+    "definition": "????",
     "documentation": "https://cloud.google.com/bigquery/docs/reference/rest/v2/",
     "transport": "http",
+    "icon":"http://google.tieto.com/images/services/google-cloud-platform/BigQuery_500px-1f2fdb56.png",
     "clients": [
       {
       "language":"java",
-      "site":"https://developers.google.com/api-client-library/java/apis/bigquery/v2",
-      "apireference":"https://developers.google.com/resources/api-libraries/documentation/bigquery/v2/java/latest/",
-      "revision":"352",
-      "version" : "1.22.0",
-      "status" : "GA",  
+      "site":"https://cloud.google.com/bigquery/docs/reference/libraries",
+      "apireference":"https://googlecloudplatform.github.io/google-cloud-java/0.22.0/apidocs/com/google/cloud/bigquery/package-summary.html",
+      "version" : "0.22.0",
+      "description":"The com.google.cloud.bigquery packages",
+      "status" : "beta",  
       "source" : "https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-bigquery",
       "languageLevel" : "1.7.0",
-      "groupId" : "com.google.apis",
-      "artifactId" : "google-api-services-bigquery",
-      "version" : "v2-rev352-1.22.0",
+      "mavencoordinates" : {
+		 "groupId" : "com.google.cloud",
+		 "artifactId" : "google-cloud-bigquery",
+	     "version" : "0.22.0-beta"
+      },
       "platforms" : [
           "Android",
           "AES-7",
--- a/src/main/java/com/google/cloud/tools/libraries/libraries.json
+++ b/src/main/java/com/google/cloud/tools/libraries/libraries.json
@@ -19,13 +19,7 @@
          "groupId" : "com.google.cloud",
          "artifactId" : "google-cloud-bigquery",
          "version" : "0.22.0-beta"
-      },
-      "compatible" : [
-          "Android",
-          "AES-7",
-          "AES-8",
-          "Java 7"
-      ]
+      }
     }
    ]
   }
--- a/src/main/java/com/google/cloud/tools/libraries/libraries.json
+++ b/src/main/java/com/google/cloud/tools/libraries/libraries.json
@@ -1,14 +1,14 @@
 [
   {
     "name": "Google BigQuery API",
-    "documentation": "https://cloud.google.com/bigquery/docs/reference/rest/v2/",
+    "documentation": "https://cloud.google.com/bigquery/docs/reference/libraries",
     "description":"Google's fully managed, petabyte scale, serverless data warehouse for analytics",
     "transport": "http",
     "icon":"http://google.tieto.com/images/services/google-cloud-platform/BigQuery_500px-1f2fdb56.png",
     "clients": [
       {
       "language":"java",
-      "site":"https://cloud.google.com/bigquery/docs/reference/libraries",
+      "site":"https://cloud.google.com/bigquery/docs/reference/libraries#client-libraries-install-java",
       "apireference":"https://googlecloudplatform.github.io/google-cloud-java/0.22.0/apidocs/com/google/cloud/bigquery/package-summary.html",
       "version" : "0.22.0",
       "infotip":"The com.google.cloud.bigquery packages",
--- a/src/main/java/com/google/cloud/tools/libraries/libraries.json
+++ b/src/main/java/com/google/cloud/tools/libraries/libraries.json
@@ -1,0 +1,29 @@
+[
+  {
+    "name": "BigQuery API v2",
+    "definition": "https://www.googleapis.com/discovery/v1/apis/bigquery/v2/rest",
+    "documentation": "https://cloud.google.com/bigquery/docs/reference/rest/v2/",
+    "transport": "http",
+    "clients": [
+      {
+      "language":"java",
+      "site":"https://developers.google.com/api-client-library/java/apis/bigquery/v2",
+      "apireference":"https://developers.google.com/resources/api-libraries/documentation/bigquery/v2/java/latest/",
+      "revision":"352",
+      "version" : "1.22.0",
+      "status" : "GA",  
+      "source" : "https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-bigquery",
+      "languageLevel" : "1.7.0",
+      "groupId" : "com.google.apis",
+      "artifactId" : "google-api-services-bigquery",
+      "version" : "v2-rev352-1.22.0",
+      "platforms" : [
+          "Android",
+          "AES-7",
+          "AES-8",
+          "Java 7"
+      ]
+    }
+   ]
+  }
+]
--- a/src/main/java/com/google/cloud/tools/libraries/libraries.json
+++ b/src/main/java/com/google/cloud/tools/libraries/libraries.json
@@ -2,6 +2,7 @@
   {
     "name": "Google BigQuery API",
     "documentation": "https://cloud.google.com/bigquery/docs/reference/rest/v2/",
+    "description":"Google's fully managed, petabyte scale, serverless data warehouse for analytics",
     "transport": "http",
     "icon":"http://google.tieto.com/images/services/google-cloud-platform/BigQuery_500px-1f2fdb56.png",
     "clients": [
@@ -10,7 +11,7 @@
       "site":"https://cloud.google.com/bigquery/docs/reference/libraries",
       "apireference":"https://googlecloudplatform.github.io/google-cloud-java/0.22.0/apidocs/com/google/cloud/bigquery/package-summary.html",
       "version" : "0.22.0",
-      "description":"The com.google.cloud.bigquery packages",
+      "infotip":"The com.google.cloud.bigquery packages",
       "status" : "beta",  
       "source" : "https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-bigquery",
       "languageLevel" : "1.7.0",
--- a/src/main/java/com/google/cloud/tools/libraries/libraries.json
+++ b/src/main/java/com/google/cloud/tools/libraries/libraries.json
@@ -1,7 +1,6 @@
 [
   {
     "name": "Google BigQuery API",
-    "definition": "????",
     "documentation": "https://cloud.google.com/bigquery/docs/reference/rest/v2/",
     "transport": "http",
     "icon":"http://google.tieto.com/images/services/google-cloud-platform/BigQuery_500px-1f2fdb56.png",
@@ -15,10 +14,10 @@
       "status" : "beta",  
       "source" : "https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-bigquery",
       "languageLevel" : "1.7.0",
-      "mavencoordinates" : {
-		 "groupId" : "com.google.cloud",
-		 "artifactId" : "google-cloud-bigquery",
-	     "version" : "0.22.0-beta"
+      "mavenCoordinates" : {
+         "groupId" : "com.google.cloud",
+         "artifactId" : "google-cloud-bigquery",
+         "version" : "0.22.0-beta"
       },
       "platforms" : [
           "Android",
--- a/src/main/java/com/google/cloud/tools/libraries/libraries.xml
+++ b/src/main/java/com/google/cloud/tools/libraries/libraries.xml
@@ -6,6 +6,7 @@
       documentation="https://cloud.google.com/bigquery/docs/reference/rest/v2/"
       transport="http"
       icon="http://google.tieto.com/images/services/google-cloud-platform/BigQuery_500px-1f2fdb56.png"
+      description="Google's fully managed, petabyte scale, serverless data warehouse for analytics"
     >
     
     <client language="java"
@@ -13,7 +14,7 @@
       apireference="https://googlecloudplatform.github.io/google-cloud-java/0.22.0/apidocs/com/google/cloud/bigquery/package-summary.html"
       version="0.22.0"
       status="beta"
-      description="The com.google.cloud.bigquery packages"
+      infotip="The com.google.cloud.bigquery packages"
       issuetracker="https://github.com/GoogleCloudPlatform/google-cloud-java/issues"
       source="https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-bigquery"
       languageLevel="1.7.0">
--- a/src/main/java/com/google/cloud/tools/libraries/libraries.xml
+++ b/src/main/java/com/google/cloud/tools/libraries/libraries.xml
@@ -3,7 +3,6 @@
 <apis>
 
   <api name="Google BigQuery API"
-      definition="????"
       documentation="https://cloud.google.com/bigquery/docs/reference/rest/v2/"
       transport="http"
       icon="http://google.tieto.com/images/services/google-cloud-platform/BigQuery_500px-1f2fdb56.png"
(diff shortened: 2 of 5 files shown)


chanseokoh:
Will we manually maintain transitive dependencies? I guess we may have to do so in some corner cases even if we can auto-populate them.

It'd be cool if we build a simple XSLT tool to convert this to an HTML chart or an external doc.

This metadata can become very large if we start considering multiple versions of APIs (e.g., BigQuery v1, v2, v3, ...). That said, I wish the Java API team maintained this metadata.

briandealwis:
Here's a few thoughts:
- ideally wouldn't be Java-specific but would support the silver languages
- multiple platforms: for Java, I can think of JavaSE ('desktop' apps: I want to write something against BigQuery), Android, AES-7, AES-8
     - some with version ranges: `JavaSE>=1.7`
- multiple environments (_os_ x _version_ x _arch_): Windows, OS X, iOS, Android
- I like the `status` flag
- I rather liked the Maven snippet, and that could be turned into doc pieces, and also allows adding additional knowledge like exclusions.

I was about to suggest modelling requirements and capabilities, and then realized we'd be re-inventing p2.

elharo:
Expanded to support multiple languages and separate API from client library.

src/test/java/com/google/cloud/tools/libraries/libraries.xml
```diff
@@ -1,22 +1,53 @@
...
+        </platforms>
+    </client>
+      
+    <client language="c#"
+      site="https://developers.google.com/api-client-library/dotnet/apis/bigquery/v2"
+      apireference="https://developers.google.com/resources/api-libraries/documentation/bigquery/v2/csharp/latest/"
+      status="release"    
+      source="https://github.com/GoogleCloudPlatform/google-cloud-dotnet/tree/master/apis/Google.Cloud.BigQuery.V2"
+      languageLevel="4.0"
+      appengine-standard="true"
```
briandealwis: For C#?

src/test/java/com/google/cloud/tools/libraries/libraries.xml
```diff
@@ -1,22 +1,53 @@
...
-    </library>
+  <api name="BigQuery API v2"
+      definition="https://www.googleapis.com/discovery/v1/apis/bigquery/v2/rest"
+      documentation="https://cloud.google.com/bigquery/docs/reference/rest/v2/"
+    >
+    
+    <client language="java"
+      site="https://developers.google.com/api-client-library/java/apis/bigquery/v2"
+      apireference="https://developers.google.com/resources/api-libraries/documentation/bigquery/v2/java/latest/"
+      revision="352"
```
briandealwis: What is `revision` and how does it differ from `version`?  Is this an internal detail — or how would we act upon it or surface it to the user?

src/test/java/com/google/cloud/tools/libraries/libraries.xml
```diff
@@ -1,22 +1,53 @@
...
-        <artifactId>google-api-services-bigquery</artifactId>
-        <version>v2-rev352-1.22.0</version>
-      </dependency>
-    </library>
+  <api name="BigQuery API v2"
+      definition="https://www.googleapis.com/discovery/v1/apis/bigquery/v2/rest"
+      documentation="https://cloud.google.com/bigquery/docs/reference/rest/v2/"
+    >
+    
+    <client language="java"
```
briandealwis: Would it make sense to use `<java …>` or `<c-sharp …>` so we can have technology-specific attributes like `javadoc`.

elharo: I'd prefer not to put the language names in element names. A client is a client is a client. This doesn't prohibit language specific attributes if we need them.

src/test/java/com/google/cloud/tools/libraries/libraries.xml
elharo: seems redundant with platforms; fixed

src/main/java/com/google/cloud/tools/libraries/libraries.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0">
+<!-- Warning: non-normative. Probably not even correct. -->
+<apis>
+
+  <api name="BigQuery API v2"
+      definition="https://www.googleapis.com/discovery/v1/apis/bigquery/v2/rest"
```
briandealwis: I'm guessing you're thinking this would be the source for API metadata:
```
{
 "kind": "discovery#restDescription",
 "etag": "\"YWOzh2SDasdU84ArJnpYek-OMdg/wbl_PWLcOQ9nI3adoCABAX0E-tY\"",
 "discoveryVersion": "v1",
 "id": "bigquery:v2",
 "name": "bigquery",
 "version": "v2",
 "revision": "20170721",
 "title": "BigQuery API",
 "description": "A data platform for customers to create, manage, share and query data.",
 "ownerDomain": "google.com",
 "ownerName": "Google",
 "icons": {
  "x16": "https://www.google.com/images/icons/product/search-16.gif",
  "x32": "https://www.google.com/images/icons/product/search-32.gif"
 },
 "documentationLink": "https://cloud.google.com/bigquery/",
```
This `definition` page is for REST — what about RPC?

src/main/java/com/google/cloud/tools/libraries/libraries.xml
```diff
@@ -0,0 +1,52 @@
...
+      documentation="https://cloud.google.com/bigquery/docs/reference/rest/v2/"
```
briandealwis: Why this page?  As compared to https://cloud.google.com/bigquery/docs/reference/  or the `documentationLink` attribute from the `definition` URL

src/main/java/com/google/cloud/tools/libraries/libraries.xml
```diff
@@ -0,0 +1,52 @@
...
+        </dependency>
+        <platforms>
+          <platform>Android</platform>
+          <platform>AES-7</platform>
+          <platform>AES-8</platform>
+          <platform>Java 7</platform>
+        </platforms>
+    </client>
+      
+    <client language="c#"
```
briandealwis: - C# would use NuGet: https://www.nuget.org/profiles/google-cloud
  - Python would use PIP: https://pypi.python.org/pypi/google-cloud

src/main/java/com/google/cloud/tools/libraries/libraries.xml
```diff
@@ -0,0 +1,52 @@
...
+      site="https://developers.google.com/api-client-library/dotnet/apis/bigquery/v2"
+      apireference="https://developers.google.com/resources/api-libraries/documentation/bigquery/v2/csharp/latest/"
+      status="release"    
+      source="https://github.com/GoogleCloudPlatform/google-cloud-dotnet/tree/master/apis/Google.Cloud.BigQuery.V2"
+      languageLevel="4.0"
+      package="Google.Apis.bigquery"
```
briandealwis: So this is the NuGet package name?  Would we want a `version` here too?  (And if not, why have `version` and `revision` in the `java` client section above?)

src/main/java/com/google/cloud/tools/libraries/libraries.xml
```diff
@@ -0,0 +1,52 @@
...
+          <platform>Java 7</platform>
+        </platforms>
+    </client>
+      
+    <client language="c#"
+      site="https://developers.google.com/api-client-library/dotnet/apis/bigquery/v2"
+      apireference="https://developers.google.com/resources/api-libraries/documentation/bigquery/v2/csharp/latest/"
+      status="release"    
+      source="https://github.com/GoogleCloudPlatform/google-cloud-dotnet/tree/master/apis/Google.Cloud.BigQuery.V2"
+      languageLevel="4.0"
```
briandealwis: By `languageLevel`, do you mean .NET Framework version?  Maybe not needed with `<platforms>` section.

src/main/java/com/google/cloud/tools/libraries/libraries.xml
```diff
@@ -0,0 +1,52 @@
...
+    >
+    
+    <client language="java"
+      site="https://developers.google.com/api-client-library/java/apis/bigquery/v2"
+      apireference="https://developers.google.com/resources/api-libraries/documentation/bigquery/v2/java/latest/"
+      revision="352"
+      version="1.22.0"
+      status="release"    
+      source="https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-bigquery"
+      languageLevel="1.7.0"
```
briandealwis: Is `languageLevel` needed with `<platforms>`?

src/main/java/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -0,0 +1,31 @@
+[
+  {
+    "name": "Google BigQuery API",
+    "documentation": "https://cloud.google.com/bigquery/docs/reference/rest/v2/",
+    "transport": "http",
+    "icon":"http://google.tieto.com/images/services/google-cloud-platform/BigQuery_500px-1f2fdb56.png",
```
patflynn: we definitely want to be using a 1st party url :)

src/main/java/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -0,0 +1,31 @@
+[
+  {
+    "name": "Google BigQuery API",
+    "documentation": "https://cloud.google.com/bigquery/docs/reference/rest/v2/",
```
patflynn: I'm guessing the content is just placeholder data but I'm going to make comment about it just in case.

patflynn: documentation url should be to a landing page like: https://cloud.google.com/bigquery/docs/

src/main/java/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -0,0 +1,31 @@
...
+    "transport": "http",
+    "icon":"http://google.tieto.com/images/services/google-cloud-platform/BigQuery_500px-1f2fdb56.png",
+    "clients": [
```
patflynn: I'm hesitant to structure the data to  support multi-clients at this point.  I'd be tempted to just keep it simple and minimal for now.

src/main/java/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -0,0 +1,31 @@
...
+      {
+      "language":"java",
+      "site":"https://cloud.google.com/bigquery/docs/reference/libraries",
+      "apireference":"https://googlecloudplatform.github.io/google-cloud-java/0.22.0/apidocs/com/google/cloud/bigquery/package-summary.html",
+      "version" : "0.22.0",
+      "description":"The com.google.cloud.bigquery packages",
```
patflynn: I'd think we really just need one description per service.

src/main/java/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -0,0 +1,31 @@
...
+      "status" : "beta",  
+      "source" : "https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-bigquery",
+      "languageLevel" : "1.7.0",
+      "mavenCoordinates" : {
+         "groupId" : "com.google.cloud",
+         "artifactId" : "google-cloud-bigquery",
+         "version" : "0.22.0-beta"
+      },
+      "platforms" : [
```
patflynn: s/platforms/compatibility/ maybe?

In terms of content I think it should just be 'Java 7'.  Going forward I don't think AES-7 and AES-8 are intended to be different from a normal java environment

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: That's a landing page for BigQuery, not for the BigQuery java API client library.

elharo: Best I've found so far. Is there a more appropriate one?

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: Agreed. I couldn't find one though. Do you happen to know of one?

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: How hard would it be to introduce this in the future if we do want multiple clients? Either way, there is information that belongs to the API as a whole, and some that is specific to the client library. It seems useful to distinguish that.

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: Will there be a non-whitelisted Java 7 on Standard? 

compatibility sounds good

src/main/java/com/google/cloud/tools/libraries/libraries.json
patflynn: sounds like there's three interesting levels of documentation

product documentation - what it does
java client documentation - getting started, samples, etc
java client API reference documentation - javadoc

src/main/java/com/google/cloud/tools/libraries/libraries.json
patflynn: perhaps our UX resource might know

patflynn:
no, app engine standard java 7 will not be upgraded to remove the whitelist.

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: I'd go to four:

product documentation - what it does
API documentation - what the API does at a protocol level
java client documentation - getting started, samples, etc
java client API reference documentation - javadoc

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: I'll ask.

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: How hard would it be to introduce this in the future if we do want multiple clients? Either way, there is information that belongs to the API as a whole, and some that is specific to the client library. It seems useful to distinguish that.

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: Is there a second one here somewhere? Or did you mean the name? In Eclipse, this one is the tooltip.

patflynn: I meant the the top level description.   This library level description doesn't appear to add any information.

src/main/java/com/google/cloud/tools/libraries/libraries.json
patflynn: ok with leaving as is. would probably flatten it if it were me.

src/main/java/com/google/cloud/tools/libraries/libraries.json
patflynn: I don't think the restful endpoint documentation is pertinent to using the java client library.  The user should be able to find that from the landing page if they're curious.

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: I'm confused. I don't see another description field. Can you copy and paste it here?

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: When wearing my client developer-API user hat, the RESTful endpoint documentation is the first thing I look for. That tells me what the true interface of the API with fewer leaky abstractions.

src/main/java/com/google/cloud/tools/libraries/libraries.json
briandealwis: https://cloud.google.com/_static/images/cloud/products/logos/svg/bigquery.svg

src/main/java/com/google/cloud/tools/libraries/libraries.json
briandealwis: `Java 7` vs `AES-7` is useful to embed in desktop apps.
`AES-7` vs `AES-8` for Datastore.

src/main/java/com/google/cloud/tools/libraries/libraries.json
briandealwis: A REST api could be explicitly called out rather than be _documentation_, and ditto for RPC.  For example, datastore has both [REST](https://cloud.google.com/datastore/reference/rest/) and [RPC](https://cloud.google.com/datastore/docs/reference/rpc/) interfaces.  Though I dunno how we'd surfaced them in a UI.

src/main/java/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -0,0 +1,31 @@
...
+    "name": "Google BigQuery API",
+    "documentation": "https://cloud.google.com/bigquery/docs/reference/rest/v2/",
+    "transport": "http",
+    "icon":"http://google.tieto.com/images/services/google-cloud-platform/BigQuery_500px-1f2fdb56.png",
+    "clients": [
+      {
+      "language":"java",
+      "site":"https://cloud.google.com/bigquery/docs/reference/libraries",
+      "apireference":"https://googlecloudplatform.github.io/google-cloud-java/0.22.0/apidocs/com/google/cloud/bigquery/package-summary.html",
+      "version" : "0.22.0",
```
elharo: FYI, in Eclipse we've stopped referencing the version when creating a new project and just pull the latest version from Maven central.

patflynn: Unless the library versions are synchronized to match the BOM you will break people's projects by simply adding latest.

src/main/java/com/google/cloud/tools/libraries/libraries.json
patflynn: Users can go to the product docs, which we link to, if they want to find out about the rest api.  This is a Java library importer wizard.  Let's focus on the information pertinent to that use-case

src/main/java/com/google/cloud/tools/libraries/libraries.json
patflynn: nice find

src/main/java/com/google/cloud/tools/libraries/libraries.json
patflynn: huh weird, I thought there was a description at the top level. perhaps you removed it? In any case the description should be at the top level, and it should be a short inline blurb about what the API is for.

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: As a Java developer, I want the REST documentation that tells me what the client library is really doing. All too often the Java client documentation is incomplete or inaccurate.

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: Probably. Eclipse needs to do some more work to query the Maven repo metadata to find the appropriate dependencies.

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: Can you give me an example of what might belong in that? 

Eclipse shows this content ("The com.google.cloud.bigquery packages") in a tooltip on the library, and it is specific to the Java client library.

patflynn:
Something from their marketing copy.

like: "A fast, economical and fully managed data warehouse for large-scale
data analytics."

your tooltip text you could easily generate from the data you already have.

src/main/java/com/google/cloud/tools/libraries/libraries.json
patflynn: Totally get it. If you want to find that information you can click into any of the other links we provide and get there from the big query documentation left nav.  Three levels of links per API in the wizard is more than enough.

elharo:
Added the description and URLs. The infotip is not necessarily just the package names.

src/main/java/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -0,0 +1,32 @@
...
+      "status" : "beta",  
+      "source" : "https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-bigquery",
+      "languageLevel" : "1.7.0",
+      "mavenCoordinates" : {
+         "groupId" : "com.google.cloud",
+         "artifactId" : "google-cloud-bigquery",
+         "version" : "0.22.0-beta"
+      },
+      "compatible" : [
+          "Android",
```
patflynn: the Android and AES-7 elements are wrong. They're explicitly unsupported for all google-cloud-java libraries.

more generally AES-8 and 'Java 7' are supposed to be equivalent by design.  i.e. AES-8 is not a 'special' Java environment. It's standard Java and we don't need to be explicit about it's compatibility IMO.

elharo: We could simply drop the platforms section for now. We can always put it back in later if we discover a need for it. What do you think?

That said, I think Java 7 should work in all environments including Android for HTTP transport based protocols per https://github.com/GoogleCloudPlatform/google-cloud-java#supported-platforms
 
"Clients in this repository use either HTTP or gRPC for the transport layer. All HTTP-based clients should work in all environments."

patflynn:
SGTM. 

Android is explicitly called out as unsupported in our documentation.

patflynn:
go for it.

PR #440, opened 2017-08-04
Add gcloud JSON output de-serialization classes

chanseokoh:
- `AppEngineDeployResult` for the `gcloud app deploy` stdout, which contains the info about deployed apps.
- `GcloudStructuredLog` for structured log entries from `gcloud`

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/JsonParseException.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/JsonParseException.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.appengine.cloudsdk;
+
+public class JsonParseException extends Exception {
+
+  public JsonParseException(Throwable cause) {
+    super(cause);
+  }
+}
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/JsonParseException.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/JsonParseException.java
@@ -16,6 +16,13 @@
 
 package com.google.cloud.tools.appengine.cloudsdk;
 
+/**
+ * Signals that there have been problems when parsing JSON input.
+ *
+ * <p>May signal general problems, including syntactic or semantic errors. However, the main reason
+ * that this exception was introduced in the library is to wrap the {@link
+ * com.google.gson.JsonSyntaxException} {@link RuntimeException} and make it a checked exception.
+ */
 public class JsonParseException extends Exception {
 
   public JsonParseException(Throwable cause) {
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/AppEngineDeployResult.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/AppEngineDeployResult.java
@@ -25,15 +25,16 @@ import java.util.List;
 /**
  * Holds de-serialized JSON result output of {@code gcloud app deploy}.
  */
-// Don't change the field names because Gson uses them for automatic de-serialization.
 public class AppEngineDeployResult {
 
   private static class Version {
-    String id;
-    String service;
-    String project;
+    // Don't change the field names because Gson uses them for automatic de-serialization.
+    private String id;
+    private String service;
+    private String project;
   }
 
+  // Don't change the field names because Gson uses them for automatic de-serialization.
   private List<Version> versions;
 
   private AppEngineDeployResult() {}  // empty private constructor
@@ -72,7 +73,8 @@ public class AppEngineDeployResult {
    * Parses a JSON string representing successful {@code gcloud app deploy} result.
    *
    * @return parsed JSON; never {@code null}
-   * @throws JsonParseException if {@code jsonString} has syntax errors or malformed JSON elements
+   * @throws JsonParseException if {@code jsonString} has syntax errors or incompatible JSON
+   *     element type
    */
   public static AppEngineDeployResult parse(String jsonString) throws JsonParseException {
     Preconditions.checkNotNull(jsonString);
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/AppEngineDeployResult.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/AppEngineDeployResult.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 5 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/GcloudStructuredLog.java
```diff
@@ -0,0 +1,61 @@
...
+   * @throws JsonSyntaxException if {@code jsonString} has syntax errors
+   * @throws JsonParseException if {@code jsonString} has semantic errors
+   *     (e.g., missing 'timestamp' property)
+   */
+  public static GcloudStructuredLog parse(String jsonString) {
+    Preconditions.checkNotNull(jsonString);
+    GcloudStructuredLog json = new Gson().fromJson(jsonString, GcloudStructuredLog.class);
+    if (json.version == null || json.verbosity == null || json.timestamp == null
+        || json.message == null) {
+      throw new JsonParseException("cannot parse gcloud structured log entry: " + jsonString);
```
elharo: this should really be a checked exception :-(

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/GcloudStructuredLog.java
```diff
@@ -0,0 +1,61 @@
...
+// Don't change the field names because Gson uses them for automatic de-serialization.
+public class GcloudStructuredLog {
+
+  public static class Error {
+    String type;
+    String stacktrace;
+    String details;
+  }
+
+  public String version;
```
elharo: These shouldn't be public, especially mutable and public.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/GcloudStructuredLog.java
chanseokoh: What's your call? Defining our own exception (or two exceptions)?

chanseokoh:
@etanshaul how do you feel about this?

chanseokoh:
After discussion with @elharo, I am going to remove the semantic checks. Any missing fields will just be `null`, and clients will decide what to do with them.

patflynn:
please include @loosebazooka on these PRs.

@loosebazooka is there a way to setup the PR template to always include you as reviewer?

patflynn:
@chanseokoh also we need more context on what problem this change is solving, how it's supposed to be used, and why and how it integrates with the core library APIs.

etanshaul:
agree with @patflynn; decoupling this from the consumption makes it difficult to follow without more explanation

chanseokoh:
`AppEngineDeployResult` is a class with 1-to-1 correspondence to the JSON that a successful `gcloud app deploy` outputs. The JSON looks like this:

```
{
  'configs': [],
  'versions': [
    {
      'id': '20160429t112518',
      'last_deployed_time': null,
      'project': 'bizarre-project',
      'service': 'display-service',
      'traffic_split': null,
      'version': null
    }
  ]
}
```

As a de-serialization of JSON to a Java class, it allows clients of `appengine-plugins-core` to get the values of the JSON in Java language. The primary usage in our plugin context is to found out the app version generated by `gcloud` (as well as the service and the project), so that we can open a browser pointing to the deployed app with the right URL.

AFAIK, Eclipse and IntelliJ both have this class in duplication. I am pushing this to `appengine-plugins-core`. As a Java representation of JSON output by `gcloud`, I think this fits into `appengine-plugins-core`.

Similarly, `GcloudStructuredLog` is a 1-to-1 de-serialization of JSON that gcloud outputs in the recently proposed structured log format. (I believe you all know what this structured log format is.) The primary usage in our plugin context is to reliably recognize and extract deploy error messages that are mixed together with normal `gcloud app deploy` output.

etanshaul:
For the deployment output, what are the proposed changes to the deploy API to support this? Returning an `AppEngineDeployResult` from `AppEngineDeployment#deploy`?

For the structured output part I'm less clear on how you plan on supporting this. Currently, the stdout is sent line by line using a `ProcessOutputLineListener`. Could you elaborate on how `GcloudStructuredLog` will be used in this context?

chanseokoh:
They are just standalone de-serialization classes correspond to the said JSON outputs. Nothing more than that. They are on their own, and they are not coupled in any way with other existing APIs.

So, clients have responsibility to capture the relevant JSON first. Then, they will simply call `.parse(jsonString)` to de-serialize it into Java. That said, clients still need to use `ProcessOutputLineListener` and do whatever is needed to get the JSON portion. We could add more supporting and convenience APIs to cover this portion in `appengine-plugins-core` too, but I'm not tempted to go this way until we complete migration to use `Future`.

etanshaul:
Aha I see. I didn't realize that the scope of this was just to encapsulate these value classes in the core-lib so that the clients don't have to duplicate them. Makes sense.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/JsonParseException.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.appengine.cloudsdk;
+
+public class JsonParseException extends Exception {
```
elharo: Javadoc

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/GcloudStructuredLog.java
```diff
@@ -0,0 +1,90 @@
...
+
+import com.google.cloud.tools.appengine.cloudsdk.JsonParseException;
+import com.google.common.base.Preconditions;
+import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
+
+/**
+ * Holds de-serialized JSON of a single instance of structured log output from {@code gcloud}.
+ */
+// Don't change the field names because Gson uses them for automatic de-serialization.
```
elharo: should probably move this comment to the field name section

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/AppEngineDeployResult.java
```diff
@@ -0,0 +1,85 @@
...
+   */
+  public String getProject(int index) {
+    return versions.get(index).project;
+  }
+
+  /**
+   * Parses a JSON string representing successful {@code gcloud app deploy} result.
+   *
+   * @return parsed JSON; never {@code null}
+   * @throws JsonParseException if {@code jsonString} has syntax errors or malformed JSON elements
```
elharo: Is a malformed JSON element not a syntax error?

chanseokoh: I should have said "incompatible JSON element type". For example, here we expect an array for the `version` property, but if a JSON string has a string type instead of an array, GSON will throw `JsonSyntaxException` (which I think isn't really correct honestly).

chanseokoh:
@loosebazooka are you good with this?

loosebazooka:
Yeah sorry I didn't get to see this. Do we have a guarantee from the cloud SDK team that the format wont change? How can we break fast when the cloud sdk makes a breaking change? Will that just manifest itself in integration tests from the IDEs?

chanseokoh:
There is no such a guarantee of course (anything can change hypothetically), though it will be very unusual for them to change the deploy output JSON format. CT4E for example will already break now if the format changes. However, for the logging message format, this is quite new in Cloud SDK. This is mainly for extracting an error message, so I expect IDEs will fall back to a default canned message if it fails to extract a message for any reason (including JSON format change or missing JSON properties).

PR #442, opened 2017-08-11
Don't validate jdk when validating the cloud sdk

### LINKED ISSUE #441: Should CloudSDK validation be coupled with jdk validation?
[CloudSdk#validateCloudSdk](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/713bff4c658b62640345e6312f59089c0e2c3437/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java#L469) also validates that the system jdk points to a proper executable. 

Should these be coupled like this? This causes all validation of the CloudSDK to fail on improper jdk config. Presumably not all usages of the SDK require this.

etanshaul:
Fixes #441 

Removes jdk validation from the SDK validation method.

I did _not_ remove the jdk validation entirely:

It is still invoked when running an appCfg command or a devAppServer1 command


etanshaul:
@elharo could you take a look as well to see if/how this would affect eclipse?

elharo:
Is this still useful? Can we get it in?

loosebazooka:
@etanshaul  We still want this, yeah?

etanshaul:
I think this has implications for the clients. Let me check and refresh my memory on this.

etanshaul:
@loosebazooka this change still makes sense, and is ok from the IntelliJ side.

If it still is ok from for Eclipse then we can merge it in. @elharo @chanseokoh

loosebazooka:
@TadCordle would we also need to update our sdk checker in build tools?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -506,7 +506,6 @@ public void validateCloudSdk()
           InvalidJavaSdkException {
```
chanseokoh: Need to remove `InvalidJavaSdkException`.

TadCordle:
> @TadCordle would we also need to update our sdk checker in build tools?

If we still want jdk validation, then yes. If we still want to do validation, is the only way of doing that now running an appCfg command or a devAppServer1 command and seeing if it throws an exception?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
etanshaul: done

etanshaul:
when would you want to call `validateJdk`? @loosebazooka

etanshaul:
@chanseokoh sorry, pushed a minor cleanup which dismissed your approval

etanshaul:
@ehlaro, yes, please approve and I'll merge it in after CI is complete. thanks

etanshaul:
@elharo actually looks like its already approved. never mind. will merge.

PR #443, opened 2017-08-25
expand JSON test coverage; add datastore and cloud storage

elharo:
@briandealwis @loosebazooka @garrettjonesgoogle @eamonnmcmanus @lesv


src/main/java/com/google/cloud/tools/libraries/MavenCoordinates.java
```diff
@@ -0,0 +1,209 @@
...
+
+  @Override
+  public String toString() {
+    return MessageFormat.format("MavenCoordinates [repository={0}, {1}:{2}:{3}:{4}:{5}]",
+                                repository, groupId, artifactId, type, classifier, version);
+  }
+  
+  @Override 
+  public int hashCode() {
+    return (groupId + ":" + artifactId + ":" + version).hashCode();
```
briandealwis: Use `Objects.hash()`?

src/main/java/com/google/cloud/tools/libraries/MavenCoordinates.java
```diff
@@ -0,0 +1,209 @@
...
+  }
+  
+  @Override 
+  public boolean equals(Object o) {
+    if (o == null || !(o instanceof MavenCoordinates)) {
+      return false;
+    }
+    MavenCoordinates other = (MavenCoordinates) o;
```
briandealwis: `equals()` needs to check `repository`, `classifier`, and `type`.   Use `Object.equals()`?

src/test/java/com/google/cloud/tools/libraries/LibrariesTest.java
```diff
@@ -44,19 +62,63 @@ public void testJson() throws FileNotFoundException, URISyntaxException {
...
+    }
+    JsonArray clients = api.getJsonArray("clients");
+    Assert.assertFalse(clients.isEmpty());
+    for (int i = 0; i < clients.size(); i++) {
+      JsonObject client = (JsonObject) clients.get(i);
+      String status = client.getString("status");
+      Assert.assertTrue(status, 
+          "beta".equals(status) || "alpha".equals(status) || "GA".equals(status));
+      new URI(client.getString("apireference"));
+      Assert.assertTrue("1.7.0".equals(client.getString("languageLevel")));
```
briandealwis: The precision seems odd.

src/test/java/com/google/cloud/tools/libraries/LibrariesTest.java
```diff
@@ -44,19 +62,63 @@ public void testJson() throws FileNotFoundException, URISyntaxException {
...
+    new URI(api.getString("site"));
+    if (api.getString("icon") != null) {
+      new URI(api.getString("icon"));
+    }
+    JsonArray clients = api.getJsonArray("clients");
+    Assert.assertFalse(clients.isEmpty());
+    for (int i = 0; i < clients.size(); i++) {
+      JsonObject client = (JsonObject) clients.get(i);
+      String status = client.getString("status");
+      Assert.assertTrue(status,
```
briandealwis: Could use Hamcrest's `org.hamcrest.collection.IsArrayContaining` so the invalid value is shown in the failure message.

src/test/java/com/google/cloud/tools/libraries/MavenCoordinatesTest.java
```diff
@@ -0,0 +1,164 @@
...
+
+package com.google.cloud.tools.libraries;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThat;
+
+import org.junit.Test;
+
+public class MavenCoordinatesTest {
```
briandealwis: Need some tests for `equals()`

loosebazooka:
The PR say's it's expanding JSON test coverages but makes changes to a non-test class? Is there more going on here?

src/test/java/com/google/cloud/tools/libraries/LibrariesTest.java
elharo: done

src/main/java/com/google/cloud/tools/libraries/MavenCoordinates.java
briandealwis: The classifier and type are pretty key coordinates.  I think you should have a different method for identifying equivalence that ignores repository. Or perhaps the repository should not be a coordinate at all.

elharo: Now that you mention it, what I really need to verify here is that the groupId:artifactId is unique, not even including version.

elharo:
We needed MavenCoordinates for the tests, or did at one point. Now I'm not so sure. I expect there'll be some libraries for processing the metadata, though I'm still not sure which piece swill land here and which in the IDE repos.

loosebazooka:
If it's only being used for tests, can it be in the `src/test/java` directory instead?

elharo:
reverted for now; it can come back later if needed

src/test/java/com/google/cloud/tools/libraries/LibrariesTest.java
```diff
@@ -43,20 +62,66 @@ public void testJson() throws FileNotFoundException, URISyntaxException {
...
-    for (JsonObject api = (JsonObject) apis.next(); apis.hasNext(); api = (JsonObject) apis.next()) {
+    while (apis.hasNext()) { 
+      JsonObject api = (JsonObject) apis.next();
+      verifyApi(api);
+    }
+  }
+
+  private static final String[] statuses = {"alpha", "beta", "GA"};
+
+  private static void verifyApi(JsonObject api) throws URISyntaxException {
```
loosebazooka: nit : would using assertApi instead of verifyApi clarify that the intention of the method is to make assertions?

src/test/java/com/google/cloud/tools/libraries/LibrariesTest.java
```diff
@@ -43,20 +62,66 @@ public void testJson() throws FileNotFoundException, URISyntaxException {
...
+    InputStream in =
+        new FileInputStream("src/main/java/com/google/cloud/tools/libraries/libraries.json");
+    JsonReader reader = factory.createReader(in); 
+    Iterator<JsonValue> apis = reader.readArray().iterator();
+    Assert.assertTrue(apis.hasNext());
     
-      new URI(api.getString("documentation"));
-      if (api.getString("icon") != null) {
-        new URI(api.getString("icon"));
+    Map<String, String> map = new HashMap<>();
```
loosebazooka: name of `map` maybe can be more descriptive.

src/test/java/com/google/cloud/tools/libraries/LibrariesTest.java
```diff
@@ -43,20 +62,66 @@ public void testJson() throws FileNotFoundException, URISyntaxException {
...
+      Assert.assertTrue(versionString.matches("\\d+\\.\\d+\\.\\d+"));
+      Assert.assertNotNull(client.getJsonObject("mavenCoordinates"));
+      if (client.getString("source") != null) {
+        new URI(client.getString("source"));
+      }
+    }
+  }
+  
+  @Test
+  public void testMavenCoordinates() throws FileNotFoundException, URISyntaxException {
```
loosebazooka: could this be named testMavenCoordinate_fileIsValid or something? It's hard for me to tell (since I didn't really look at the previous CL) what this test is doing.

src/main/java/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -7,10 +7,10 @@
     "icon":"https://cloud.google.com/_static/images/cloud/products/logos/svg/bigquery.svg",
     "clients": [
       {
+      "name":"Google Cloud Java Client for BigQuery",
       "language":"java",
       "site":"https://cloud.google.com/bigquery/docs/reference/libraries#client-libraries-install-java",
-      "apireference":"https://googlecloudplatform.github.io/google-cloud-java/0.22.0/apidocs/com/google/cloud/bigquery/package-summary.html",
-      "version" : "0.22.0",
+      "apireference":"https://googlecloudplatform.github.io/google-cloud-java/latest/apidocs/com/google/cloud/bigquery/package-summary.html",
```
briandealwis: Would this be clearer as a language-specific field called `javadoc`?
Oh, actually we should have a `javadoc` field for programmatic access (e.g., Eclipse javadoc hovers) as well as a the more human-meaningful API starting point.

src/main/java/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -7,10 +7,10 @@
...
       "infotip":"The com.google.cloud.bigquery packages",
```
briandealwis: This might be useful to encode in the metadata to enable UI discovery (e.g., a quick-fix on unresolved symbols)

src/main/java/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -19,8 +19,56 @@
...
       }
-    }
-   ]
+    ]
+  },
+  {
+    "name": "Google Cloud Datastore",
+    "documentation": "https://cloud.google.com/datastore/docs/reference/libraries",
+    "description":"Google Cloud Datastore is a NoSQL key-value database built for automatic scaling.",
+    "transport": "http",
```
briandealwis: - Should `transport` be an array?  What do we do for libraries that support both `http` and `grpc`?
- Shouldn't this really be in the `clients` section?  We may have libraries that just use the `http` REST APIs, even though the API supports both

src/main/java/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -19,8 +19,56 @@
...
+    "name": "Google Cloud Storage",
+    "documentation": "https://cloud.google.com/storage/docs/reference/libraries",
+    "description":"Unified BLOB storage for developers and enterprises, from live data serving to data analytics/ML to data archiving.",
+    "transport": "http",
+    "icon":"https://cloud.google.com/_static/images/cloud/products/logos/svg/storage.svg",
+    "clients": [
+      {
+      "name":"Google Cloud Storage Client Library for Java",
+      "language":"java",
+      "site":"https://cloud.google.com/storage/docs/reference/libraries#client-libraries-install-java",
```
briandealwis: Wouldn't the `#using_the_client_library` section be more appropriate, since the user will be using the consuming UI to automate the installation of this library?

src/test/java/com/google/cloud/tools/libraries/LibrariesTest.java
```diff
@@ -36,27 +65,59 @@ public void testWellFormed() throws ParserConfigurationException, SAXException,
...
-      for (int i = 0; i < clients.size(); i++) {
-        JsonObject client = (JsonObject) clients.get(i);
-        String status = client.getString("status");
-        Assert.assertTrue(status, 
-            "beta".equals(status) || "alpha".equals(status) || "GA".equals(status));
+      JsonObject coordinates =
+          ((JsonObject) api.getJsonArray("clients").get(0)).getJsonObject("mavenCoordinates");
+      String mavenCoordinates = 
+          coordinates.getString("groupId") + ":" + coordinates.getString("artifactId");
+      if (apiCoordinates.containsValue(mavenCoordinates)) {
```
briandealwis: It should be ok to provide different versions of an API library, shouldn't it?  E.g., Java libraries for BigQuery 1.0.0 vs 2.0.0?

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: can you elaborate? I'm not sure exactly what you're proposing here.

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: do any libraries support more than one transport? let me check with the api team.

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: What I really wanted was the entire page, but that won't pick up Java as opposed to Python. This is the nearest link to the top.

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: still waiting for a reply but the docs do say, "Clients in this repository use either HTTP or gRPC for the transport layer. The README of each client documents the transport layer the client uses."

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: what would the difference be. For example, what javadoc URL would Eclipse need here?

src/main/java/com/google/cloud/tools/libraries/libraries.json
briandealwis: Expressing this as:
```
"javaPackages":["com.google.cloud.bigquery", "com.google.cloud.bigquery.**"],
```

For example, PDE contributes a quick-fix to suggesting adding an `Import-Package` or `Require-Bundle` on an unresolved symbol that is found in a bundle:

screen shot 2017-08-28 at 4 22 18 pm

src/main/java/com/google/cloud/tools/libraries/libraries.json
briandealwis: When I try setting the Javadoc URL for a jar to this URL (right-click on a jar, like a bundle in the _Plug-in Dependencies_, and select _Javadoc Location_, and set as a URL), Eclipse shown an error:
screen shot 2017-08-28 at 4 32 01 pm

The java docs have a specific layout, as does the HTML.  So Eclipse is able to find and parse out the appropriate Javadoc to be shown in a hover.

elharo: To make that work, I think we first need to get google-cloud-java to publish individual javadoc for each API. The file it's looking for is at

https://googlecloudplatform.github.io/google-cloud-java/latest/apidocs/

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: That would be something different. This is human readable text for the infotip on the library.

src/test/java/com/google/cloud/tools/libraries/LibrariesTest.java
elharo: I think that would have a different group or artifact ID. I don't think we want both 1.3.0 and 1.4.0.

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: We should probably pull IDE readable source and javadoc jars from Maven Central rather than calling them out as a separate  artifact.

garrettjonesgoogle: +1 to @elharo 's comment - then the javadoc and source will actually match the classes that the user is using.

src/main/java/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -19,8 +19,56 @@
...
+    "transport": "http",
+    "icon":"https://cloud.google.com/_static/images/cloud/products/logos/svg/datastore.svg",
+    "clients": [
+      {
+      "name":"Cloud Datastore Client Library for Java",
+      "language":"java",
+      "site":"https://cloud.google.com/datastore/docs/reference/libraries#client-libraries-install-java",
+      "apireference":"https://googlecloudplatform.github.io/google-cloud-java/latest/apidocs/index.html?com/google/cloud/datastore/package-summary.html",
+      "infotip":"The com.google.cloud.datastore packages",
+      "status" : "GA",
```
garrettjonesgoogle: I think `releaseLevel` would be a more accurate description of the Alpha/Beta/GA values

src/main/java/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -19,8 +19,56 @@
...
+      "site":"https://cloud.google.com/datastore/docs/reference/libraries#client-libraries-install-java",
+      "apireference":"https://googlecloudplatform.github.io/google-cloud-java/latest/apidocs/index.html?com/google/cloud/datastore/package-summary.html",
+      "infotip":"The com.google.cloud.datastore packages",
+      "status" : "GA",  
+      "source" : "https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-bigquery",
+      "languageLevel" : "1.7.0",
+      "mavenCoordinates" : {
+         "groupId" : "com.google.cloud",
+         "artifactId" : "google-cloud-datastore",
+         "version" : "1.4.0"
```
garrettjonesgoogle: Will this version have to be updated with every google-cloud-java release?

src/main/java/com/google/cloud/tools/libraries/libraries.json
briandealwis: It would be good to have this in a machine readable format too for the tooling reasons above.

Maybe it would help to add additional information here.  For example, there are sometimes alpha versions of libraries that users may want to try.  Or if this is the Google API Client library, we could add a pointer to the Google Client library?

src/main/java/com/google/cloud/tools/libraries/libraries.json
briandealwis: Stepping back, what will we do with this information?  Are we going to use it as a filter — e.g., if `grpc` then cannot be used on AESv7?

src/main/java/com/google/cloud/tools/libraries/libraries.json
briandealwis: Yes

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: CT for IntelliJ plans to expose these links to the end user so it needs human browsable URLs.

The links to the Maven source/javadoc jars don't need to be explicitly included here because they can be derived from the Maven coordinates.

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: wikipedia seems to call this a "stage" though I can't say I've heard that before: https://en.wikipedia.org/wiki/Software_release_life_cycle#Stages_of_development

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: Yes, exactly that.

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: Sounds good. How do we tell Eclipse about our packages when adding a library?

src/main/java/com/google/cloud/tools/libraries/libraries.json
garrettjonesgoogle: It looks like "launch stage" is actually the terminology used by Google Cloud Platform: https://cloud.google.com/terms/launch-stages

However, we have to keep in mind that the release level of client libraries will lag behind the service.

src/main/java/com/google/cloud/tools/libraries/libraries.json
briandealwis: So is better to encode this platform-compatibility knowledge into the consumers of this API, or to make it explicit in a compatibility/platform section?  I lean towards the latter.

src/main/java/com/google/cloud/tools/libraries/libraries.json
briandealwis: I think we'd have a quick-fix on an unresolved symbol like [PDE's quick-fix](http://git.eclipse.org/c/pde/eclipse.pde.ui.git/tree/ui/org.eclipse.pde.ui/plugin.xml?h=master#n1656).  Our quick-fix would iterate on the library definitions to see if it can find a match.

src/main/java/com/google/cloud/tools/libraries/libraries.json
briandealwis: Sounds good.

src/main/java/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -3,24 +3,72 @@
...
+       }
+      }
+    ]
+  },
+  {
+    "name": "Google Cloud Storage",
+    "documentation": "https://cloud.google.com/storage/docs/reference/libraries",
+    "description":"Unified BLOB storage for developers and enterprises, from live data serving to data analytics/ML to data archiving.",
+    "transport": ["http"],
```
briandealwis: Pluralize transport*s*?  since we have `clients` and `mavenCoordinates`?

src/main/java/com/google/cloud/tools/libraries/libraries.json
briandealwis: oops, ping on this request too.

elharo: done

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: done

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: Let me file a bug requesting a URL for the whole page.

src/main/java/com/google/cloud/tools/libraries/libraries.json
elharo: Of course clients do have the option to ignore the version here and query Maven central for the latest release version. CT4E contains code to do this.

elharo:
Pinging everyone. Are there any final blockers that prevent us from checking this in and moving forward with the rest of the google-cloud-java APIs? 

We can easily add additional content later. Changing or removing existing fields should be done here if possible.

If not, it needs an approval.

PR #449, opened 2017-09-12
Add remaining metadata

elharo:
@briandealwis @loosebazooka  @patflynn @etanshaul got a bit confused by the capitalization, but this is the current branch

### DIFF
--- a/src/main/java/com/google/cloud/tools/libraries/libraries.json
+++ b/src/main/java/com/google/cloud/tools/libraries/libraries.json
@@ -97,6 +97,30 @@
       }
     ]
   },
+  {
+    "name": "Stackdriver Trace",
+    "documentation": "https://cloud.google.com/trace/docs/reference/libraries",
+    "description":"A distributed tracing system that collects latency data from App Engine applications.",
+    "transports": ["grpc"],
+    "icon":"https://cloud.google.com/_static/images/cloud/products/logos/svg/trace.svg",
+    "clients": [
+      {
+      "name":"Java Client Libraries for the Stackdriver Trace API",
+      "compute":"java",
+      "apireference":"https://googlecloudplatform.github.io/google-cloud-java/latest/apidocs/index.html?com/google/cloud/trace/v1/package-summary.html",
+      "infotip":"The com.google.cloud.trace.v1 packages",
+      "launchStage" : "alpha",  
+      "packages" : ["com.google.cloud.trace.v1"],
+      "source" : "https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-trace",
+      "languageLevel" : "1.7.0",
+      "mavenCoordinates" : {
+         "groupId" : "com.google.cloud",
+         "artifactId" : "google-cloud-trace",
+         "version" : "0.23.0-alpha"
+       }
+      }
+    ]
+  },
  {
     "name": "Cloud DNS",
     "documentation": "https://cloud.google.com/dns/docs/reference/libraries",
--- a/src/main/java/com/google/cloud/tools/libraries/libraries.json
+++ b/src/main/java/com/google/cloud/tools/libraries/libraries.json
@@ -44,7 +44,7 @@
       "mavenCoordinates" : {
          "groupId" : "com.google.cloud",
          "artifactId" : "google-cloud-vision",
-         "version" : "0.22.0-beta"
+         "version" : "0.23.0-beta"
        }
       }
     ]
@@ -69,7 +69,7 @@
       "mavenCoordinates" : {
          "groupId" : "com.google.cloud",
          "artifactId" : "google-cloud-language",
-         "version" : "0.22.0-beta"
+         "version" : "0.23.0-beta"
        }
       }
     ]
@@ -94,7 +94,7 @@
       "mavenCoordinates" : {
          "groupId" : "com.google.cloud",
          "artifactId" : "google-cloud-spanner",
-         "version" : "0.22.0-beta"
+         "version" : "0.23.0-beta"
        }
       }
     ]
@@ -119,7 +119,7 @@
       "mavenCoordinates" : {
          "groupId" : "com.google.cloud",
          "artifactId" : "google-cloud-pubsub",
-         "version" : "0.22.0-beta"
+         "version" : "0.23.0-beta"
        }
       }
     ]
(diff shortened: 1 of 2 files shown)


loosebazooka:
I don't know enough about this api stuff, perhaps someone on the intellij team who might be using it can review?

src/main/java/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -1,5 +1,320 @@
...
+      "apireference":"https://googlecloudplatform.github.io/google-cloud-java/latest/apidocs/?com/google/cloud/resourcemanager/package-summary.html",
+      "infotip":"The com.google.cloud.speech.resourcemanager packages",
+      "launchStage" : "alpha",  
+      "packages" : ["com.google.cloud.resourcemanager"],
+      "source" : "https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-resourcemanager",
+      "languageLevel" : "1.7.0",
+      "mavenCoordinates" : {
+         "groupId" : "com.google.cloud",
+         "artifactId" : "google-cloud-resourcemanager",
+         "version" : "0.23.0-alpha"
```
patflynn: Should this be version**s**?  how do we encode the BOM data?

src/main/java/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -1,5 +1,320 @@
...
+    "icon":"https://cloud.google.com/_static/images/cloud/products/logos/svg/dlp.svg",
+    "clients": [
+      {
+      "name":"Java Client Libraries for the Data Loss Prevention API",
+      "language":"java",
+      "site":"https://cloud.google.com/dlp/docs/reference/libraries#client-libraries-install-java",
+      "apireference":"https://googlecloudplatform.github.io/google-cloud-java/latest/apidocs/index.html?com/google/cloud/dlp/v2beta1/package-summary.html",
+      "infotip":"The com.google.cloud.dlp.v2beta1 packages",
+      "launchStage" : "alpha",  
+      "packages" : ["com.google.cloud.dlp.v2beta1"],
```
patflynn: what does packages represent?

briandealwis: The Java packages exported by the library.  We can use them in quick-fixes to add a library on an unresolved symbol.

patflynn: We have to manually manage this though. Couldn't you just build a package index from the library?

briandealwis: That would require downloading each of the libraries (and possibly dependencies) and process them.  And it also doesn't help us avoid exposing shaded jars (e.g., `*.repackaged.*`).

elharo:
Ping

src/main/java/com/google/cloud/tools/libraries/libraries.json
briandealwis: These entries should be defined in relation to a common BOM: we don't want to mix-and-match between BOMs, right?  The BOM definition could be be external (e.g., encoded in the file-name like `libraries-0.22.json`), or we could make that an outer element if you wanted a single library definition.

elharo: We could pull this from the google-cloud-java pom.xml though it's an extra lookup.
E.g. https://search.maven.org/remotecontent?filepath=com/google/cloud/google-cloud-pom/0.23.1-alpha/google-cloud-pom-0.23.1-alpha.pom

elharo:
I'm checking with the internal Javascript stylists to see if they prefer "foo":"bar" or "foo" : "bar".

src/main/java/com/google/cloud/tools/libraries/libraries.json
briandealwis: It would be worth having a test that pulls the dependencies for each and ensures there's no conflicting versions.

etanshaul:
The PRD specifies that the selector should also include a link to the issue tracker for each library. Should we add this?

elharo:
What's the purpose of the issue tracker link? Is this something we want to expose in the UI?

briandealwis:
For making requests for changes or enhancements to the APIs?  Saves the developer from having to hunt the right tracker down, and becoming confused by the com.google.api.* vs com.google.cloud.* repositories.

src/main/java/com/google/cloud/tools/libraries/libraries.json
patflynn: Aha I see how this can be useful, but I'm still skeptical of our ability to maintain this data.  Until we have concrete plans to use this data I'd prefer if we left it out.

I really like the idea of suggesting libraries based off of usage, but I'd like that feature designed first.

briandealwis: This is already a feature available for plugin development in Eclipse, where Eclipse PDE scans for the packages and classes offered by the available plugins and offers to configure the plugin classpath on an unresolved symbol:

screen shot 2017-09-18 at 4 38 08 pm

That said, the real help is PDE recognizing class names, not package names.  E.g., that `BigtableDataClient` &rarr; `com.google.cloud.bigtable`.  So +1 to removing the package names.

patflynn: @briandealwis thanks for the details and screenshot!

elharo:
I'm very skeptical of exposing the API issue trackers in the IDEs. This is unlikely to appear at a place where the user might want it (on the code). It's more likely to collect random bugs that have nothing to do with the APIs or CT4E.

etanshaul:
I think its convenient (and can't hurt) to have this link. At a minimum, it gives the user the ability to quickly view open issues & activity prior to adding the library. I do this sometimes myself.

However I'll defer to the PRD authors. @patflynn wdyt?

elharo:
It's easier to add something later than take it out, so I default to leaving fields out until we have consensus that they're needed.

Also note that for these libraries there is exactly one issue tracker for all 23 APIs, not separate issue trackers for each.

patflynn:
I think it's something we want eventually but I'm fine leaving it out for now.  Like Elliotte said, for this first batch of libraries they're all in the same github tracker and the phase 0 UX would probably only display the tracker info way before the user has had a chance to find any issues.

PR #452, opened 2017-09-18
Adds GAE_ENV and GAE_RUNTIME environment variables for local development.

### LINKED ISSUE #444: Add GAE_ENV and GAE_RUNTIME environment variables for local development
Please add the following environment variables to dev app server environment:
GAE_ENV = "localdev"
GAE_RUNTIME = "java7" or "java8" (depending on which runtime is used).

These variables are now available in GAE Java runtime, and should also be available in local development environment.

coollog:
Fixes #444 . 

Does this for CloudSdkAppEngineDevServer1. Will write the tests if this is the way we want to do it.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
@@ -110,7 +110,9 @@ public class CloudSdkAppEngineDevServer1 implements AppEngineDevServer {
       arguments.addAll(additionalArguments);
     }
 
-    if (isJava8(config.getServices())) {
+    boolean isJava8 = isJava8(config.getServices());
+
+    if (isJava8) {
       jvmArguments.add("-Duse_jetty9_runtime=true");
       jvmArguments.add("-D--enable_all_permissions=true");
       arguments.add("--no_java_agent");
@@ -131,7 +133,7 @@ public class CloudSdkAppEngineDevServer1 implements AppEngineDevServer {
           + Joiner.on(",").withKeyValueSeparator("=").join(appEngineEnvironment));
     }
 
-    String gaeRuntime = getGaeRuntimeJava(isJava8(config.getServices()));
+    String gaeRuntime = getGaeRuntimeJava(isJava8);
     appEngineEnvironment.putAll(getLocalAppEngineEnvironmentVariables(gaeRuntime));
 
     if (config.getEnvironment() != null) {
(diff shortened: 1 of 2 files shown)


coollog:
Okay, changed the tests to fit these new environment variables.

Also, I've noticed that `CloudSdkAppEngineDevServer1Test#testPrepareCommand_booleanFlags` and `CloudSdkAppEngineDevServer1Test#testPrepareCommand_noFlags` are exactly the same.

loosebazooka:
Yeah it's quite possible they're the same, I'll have to take a look and see if our intentions were initially different

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -242,6 +245,24 @@ boolean isJava8(List<File> services) {
...
   }
 
+  /**
+   * Gets a {@code Map<String, String>} of the environment variables for running the {@link
+   * AppEngineDevServer}.
+   *
+   * @param isJava8 if {@code true}, uses java8 as the GAE_RUNTIME; otherwise, uses java7
+   * @return {@code Map<String, String>} that maps from the environment variable name to its value
+   */
+  private static Map<String, String> getLocalAppEngineEnvironmentVariables(boolean isJava8) {
```
elharo: Might want to pass the runtime as a string. There are more than two.

src/test/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1Test.java
```diff
@@ -81,7 +80,11 @@
       Paths.get("src/test/resources/projects/Standard8Project2EnvironmentVariables");
   private final File java8Service2EnvVars = pathToJava8Service2WithEnvVars.toFile();
 
-  private final Map<String, String> environment = Maps.newHashMap();
+  // Environment variables included in running the dev server for Java 7/8 runtimes.
```
elharo: FYI, PRs are more reliable when tests are changed or added *before* the model code is changed. That is, first create a failing test. Then write only as much code as needed to make it pass.

coollog:
Okay, just changed getLocalAppEngineEnvironmentVariables to take the runtime ID as a string.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -131,6 +131,9 @@ public void run(RunConfiguration config) throws AppEngineException {
           + Joiner.on(",").withKeyValueSeparator("=").join(appEngineEnvironment));
     }
 
+    String gaeRuntime = getGaeRuntimeJava(isJava8(config.getServices()));
```
loosebazooka: This section kinda feels weird to me. Is there any way we can integrate this into line [113](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/1d6a525faa0f4f36a2e482b838d8e2394b62a2cb/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java#L113) where we're distinguishing between java7 and java8 already? That way we keep the java7/java8 branching in one place? It'll probably get rid of the `getGaeRuntimeJava` function.

coollog:
Okay, unit tests added!

PR #454, opened 2017-09-22
Downloader component of Managed Cloud SDK.

loosebazooka:
First part of the Cloud SDK manager code is the downloader. The first few PRs will all just be generic functionality to keep commits small. Once that's done, we'll stitch it all together.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -103,6 +103,12 @@
       <artifactId>javax.json</artifactId>
       <version>1.0.4</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-compress</artifactId>
+      <version>1.14</version>
+    </dependency>
+
     <!-- test dependencies -->
     <dependency>
       <groupId>junit</groupId>
@@ -127,6 +133,11 @@
        <artifactId>hamcrest-library</artifactId>
        <version>1.3</version>
        <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
+      <version>2.5</version>
     </dependency>
   </dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -103,11 +103,6 @@
       <artifactId>javax.json</artifactId>
       <version>1.0.4</version>
     </dependency>
-    <dependency>
-      <groupId>org.apache.commons</groupId>
-      <artifactId>commons-compress</artifactId>
-      <version>1.14</version>
-    </dependency>
 
     <!-- test dependencies -->
     <dependency>
@@ -133,11 +128,6 @@
        <artifactId>hamcrest-library</artifactId>
        <version>1.3</version>
        <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>commons-io</groupId>
-      <artifactId>commons-io</artifactId>
-      <version>2.5</version>
     </dependency>
   </dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -103,7 +103,6 @@
       <artifactId>javax.json</artifactId>
       <version>1.0.4</version>
     </dependency>
-
     <!-- test dependencies -->
     <dependency>
       <groupId>junit</groupId>
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloadProgressListener.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloadProgressListener.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.managedcloudsdk.internal.download;
+
+/** Listener on {@link Downloader} tasks. */
+public interface DownloadProgressListener {
+
+  /**
+   * Use this method to display progress, {@code percent downloaded = current * 100 / total}.
+   *
+   * @param lastChunk bytes read since last update.
+   * @param completed bytes downloaded so far.
+   * @param total file size in bytes.
+   */
+  void updateProgress(long lastChunk, long completed, long total);
+}
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloadProgressListener.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloadProgressListener.java
@@ -22,9 +22,9 @@ public interface DownloadProgressListener {
   /**
    * Use this method to display progress, {@code percent downloaded = current * 100 / total}.
    *
-   * @param lastChunk bytes read since last update.
-   * @param completed bytes downloaded so far.
-   * @param total file size in bytes.
+   * @param lastChunk bytes read since last update
+   * @param completed bytes downloaded so far
+   * @param total file size in bytes
    */
   void updateProgress(long lastChunk, long completed, long total);
 }
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloadProgressListener.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloadProgressListener.java
@@ -20,7 +20,7 @@ package com.google.cloud.tools.managedcloudsdk.internal.download;
 public interface DownloadProgressListener {
 
   /**
-   * Use this method to display progress, {@code percent downloaded = current * 100 / total}.
+   * Use this method to display progress: {@code percent downloaded = current * 100 / total}.
    *
    * @param lastChunk bytes read since last update
    * @param completed bytes downloaded so far
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/Downloader.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/Downloader.java
@@ -20,7 +20,7 @@ import java.nio.file.Path;
 import java.util.concurrent.Callable;
 
 /**
- * Downloader for downloading a single Cloud SDK archive, it is a callable we use in chained
+ * Downloader for downloading a single Cloud SDK archive. It is a callable we use in chained
  * asynchronous calls.
  */
 public interface Downloader extends Callable<Path> {
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/Downloader.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/Downloader.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.managedcloudsdk.internal.download;
+
+import java.nio.file.Path;
+import java.util.concurrent.Callable;
+
+/**
+ * Downloader for downloading a single Cloud SDK archive, it is a callable we use in chained
+ * asynchronous calls.
+ */
+public interface Downloader extends Callable<Path> {
+
+  /** Download and return a {@link Path} to the downloaded archive. */
+  @Override
+  Path call() throws Exception;
+}
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloaderFactory.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloaderFactory.java
@@ -20,21 +20,27 @@ import java.net.URL;
 import java.nio.file.Path;
 
 /** Downloader factory. */
-public class DownloaderFactory {
+public final class DownloaderFactory {
 
   private final String userAgentString;
 
+  /**
+   * Creates a new factory.
+   *
+   * @param userAgentString for server side tracking of clients downloading the sdk. For example,
+   *     "Cloud Tools for Eclipse" or "com.google.cloud.tools.appengine-maven-plguin".
+   */
   public DownloaderFactory(String userAgentString) {
     this.userAgentString = userAgentString;
   }
 
   /**
-   * Get a new {@link Downloader} implementation.
+   * Returns a new {@link Downloader} implementation.
    *
-   * @param source URL of file to download (remote).
-   * @param destination Path on local file system to save the file.
-   * @param downloadProgressListener Optional progress listener.
-   * @return location of downloaded file.
+   * @param source URL of file to download (remote)
+   * @param destination Path on local file system to save the file
+   * @param downloadProgressListener Optional progress listener
+   * @return a {@link Downloader} instance.
    */
   public Downloader newDownloader(
       URL source, Path destination, DownloadProgressListener downloadProgressListener) {
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloaderFactory.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloaderFactory.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.managedcloudsdk.internal.download;
+
+import java.net.URL;
+import java.nio.file.Path;
+
+/** Downloader factory. */
+public class DownloaderFactory {
+
+  private final String userAgentString;
+
+  public DownloaderFactory(String userAgentString) {
+    this.userAgentString = userAgentString;
+  }
+
+  /**
+   * Get a new {@link Downloader} implementation.
+   *
+   * @param source URL of file to download (remote).
+   * @param destination Path on local file system to save the file.
+   * @param downloadProgressListener Optional progress listener.
+   * @return location of downloaded file.
+   */
+  public Downloader newDownloader(
+      URL source, Path destination, DownloadProgressListener downloadProgressListener) {
+    return new StandardDownloader(source, destination, userAgentString, downloadProgressListener);
+  }
+}
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloaderFactory.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloaderFactory.java
@@ -40,7 +40,7 @@ public final class DownloaderFactory {
    * @param source URL of file to download (remote)
    * @param destination Path on local file system to save the file
    * @param downloadProgressListener Optional progress listener
-   * @return a {@link Downloader} instance.
+   * @return a {@link Downloader} instance
    */
   public Downloader newDownloader(
       URL source, Path destination, DownloadProgressListener downloadProgressListener) {
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
@@ -29,7 +29,7 @@ import java.nio.file.Path;
 /** Standard implementation of {@link Downloader}. */
 public final class StandardDownloader implements Downloader {
 
-  static final int BUFFER_SIZE = 10 * 1024;
+  static final int BUFFER_SIZE = 8 * 1024;
   private final URL address;
   private final Path destinationFile;
   private final String userAgentString;
@@ -84,18 +84,15 @@ public final class StandardDownloader implements Downloader {
             throw new InterruptedException("Downloader was interrupted.");
           }
           totalBytesRead += bytesRead;
-          if (totalBytesRead - lastUpdated > updateThreshold) {
+          long bytesSinceLastUpdate = totalBytesRead - lastUpdated;
+          if (totalBytesRead == contentLength || bytesSinceLastUpdate > updateThreshold) {
             if (downloadProgressListener != null) {
               downloadProgressListener.updateProgress(
-                  totalBytesRead - lastUpdated, totalBytesRead, contentLength);
+                  bytesSinceLastUpdate, totalBytesRead, contentLength);
             }
             lastUpdated = totalBytesRead;
           }
           out.write(buffer, 0, bytesRead);
-        }
-        if (downloadProgressListener != null && (lastUpdated != totalBytesRead)) {
-          downloadProgressListener.updateProgress(
-              totalBytesRead - lastUpdated, totalBytesRead, contentLength);
         }
       }
     }
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
@@ -62,8 +62,7 @@ public final class StandardDownloader implements Downloader {
       Files.createDirectories(destinationDir);
     }
 
-    URLConnection connection;
-    connection = address.openConnection();
+    URLConnection connection = address.openConnection();
     connection.setRequestProperty("User-Agent", userAgentString);
 
     try (BufferedOutputStream out =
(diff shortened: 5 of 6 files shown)


src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloadProgressListener.java
```diff
@@ -0,0 +1,30 @@
...
+
+package com.google.cloud.tools.managedcloudsdk.internal.download;
+
+/** Listener on {@link Downloader} tasks. */
+public interface DownloadProgressListener {
+
+  /**
+   * Use this method to display progress, {@code percent downloaded = current * 100 / total}.
+   *
+   * @param lastChunk bytes read since last update.
```
nkibler: nit: If there are no following statements after the initial clause, block tags should not have terminating punctuation:

```java
/**
 * ...
 *
 * @param lastChunk bytes read since last update
 * @param myVar an example with punctuation. The period only exists when there are more statements.
 */
```

Apply this here and everywhere else.

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloaderFactory.java
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.managedcloudsdk.internal.download;
+
+import java.net.URL;
+import java.nio.file.Path;
+
+/** Downloader factory. */
+public class DownloaderFactory {
```
nkibler: nit: Can this be final? The recommended advice is to "design and document for inheritance or else prohibit it." - http://go/java-practices/final#classes

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloaderFactory.java
```diff
@@ -0,0 +1,43 @@
...
+
+  private final String userAgentString;
+
+  public DownloaderFactory(String userAgentString) {
```
nkibler: Can you add a Javadoc with a `@param` tag for `userAgentString` describing what it is and maybe an example?

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloaderFactory.java
```diff
@@ -0,0 +1,43 @@
...
+    this.userAgentString = userAgentString;
+  }
+
+  /**
+   * Get a new {@link Downloader} implementation.
```
nkibler: nit: s/Get/Returns/

Use a declarative verb-phrase instead of an imperative statement for summary fragments. http://go/java-style#s7.2-summary-fragment

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
```diff
@@ -0,0 +1,110 @@
...
+import java.io.InputStream;
+import java.net.URL;
+import java.net.URLConnection;
+import java.nio.file.FileAlreadyExistsException;
+import java.nio.file.Files;
+import java.nio.file.NotDirectoryException;
+import java.nio.file.Path;
+
+/** Standard implementation of {@link Downloader}. */
+public class StandardDownloader implements Downloader {
```
nkibler: Ditto about final

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
```diff
@@ -0,0 +1,110 @@
...
+      DownloadProgressListener downloadProgressListener) {
+    this.address = source;
+    this.destinationFile = destinationFile;
+    this.userAgentString = userAgentString;
+    this.downloadProgressListener = downloadProgressListener;
+  }
+
+  @Override
+  public Path call() throws IOException, InterruptedException {
+
```
nkibler: nit: Remove extra newline here.

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
```diff
@@ -0,0 +1,110 @@
...
+    createDestinationDirectory();
+    return downloadUrl();
+  }
+
+  @VisibleForTesting
```
nkibler: Typically you should test via public APIs and don't create backdoors specifically for tests: http://go/unit-test-practices#public-apis

Since nobody will call these methods in isolation (they will always use `call()`, consider making these methods private and only invoke `call()` in your tests.

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
```diff
@@ -0,0 +1,110 @@
...
+    URLConnection conn;
+    conn = address.openConnection();
+    conn.setRequestProperty("User-Agent", userAgentString);
+
+    try (BufferedOutputStream out =
+        new BufferedOutputStream(Files.newOutputStream(destinationFile))) {
+      try (InputStream in = conn.getInputStream()) {
+
+        long contentLength = conn.getContentLengthLong();
+        long updateThreshold = contentLength / 100;
```
coollog: Maybe add some comments about this updating the progress listener for every 1% of progress?

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
```diff
@@ -0,0 +1,110 @@
...
+        while ((bytesRead = in.read(buffer)) != -1) {
+          if (Thread.currentThread().isInterrupted()) {
+            // throw exception here? todo?
+            throw new InterruptedException("Downloader was interrupted.");
+          }
+          totalBytesRead += bytesRead;
+          if (totalBytesRead - lastUpdated > updateThreshold) {
+            lastUpdated = totalBytesRead;
+            if (downloadProgressListener != null) {
+              downloadProgressListener.updateProgress(bytesRead, totalBytesRead, contentLength);
```
coollog: According the the annotation for `DownloadProgressListener#updateProgress`, the first parameter is `@param lastChunk bytes read since last update.` Should this be what the value of `totalBytesRead - lastUpdated` was rather than `bytesRead`?

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
```diff
@@ -0,0 +1,110 @@
...
+            }
+          }
+          out.write(buffer, 0, bytesRead);
+        }
+      }
```
coollog: It may be that `totalBytesRead - lastUpdated > 0` here, and the last few bytes are not added to the progress listener. Not sure if that's important, but might be good to have the progress listener finish with `completed == total` at the end.

src/test/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloaderTest.java
```diff
@@ -0,0 +1,179 @@
...
+      throws IOException, InterruptedException {
+
+    Path destination = tmp.getRoot().toPath().resolve("destinationFile");
+    Path testSourceFile = createTestRemoteResource(1);
+    URL fakeRemoteResource = testSourceFile.toUri().toURL();
+
+    StandardDownloader downloader =
+        new StandardDownloader(fakeRemoteResource, destination, null, null);
+
+    downloader.downloadUrl();
```
coollog: Wrap with a check for the return?

`Assert.assertEquals(destination, downloader.downloadUrl());`

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
loosebazooka: good call, I added that in late, and my test didn't really account for it. :\

src/test/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloaderTest.java
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  @Test
+  public void testDownloadURL() throws IOException, InterruptedException {
+    Path destination = tmp.getRoot().toPath().resolve("destinationFile");
+    Path testSourceFile = createTestRemoteResource(StandardDownloader.BUFFER_SIZE * 10);
```
coollog: As mentioned in the StandardDownloader file, changing this to `BUFFER_SIZE * 161` for example, breaks this test.

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloadProgressListener.java
```diff
@@ -0,0 +1,30 @@
...
+public interface DownloadProgressListener {
+
+  /**
+   * Use this method to display progress, {@code percent downloaded = current * 100 / total}.
+   *
+   * @param lastChunk bytes read since last update
+   * @param completed bytes downloaded so far
+   * @param total file size in bytes
+   */
+  void updateProgress(long lastChunk, long completed, long total);
```
coollog: Should the DownloadProgressListener be initializEd with the `total` file size, and `updateProgress` only updates with `lastChunk` and `completed`?

src/test/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloaderTest.java
```diff
@@ -0,0 +1,174 @@
...
+  @Test
+  public void testDownloadURL_createsNewDirectory() throws IOException, InterruptedException {
+    Path destination = tmp.getRoot().toPath().resolve("dir-to-create").resolve("destination-file");
+    Path testSourceFile = createTestRemoteResource(1);
+    URL fakeRemoteResource = testSourceFile.toUri().toURL();
+
+    StandardDownloader downloader =
+        new StandardDownloader(fakeRemoteResource, destination, null, null);
+
+    downloader.call();
```
coollog: I think my original comment got outdated.

Wrap with a check for the return?

`Assert.assertEquals(destination, downloader.call());`

Ditto below.

src/test/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloaderTest.java
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  @Test
+  public void testDownloadURL() throws IOException, InterruptedException {
+    Path destination = tmp.getRoot().toPath().resolve("destination-file");
+    Path testSourceFile = createTestRemoteResource(StandardDownloader.BUFFER_SIZE * 10);
```
coollog: I think my original comment got outdated:

As mentioned in the StandardDownloader file, changing this to BUFFER_SIZE * 161 for example, breaks this test.

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
```diff
@@ -0,0 +1,104 @@
...
+import java.net.URLConnection;
+import java.nio.file.FileAlreadyExistsException;
+import java.nio.file.Files;
+import java.nio.file.NotDirectoryException;
+import java.nio.file.Path;
+
+/** Standard implementation of {@link Downloader}. */
+public final class StandardDownloader implements Downloader {
+
+  static final int BUFFER_SIZE = 10 * 1024;
```
coollog: Is there a particular reason for using a buffer size of 10kB? Perhaps using a power of 2 might maximize possible compatibility with disk block and cache sizes, like (8kB or 16kB)?

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloadProgressListener.java
loosebazooka: This is more of a callback, so we don't initialize it.

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
```diff
@@ -0,0 +1,104 @@
...
+          if (totalBytesRead - lastUpdated > updateThreshold) {
+            if (downloadProgressListener != null) {
+              downloadProgressListener.updateProgress(
+                  totalBytesRead - lastUpdated, totalBytesRead, contentLength);
+            }
+            lastUpdated = totalBytesRead;
+          }
+          out.write(buffer, 0, bytesRead);
+        }
+        if (downloadProgressListener != null && (lastUpdated != totalBytesRead)) {
```
coollog: This can be combined to the one above by changing the expression on line 87 to like

`totalBytesRead == contentLength || totalBytesRead - lastUpdated > updateThreshold`

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
```diff
@@ -0,0 +1,104 @@
...
+        long lastUpdated = 0;
+        long totalBytesRead = 0;
+
+        int bytesRead;
+        while ((bytesRead = in.read(buffer)) != -1) {
+          if (Thread.currentThread().isInterrupted()) {
+            throw new InterruptedException("Downloader was interrupted.");
+          }
+          totalBytesRead += bytesRead;
+          if (totalBytesRead - lastUpdated > updateThreshold) {
```
coollog: Maybe have a variable like `lastChunk` set to `totalBytesRead - lastUpdated` to use here and at line 90.

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
loosebazooka: I think the sample I was working with used this. But you make a good point.

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
loosebazooka: yup, much cleaner.

pom.xml
```diff
@@ -128,6 +134,11 @@
        <version>1.3</version>
        <scope>test</scope>
     </dependency>
+    <dependency>
```
elharo: This is very rarely needed. Guava can do almost everything this can do. You only seem to use this in one place, so set scope to test if you must use it.

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/Downloader.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.managedcloudsdk.internal.download;
+
+import java.nio.file.Path;
+import java.util.concurrent.Callable;
+
+/**
+ * Downloader for downloading a single Cloud SDK archive, it is a callable we use in chained
```
elharo: , it is -->
. It is

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/Downloader.java
```diff
@@ -0,0 +1,31 @@
...
+ * asynchronous calls.
+ */
+public interface Downloader extends Callable<Path> {
```
elharo: Interface feels like overkill if there's only one implementation. Have you considered merging this with StandardDownloader?

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+    if (Files.exists(destinationFile)) {
+      throw new FileAlreadyExistsException(
+          "Cannot write to " + destinationFile + " because it already exists");
+    }
+    if (!Files.exists(destinationDir)) {
+      Files.createDirectories(destinationDir);
+    }
+
+    URLConnection conn;
```
elharo: one line;
conn --> connection

src/test/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloaderTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+  private Path createTestRemoteResource(long sizeInBytes) throws IOException {
+
+    Path testFile = tmp.newFile().toPath();
+    try (BufferedWriter writer = Files.newBufferedWriter(testFile, Charset.defaultCharset())) {
+      for (long i = 0; i < sizeInBytes; i++) {
+        writer.write('a');
+      }
+    }
+    return testFile;
```
elharo: flush the writer

src/test/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloaderTest.java
```diff
@@ -0,0 +1,178 @@
...
+    URL fakeRemoteResource = testSourceFile.toUri().toURL();
+    DownloadProgressListener mockListener = Mockito.mock(DownloadProgressListener.class);
+
+    StandardDownloader downloader =
+        new StandardDownloader(fakeRemoteResource, destination, null, mockListener);
+
+    Path downloaderDestination = downloader.call();
+    Assert.assertEquals(destination, downloaderDestination);
+
+    Assert.assertTrue(FileUtils.contentEquals(destination.toFile(), testSourceFile.toFile()));
```
elharo: could simply Files.readAllBytes and compare the arrays to avoid the dependency.

pom.xml
```diff
@@ -103,6 +103,12 @@
       <artifactId>javax.json</artifactId>
       <version>1.0.4</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-compress</artifactId>
```
elharo: This dependency doesn't seem to be used.

src/test/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloaderTest.java
loosebazooka: Writer flushes when closing : https://docs.oracle.com/javase/7/docs/api/java/io/Writer.html#close()

src/test/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloaderTest.java
loosebazooka: Makes sense.

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/Downloader.java
loosebazooka: Kind of a style thing I guess, it is consistent with the other code for this feature that will be uploaded.

pom.xml
loosebazooka: removed.

pom.xml
loosebazooka: It's used in another PR (extractor functionality) but snuck in here some how. I'll remove.

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloadProgressListener.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.managedcloudsdk.internal.download;
+
+/** Listener on {@link Downloader} tasks. */
+public interface DownloadProgressListener {
+
+  /**
+   * Use this method to display progress, {@code percent downloaded = current * 100 / total}.
```
elharo: nit: progress, --> progress:

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/DownloaderFactory.java
```diff
@@ -0,0 +1,49 @@
...
+    this.userAgentString = userAgentString;
+  }
+
+  /**
+   * Returns a new {@link Downloader} implementation.
+   *
+   * @param source URL of file to download (remote)
+   * @param destination Path on local file system to save the file
+   * @param downloadProgressListener Optional progress listener
+   * @return a {@link Downloader} instance.
```
elharo: no period since not a complete sentence

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+    if (Files.exists(destinationFile)) {
+      throw new FileAlreadyExistsException(
+          "Cannot write to " + destinationFile + " because it already exists");
+    }
+    if (!Files.exists(destinationDir)) {
+      Files.createDirectories(destinationDir);
+    }
+
+    URLConnection connection;
```
elharo: merge with next line

src/test/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloaderTest.java
```diff
@@ -0,0 +1,177 @@
...
+  public void testValidateAndCreateFiles_targetDirectoryIsFile()
+      throws IOException, InterruptedException {
+    Path fileWithBadParent = tmp.newFile().toPath().resolve("xyz");
+
+    StandardDownloader downloader = new StandardDownloader(null, fileWithBadParent, null, null);
+    try {
+      downloader.call();
+      Assert.fail("NotDirectoryException expected but not thrown");
+    } catch (NotDirectoryException nde) {
+      // pass
```
elharo: you can test the exception message here

src/test/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloaderTest.java
```diff
@@ -0,0 +1,177 @@
...
+  public void testValidateAndCreateFiles_fileAlreadyExists()
+      throws IOException, InterruptedException {
+    Path fileThatExists = tmp.newFile().toPath();
+
+    StandardDownloader downloader = new StandardDownloader(null, fileThatExists, null, null);
+    try {
+      downloader.call();
+      Assert.fail("FileAlreadyExists expected but not thrown");
+    } catch (FileAlreadyExistsException faee) {
+      // pass
```
elharo: you can test the exception message here

src/test/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloaderTest.java
```diff
@@ -0,0 +1,177 @@
...
+    URLStreamHandler testHandler =
+        new URLStreamHandler() {
+          @Override
+          protected URLConnection openConnection(URL url) throws IOException {
+            return mockConnection;
+          }
+        };
+
+    // create a URL with a custom streamHandler so we can get our mock connection
+    URL testURL = new URL("", "", 80, "", testHandler);
```
elharo: testUrl

src/test/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloaderTest.java
```diff
@@ -0,0 +1,177 @@
...
+    StandardDownloader downloader =
+        new StandardDownloader(testURL, destination, "test-user-agent", null);
+
+    try {
+      downloader.call();
+    } catch (Exception e) {
```
elharo: e --> ex

src/test/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloaderTest.java
loosebazooka: done

src/test/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloaderTest.java
loosebazooka: done.

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
```diff
@@ -0,0 +1,100 @@
...
+        byte[] buffer = new byte[BUFFER_SIZE];
+        long lastUpdated = 0;
+        long totalBytesRead = 0;
+
+        int bytesRead;
+        while ((bytesRead = in.read(buffer)) != -1) {
+          if (Thread.currentThread().isInterrupted()) {
+            throw new InterruptedException("Downloader was interrupted.");
+          }
+          totalBytesRead += bytesRead;
```
coollog: Since this loop does two things (writing the read bytes to the output file, and updating the progress bar), it might be good to separate out the progress update into another function to make this function easier to read, like move lines 74, 77-78, and 85-93 to a helper function like `updateProgress(bytesRead, contentLength)`, and maintain the state in member variables or a separate download state DDO.

loosebazooka: moved stuff around a little, adding a second function would probably require us to save state across calls, so I avoided that, hopefully it's easier to read now.

loosebazooka:
@coollog @nkibler anything else? If not, gonna submit.

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
```diff
@@ -0,0 +1,103 @@
...
+        long totalBytesRead = 0;
+
+        while ((bytesRead = in.read(buffer)) != -1) {
+          if (Thread.currentThread().isInterrupted()) {
+            throw new InterruptedException("Downloader was interrupted.");
+          }
+          out.write(buffer, 0, bytesRead);
+
+          // update progress
+          if (downloadProgressListener != null) {
```
coollog: I was thinking that it might make more sense to have this thresholding logic go in whatever is listening on progress updates or else this logic would have to repeated in every implementation of Downloader, but this can just be refactored out later.

loosebazooka: Oh I see, yeah, I don't anticipate any other kind of downloader at the moment.

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/download/StandardDownloader.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+    if (Files.exists(destinationFile)) {
+      throw new FileAlreadyExistsException(
+          "Cannot write to " + destinationFile + " because it already exists");
+    }
+    if (!Files.exists(destinationDir)) {
+      Files.createDirectories(destinationDir);
+    }
+
+    URLConnection connection = address.openConnection();
```
briandealwis: Could you extract the open-URL-InputStream to a protected method? I suspect we may need to do something different in Eclipse to support its proxying configuration.

loosebazooka: #457 will add in followup.

PR #455, opened 2017-09-25
Extractor component of Managed Cloud SDK.

loosebazooka:
Second part of the Cloud SDK manager code is the extractor with support for `tar.gz` and `zip` files.


loosebazooka:
extractor component.

src/test/java/com/google/cloud/tools/managedcloudsdk/internal/extract/ExtractorFactoryTest.java
```diff
@@ -0,0 +1,56 @@
...
+  @Test
+  public void testNewExtractor_unknownArchiveType() throws IOException {
+    // make sure out check starts from end of filename
+    Path archive = tmp.newFile("test-bad.tar.gz.zip.bad").toPath();
+
+    try {
+      new ExtractorFactory().newExtractor(archive, null, null);
+      Assert.fail("UnknownArchiveTypeException expected but not thrown");
+    } catch (UnknownArchiveTypeException e) {
+      // pass
```
coollog: Test the exception message here?

`"Unknown archive : " + archive.toString()`

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/extract/TarGzExtractor.java
```diff
@@ -0,0 +1,82 @@
...
+            }
+          }
+        } else {
+          // we don't know what kind of entry this is (we only process directories and files).
+        }
+      }
+    }
+
+    // this is convention
+    return destination.resolve("google-cloud-sdk");
```
coollog: Should there be a check that this directory actually got extracted?

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/extract/TarGzExtractor.java
```diff
@@ -0,0 +1,82 @@
...
+ * <p>NOTE: this does not handle links or symlinks or any other kind of special types in the tar. It
+ * will only create files and directories.
+ */
+public class TarGzExtractor implements Extractor {
+
+  private final Path archive;
+  private final Path destination;
+  private final ExtractorMessageListener listener;
+
+  TarGzExtractor(Path archive, Path destination, ExtractorMessageListener listener) {
```
coollog: Add comment: `/** Use {@link ExtractorFactory} to instantiate. */` ?

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/extract/ZipExtractor.java
```diff
@@ -0,0 +1,71 @@
...
+  private final ExtractorMessageListener listener;
+
+  ZipExtractor(Path archive, Path destination, ExtractorMessageListener listener) {
+    this.archive = archive;
+    this.destination = destination;
+    this.listener = listener;
+  }
+
+  @Override
+  public Path call() throws IOException {
```
coollog: It seems that there is a lot of boilerplate code that is the same as in TarGzExtractor. Perhaps have an abstract skeletal implementation for both to inherit from? But probably unnecessary unless more archive types would be added.

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/extract/TarGzExtractor.java
```diff
@@ -0,0 +1,82 @@
...
+    GzipCompressorInputStream gzipIn = new GzipCompressorInputStream(Files.newInputStream(archive));
+    try (TarArchiveInputStream in = new TarArchiveInputStream(gzipIn)) {
+      TarArchiveEntry entry;
+      while ((entry = in.getNextTarEntry()) != null) {
+        final Path entryPath = destination.resolve(entry.getName());
+        if (listener != null) {
+          listener.message(entryPath.toString());
+        }
+        if (entry.isDirectory()) {
+          if (!Files.exists(entryPath)) {
```
coollog: Why is this checking if the `entryPath` doesn't exist? Shouldn't that never happen, and if it did happen, wouldn't it mean something went wrong?

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/extract/ZipExtractor.java
loosebazooka: seemed fine before, but I think I'm going to restructure this if we're adding more checks like "check if actually extracted"

src/test/java/com/google/cloud/tools/managedcloudsdk/internal/extract/GenericArchivesVerifier.java
```diff
@@ -0,0 +1,76 @@
...
+    PosixFileAttributeView allAttributesFile2 =
+        Files.getFileAttributeView(file2, PosixFileAttributeView.class);
+    Assert.assertThat(
+        allAttributesFile2.readAttributes().permissions(),
+        Matchers.containsInAnyOrder(PosixFilePermission.values()));
+  }
+
+  public static void assertListenerReceivedExtractionMessages(
+      ExtractorMessageListener listener, Path testDir, Path archive) {
+    assert MockUtil.isMock(listener);
```
briandealwis: This seems odd?

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/extract/ZipExtractorProvider.java
```diff
@@ -0,0 +1,59 @@
...
+        final Path entryPath = destination.resolve(entry.getName());
+        if (extractorMessageListener != null) {
+          extractorMessageListener.message(entryPath.toString());
+        }
+        if (entry.isDirectory()) {
+          if (!Files.exists(entryPath)) {
+            Files.createDirectories(entryPath);
+          }
+        } else {
+          try (OutputStream out = new BufferedOutputStream(Files.newOutputStream(entryPath))) {
```
briandealwis: Don't you need to check for setting the executable bits here?  Otherwise `gcloud` and `dev_appserver.py` won't be write.  Looks like you might be able to use `ZipArchiveEntry#getUnixMode()`?

loosebazooka: I had the code in there, but ran into some problems and since we only use zip for windows I avoided it. Lemme try again for completeness.

loosebazooka: found out why for anyone interested, from the source : 
```
    /**
     * Retrieves the external file attributes.
     *
     * <p><b>Note</b>: {@link ZipArchiveInputStream} is unable to fill
     * this field, you must use {@link ZipFile} if you want to read
     * entries using this attribute.</p>
     *
     * @return the external file attributes
     */
    public long getExternalAttributes() {
        return externalAttributes;
    }
```

I have to use `ZipFile` instead of `ZipArchiveInputStream` it appears to have to something to do with being able to seek on the file, but I don't know for sure.

briandealwis: Ah, ok.  Definitely worth documenting.

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/extract/ExtractorFactory.java
```diff
@@ -0,0 +1,52 @@
...
+package com.google.cloud.tools.managedcloudsdk.internal.extract;
+
+import java.nio.file.Path;
+
+/** {@link Extractor} Factory. */
+public final class ExtractorFactory {
+
+  /**
+   * Creates a new extractor based on filetype. Filetype determination is based on the filename
+   * string, this method makes no attempt to validate the actual file contents to verify they are
```
elharo: string. This
delete "actual" and "indeed of the"

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/extract/ExtractorFactory.java
```diff
@@ -0,0 +1,52 @@
...
+   * indeed of the type defined by the file extension.
+   *
+   * @param archive The archive to extract
+   * @param destination The destination folder for extracted files
+   * @param extractorMessageListener An listener for extraction messages
```
elharo: a listener

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/extract/ExtractorFactory.java
```diff
@@ -0,0 +1,52 @@
...
+   * @return {@link ConfigurableExtractor} with {@link TarGzExtractorProvider} for ".tar.gz", {@link
+   *     ZipExtractorProvider} for ".zip"
+   * @throws UnknownArchiveTypeException if not ".tar.gz" or ".zip"
+   */
+  public Extractor newExtractor(
+      Path archive, Path destination, ExtractorMessageListener extractorMessageListener)
+      throws UnknownArchiveTypeException {
+
+    if (archive.toString().endsWith(".tar.gz")) {
```
elharo: Path has an endsWith method. No need to call toString

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/extract/ExtractorProvider.java
```diff
@@ -0,0 +1,35 @@
...
+
+/** Provide a archive extractor implementation. */
+public interface ExtractorProvider {
+
+  /**
+   * Extracts a single file archive into target destination folder.
+   *
+   * @param archive The archive to extract
+   * @param destination The destination folder for extracted files
+   * @param extractorMessageListener An listener for extraction messages
```
elharo: a listener

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/extract/TarGzExtractorProvider.java
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * {@link ExtractorProvider} implementation for *.tar.gz files.
+ *
+ * <p>NOTE: this does not handle links or symlinks or any other kind of special types in the tar. It
+ * will only create files and directories.
+ */
+public final class TarGzExtractorProvider implements ExtractorProvider {
+
+  /** Only instantiated in {@link ExtractorFactory}. */
+  @VisibleForTesting
```
elharo: don't need @VisibleForTesting

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/extract/TarGzExtractorProvider.java
```diff
@@ -0,0 +1,75 @@
...
+        if (extractorMessageListener != null) {
+          extractorMessageListener.message(entryTarget.toString());
+        }
+        if (entry.isDirectory()) {
+          if (!Files.exists(entryTarget)) {
+            Files.createDirectories(entryTarget);
+          }
+        } else if (entry.isFile()) {
+          try (OutputStream out = new BufferedOutputStream(Files.newOutputStream(entryTarget))) {
+            IOUtils.copy(in, out);
```
elharo: or ByteStreams.copy

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/extract/TarGzExtractor.java
loosebazooka: I think directories can be out of order.
For example, the following situation where a sub dir is created before a parent, but all entries are in the tar :
```
$ tree .
.
└── a
    ├── b
    │   └── c.txt
    └── d.txt

$ tar -cf test.tar a/b/ a/

$ tar -tf test.tar 
a/b/
a/b/c.txt
a/
a/b/
a/b/c.txt
a/d.txt
```

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/extract/ExtractorFactory.java
loosebazooka: I believe this is actually related to path elements, so it wont work for us : https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#endsWith-java.lang.String-

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/extract/TarGzExtractorProvider.java
loosebazooka: Already using commons-compress in this class.

src/main/java/com/google/cloud/tools/managedcloudsdk/internal/extract/TarGzExtractor.java
coollog: Hmm in that case, if a file came before its directory, like this:

```
$ tar -cf test.tar a/b/c.txt a/b/ a/
$ tar -tf test.tar 
a/b/c.txt
a/b/
...
```

Then the code breaks with a NoSuchFileException.

loosebazooka: ugh, stupid tar, okay yeah I'll handle that.

loosebazooka:
@elharo, correct, needs to fixed #472

PR #463, opened 2017-10-03
bundle libraries.json

elharo:
@loosebazooka @chanseokoh

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -331,6 +331,13 @@
         </plugin>
       </plugins>
     </pluginManagement>
+
+    <resources>
+      <resource>
+        <directory>src/main/java/com/google/cloud/tools/libraries</directory>
+      </resource>
+    </resources>
+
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -331,6 +331,13 @@
         </plugin>
       </plugins>
     </pluginManagement>
+
+    <resources>
```
loosebazooka: this can't go in src/main/resources?

elharo: It could. It doesn't have to.

loosebazooka: I think, as much as we can, we should follow the standard directory format for maven : https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html

Given that resources is already defined, can you move the file into `src/main/resources`?

loosebazooka: it would remove extra configuration from pom.xml.

PR #464, opened 2017-10-03
Delete proof-of-concept libraries.xml

chanseokoh:


loosebazooka:
Looks like you need to remove the test as well?

PR #465, opened 2017-10-03
Release 0.3.8

elharo:
@loosebazooka @chanseokoh @coollog

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
 
   <groupId>com.google.cloud.tools</groupId>
   <artifactId>appengine-plugins-core</artifactId>
-  <version>0.3.8-SNAPSHOT</version>
+  <version>0.3.8</version>
 
   <name>App Engine Plugins Core Library</name>
   <description>
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
 
   <groupId>com.google.cloud.tools</groupId>
   <artifactId>appengine-plugins-core</artifactId>
-  <version>0.3.8</version>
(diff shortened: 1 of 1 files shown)


loosebazooka:
I don't fully understand why we need this resource directory in the source directory? see : https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/463/files

chanseokoh:
@loosebazooka I agree that the JSON is not a Java source file, so it shouldn't be under `src/main/java`. (Anyway, it has been like that for a while.)

elharo:
It doesn't have to be under src/main/java. It can be there, however. Eclipse projects, for example, put numerous non-Java files directly into their package directories. It's a little easier to edit things if all the related files are together rather than in separate trees.

chanseokoh:
There are also some downsides.
- The directory is deep inside the directory hierarchy. It's easy to forget it's a resource directory. If we start adding resource directories in this fashion, we may lose track easily.
- But more importantly, this prevents creating Java packages with the prefix of `com/google/cloud/tools/libraries`. I did some testing, and looks like anything put under `src/main/java/com/google/cloud/tools/libraries` (including `.java` files and subdirectories) will be packaged into the final JAR.

elharo:
@chanseokoh's last comment is convincing. That said, that is a separate issue which I've filed as #467.

Unless you think the code is currently broken or that the external interface will change (i.e. the path by which we'll load the resource will be different), that should not block us from moving forward with this PR.

chanseokoh:
Need to merge with master.

loosebazooka:
I would probably kill this PR, delete the tag, and remake with a new one.

PR #468, opened 2017-10-03
Move libraries.json to src/main/resources

### LINKED ISSUE #467: move libraries.json to src/main/resources

chanseokoh:
Fixes #467.


loosebazooka:
Don't you need to move the actual file?

src/test/java/com/google/cloud/tools/libraries/LibrariesTest.java
```diff
@@ -42,8 +42,7 @@
   @Before
   public void parseJson() throws FileNotFoundException {
     JsonReaderFactory factory = Json.createReaderFactory(null);
-    InputStream in =
-        new FileInputStream("src/main/java/com/google/cloud/tools/libraries/libraries.json");
+    InputStream in = new FileInputStream("src/main/resources/libraries.json");
```
loosebazooka: Question : Would you like it to be in the same package path as the class instead of in the base dir of resources?

I believe you can access the resources via the `ClassLoader.getResourceAsStream()` mechanism as an extra way to make sure we're including it in the classpath?

chanseokoh: - I am not sure. The base directory looks fine to me.
- I'll use `getResourceAsStream()`.

PR #469, opened 2017-10-03
Metadata

elharo:
@chanseokoh @loosebazooka Update library versions and fix a couple of markup glitches


loosebazooka:
I'll defer to @chanseokoh here. I haven't been following these CLs closely.

chanseokoh:
I'm not following either, but looks good. BTW, there is a formatting error that's causing a build error.

PR #473, opened 2017-10-05
Enables the google-java-format plugin

nkibler:


loosebazooka:
Waiting on check to see what happens if user doesn't have plugin installed.

nkibler:
Confirmed that nothing happens without the `google-java-format` plugin installed; the IDE simply uses its default formatter instead.

PR #477, opened 2017-10-11
Adds utility classes around the libraries.json file

nkibler:
This will either be here, next to the `libraries.json` file and synchronized with the changes made to the JSON, or it would need to live independently in the IntelliJ plugin, so I opted to place it here. Feel free to suggest different approaches as well.

### DIFF
--- a/src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
+++ b/src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.libraries;
+
+import com.google.cloud.tools.libraries.json.CloudLibrary;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
+import com.google.gson.stream.JsonReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.lang.reflect.Type;
+import java.util.List;
+
+/** Returns helpful metadata for supported Google Cloud libraries. */
+public final class CloudLibraries {
+
+  private static final String LIBRARIES_JSON = "libraries.json";
+
+  private final String librariesJsonPath;
+
+  @VisibleForTesting
+  CloudLibraries(String librariesJsonPath) {
+    this.librariesJsonPath = librariesJsonPath;
+  }
+
+  /**
+   * Returns the list of {@link CloudLibrary} objects deserialized from the {@code libraries.json}
+   * file.
+   *
+   * @throws IOException if there was a problem reading the {@code libraries.json} file
+   */
+  public static List<CloudLibrary> getCloudLibraries() throws IOException {
+    return new CloudLibraries(LIBRARIES_JSON).getLibraries();
+  }
+
+  @VisibleForTesting
+  List<CloudLibrary> getLibraries() throws IOException {
+    try (InputStream inputStream = CloudLibraries.class.getResourceAsStream(librariesJsonPath)) {
+      if (inputStream == null) {
+        throw new IOException("Resource not found: " + librariesJsonPath);
+      }
+
+      InputStreamReader reader = new InputStreamReader(inputStream);
+      JsonReader jsonReader = new JsonReader(reader);
+      Type listType = new TypeToken<List<CloudLibrary>>() {}.getType();
+      return new Gson().fromJson(jsonReader, listType);
+    }
+  }
+}
--- a/src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
+++ b/src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
@@ -53,7 +53,7 @@ public final class CloudLibraries {
   List<CloudLibrary> getLibraries() throws IOException {
     try (InputStream inputStream = CloudLibraries.class.getResourceAsStream(librariesJsonPath)) {
       if (inputStream == null) {
-        throw new IOException("Resource not found: " + librariesJsonPath);
+        throw new AssertionError("Resource not found: " + librariesJsonPath);
       }
 
       InputStreamReader reader = new InputStreamReader(inputStream);
--- a/src/main/java/com/google/cloud/tools/libraries/json/CloudLibrary.java
+++ b/src/main/java/com/google/cloud/tools/libraries/json/CloudLibrary.java
@@ -34,14 +34,6 @@ public final class CloudLibrary {
   /** Use {@link #parse(String)} instead. */
   private CloudLibrary() {}
 
-  /**
-   * Parses the given JSON representation of a {@link CloudLibrary} and returns a new instance
-   * deserialized from the given JSON.
-   */
-  public static CloudLibrary parse(String json) {
-    return new Gson().fromJson(json, CloudLibrary.class);
-  }
-
   /** Returns the name of this library. */
   @Nullable
   public String getName() {
--- a/src/main/java/com/google/cloud/tools/libraries/json/CloudLibrary.java
+++ b/src/main/java/com/google/cloud/tools/libraries/json/CloudLibrary.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.libraries.json;
+
+import com.google.gson.Gson;
+import java.util.List;
+import javax.annotation.Nullable;
+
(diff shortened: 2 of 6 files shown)


etanshaul:
I think it makes sense to put this here

src/main/java/com/google/cloud/tools/libraries/json/CloudLibrary.java
```diff
@@ -0,0 +1,86 @@
...
+  private List<CloudLibraryClient> clients;
+
+  /** Use {@link #parse(String)} instead. */
+  private CloudLibrary() {}
+
+  /**
+   * Parses the given JSON representation of a {@link CloudLibrary} and returns a new instance
+   * deserialized from the given JSON.
+   */
+  public static CloudLibrary parse(String json) {
```
etanshaul: whats the use of this method in practice? Given the json file, wouldn't the user first parse it into a `List<CloudLibrary>`?

nkibler: Yeah that's true, my original intention was to provide a way to parse a single `CloudLibrary` so it's not strictly limited to the existing content of the `libraries.json` file, but it really isn't necessary (and it's already pretty limited to the `libraries.json` anyway).

PTAL, I've removed this method.

src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
```diff
@@ -0,0 +1,65 @@
...
+import com.google.gson.reflect.TypeToken;
+import com.google.gson.stream.JsonReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.lang.reflect.Type;
+import java.util.List;
+
+/** Returns helpful metadata for supported Google Cloud libraries. */
+public final class CloudLibraries {
```
elharo: This is going to have a name conflict with existing classes in CT4E.

etanshaul:
lgtm pending fixing the build

src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
```diff
@@ -0,0 +1,65 @@
...
+  @VisibleForTesting
+  CloudLibraries(String librariesJsonPath) {
+    this.librariesJsonPath = librariesJsonPath;
+  }
+
+  /**
+   * Returns the list of {@link CloudLibrary} objects deserialized from the {@code libraries.json}
+   * file.
+   *
+   * @throws IOException if there was a problem reading the {@code libraries.json} file
```
elharo: Since this is bundled in the jar this is a case where a runtime exception is more appropriate.

src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
```diff
@@ -0,0 +1,65 @@
...
+  List<CloudLibrary> getLibraries() throws IOException {
+    try (InputStream inputStream = CloudLibraries.class.getResourceAsStream(librariesJsonPath)) {
+      if (inputStream == null) {
+        throw new IOException("Resource not found: " + librariesJsonPath);
+      }
+
+      InputStreamReader reader = new InputStreamReader(inputStream);
+      JsonReader jsonReader = new JsonReader(reader);
+      Type listType = new TypeToken<List<CloudLibrary>>() {}.getType();
+      return new Gson().fromJson(jsonReader, listType);
```
elharo: This tightly couples the JSON data to the Java class, while still defining each independently. That's where brittleness comes from.

src/test/java/com/google/cloud/tools/libraries/json/CloudLibraryTest.java
```diff
@@ -0,0 +1,169 @@
...
+  private static final String CLIENT_INFO_TIP = "info tip";
+  private static final String CLIENT_LAUNCH_STAGE = "alpha";
+  private static final String CLIENT_SOURCE = "http://client-source";
+  private static final String CLIENT_LANGUAGE_LEVEL = "1.8.0";
+  private static final String CLIENT_MAVEN_GROUP_ID = "my-group";
+  private static final String CLIENT_MAVEN_ARTIFACT_ID = "my-artifact";
+  private static final String CLIENT_MAVEN_VERSION = "1.0.0-alpha";
+
+  @Test
+  public void parse_withFullyPopulatedJson_returnsPopulatedObject() {
```
elharo: Since the json file is bundled into the jar and only read from there, we can use tests for the actual JSON data, not mock data. I.e. we're not trying to read any arbitrary external document but one specific document from the same repo.

nkibler: You could, but this is meant to be a lightweight unit test that has no external dependencies. Reading from the filesystem would make it much slower and it would test the integration between the model and the file. Read more at http://go/smalltest for rationale.

src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
nkibler: Maybe I'm missing something, but the point of this class is to represent the data in the JSON file in a structured way. The file and this class are designed to be coupled together, but I disagree that the coupling is so tight that it would cause *brittleness* with changes made to either. If you change the JSON (e.g. new keys, remove old keys, etc.), the model class can still be built without error. What alternative would you suggest?

src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
nkibler: Is the package the same? If the package isn't the same, then I don't believe there's an issue. If CT4E migrates to use these classes, I doubt it would reference both at the same time, requiring it to fully specify the class references.

src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
nkibler: Could you explain more? IMO throwing an unchecked exception here does not give the caller an opportunity to gracefully degrade their user experience if this happens to fail (which could happen even if `libraries.json` does exist in the JAR). Sure they could catch any `Throwable` or `RuntimeException` coming from this method, but the context isn't as clear and it's not enforced by the method signature.

src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
elharo: The package isn't the same, but it's still inconvenient to have classes with the same name, if not exactly impossible.

src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
elharo: Checked exceptions are for problems external to the program such as an IOException or syntax error in an external document.
Runtime exceptions are for program bugs.

This is a really unusual case where the JSON is not an external file but program source code bundled into and read from the jar. A failure here is a bug in the program. Thus runtime exception. 

Most similar cases should indeed be checked exceptions.

src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
elharo: Think of it this way: ask for what you need, not for what the supplier has. CloudLibraries in CT4E takes a very different approach.

src/test/java/com/google/cloud/tools/libraries/json/CloudLibraryTest.java
elharo: The JSON is internal program source code, not an external dependency. To support the tests you're creating code paths to handle things that don't happen and making the code more general and complex than it needs to be.

src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
etanshaul: @elharo I'm also a little confused by what you are suggesting. Are you suggesting eclipse and intellij duplicate this logic? I assume CT4E has logic to parse this json file, just like CT4IJ will need.

src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
etanshaul: I think that if we can agree to share the parsing logic across clients in appengine-plugins-core then this naming issue will go away.

src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
chanseokoh: > If you change the JSON (e.g. new keys, remove old keys, etc.), the model class can still be built without error.

Nit-picking: it's not always the case. For example, if you change an int property to a string, or a string property to an array, `fromJson()` will throw a (baffling) malformed-JSON runtime exception. However, I admit this kind of changes would be rare. Also, if the point of this class is to be tightly coupled together by its design, I guess talking about brittleness is irrelevant anyway IMO.

FYI, the approach in Eclipse is (conceptually and supposed to be) like querying using JsonPath (an XPath counterpart for JSON) or such (although in reality, this is implemented in a naive way).

src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
```diff
@@ -0,0 +1,65 @@
...
+   */
+  public static List<CloudLibrary> getCloudLibraries() throws IOException {
+    return new CloudLibraries(LIBRARIES_JSON).getLibraries();
+  }
+
+  @VisibleForTesting
+  List<CloudLibrary> getLibraries() throws IOException {
+    try (InputStream inputStream = CloudLibraries.class.getResourceAsStream(librariesJsonPath)) {
+      if (inputStream == null) {
+        throw new IOException("Resource not found: " + librariesJsonPath);
```
patflynn: this on the other hand does look like a misconfiguration of the source if it occurs and should be an assertion error or runtime exception.

src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
patflynn: @nkibler does have a point that a failure here doesn't always indicate a bug.  It could be an IO issue of some sort.

src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
patflynn: I don't object to @nkibler's approach here.  We're already happily using GSON databinding with the CloudSdkComponent and in IntelliJ.   There can be issues using object mapping at scale on large class hierarchies.  I don't see a risk of that happening here.

nkibler: @chanseokoh that's correct, it will break if the types are incompatible. I added a unit test to explicitly call this out for clarity.

src/main/java/com/google/cloud/tools/libraries/CloudLibraries.java
nkibler: PTAL, I've changed this to an `AssertionError`.

PR #481, opened 2017-10-24
Adds script to do release setup.

### LINKED ISSUE #479: Add script to do release setup.
Currently, there are many manual commands that need to be run to push a new release branch and tag. This can be simplified into a simple script that can be run:

`./release.sh <release version> <next version>`

For example, to release version `1.5.7`, just run the command:

`./release.sh 1.5.7 1.5.8`

coollog:
Fixes #479 .


loosebazooka:
presumably there's no maven mechanism for this?

coollog:
This is essentially what we have for our manual instructions, just put in script for so we can just run it with a version number. It uses the `mvn versions:set versions:commit` command.

coollog:
Added the newline at the end of the file.

PR #483, opened 2017-10-24
Add service name to json client lib metadata

### LINKED ISSUE #482: Client library json metadata missing service name
Each library needs a service name identifier:

image

This is used by (at least) the service management API enable calls for programmatically enabling the API.

etanshaul:
fixes #482


etanshaul:
note: I didn't replace the existing ID field with this since @elharo mentioned eclipse is already key'ing off of this.

etanshaul:
@elharo I added tests to cover presence and duplicates. I don't think we should assume format so I didn't add a test for that.

PR #484, opened 2017-10-24
Remove unnecessary Callable interface. Fortify checks and cleanup.

loosebazooka:
It turns out chaining callables wasn't the best idea, easier to just async once. This CL also adds some more checks and cleanup for different parts.

### DIFF
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/install/Downloader.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/install/Downloader.java
@@ -22,16 +22,13 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 import java.net.URLConnection;
+import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
-import java.util.concurrent.Callable;
 
-/**
- * Downloader for downloading a single Cloud SDK archive. It is a callable we use in chained
- * asynchronous calls.
- */
-final class Downloader implements Callable<Path> {
+/** Downloader for downloading a single Cloud SDK archive. */
+final class Downloader {
 
   static final int BUFFER_SIZE = 8 * 1024;
   private final URL address;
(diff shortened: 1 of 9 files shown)


loosebazooka:
I think I may have overcomplicated the downloader. I'll probably move the already "downloaded check" out of there in a subsequent PR.

src/main/java/com/google/cloud/tools/io/FileUtil.java
```diff
@@ -76,4 +77,38 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs)
...
+   * @param target directory to delete
+   * @throws IOException if target is not a directory or underlying file mechanisms fail
+   */
+  public static void deleteDirectory(final Path target) throws IOException {
+    if (!Files.isDirectory(target)) {
+      throw new NotDirectoryException(target.toString());
+    }
+    Files.walkFileTree(
+        target,
+        new SimpleFileVisitor<Path>() {
```
elharo: separate outer class please.
also see #314

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Downloader.java
```diff
@@ -48,24 +44,40 @@
...
-    try (BufferedOutputStream out =
-        new BufferedOutputStream(
-            Files.newOutputStream(destinationFile, StandardOpenOption.CREATE_NEW))) {
-      try (InputStream in = connection.getInputStream()) {
-        messageListener.message("Downloading " + address);
+    try (InputStream in = connection.getInputStream()) {
+      long contentLength = connection.getContentLengthLong();
+
+      if (Files.isRegularFile(destinationFile)) {
+        if (Files.size(destinationFile) == contentLength) {
```
elharo: what if it has same length and different content?

loosebazooka: Yeah this was initially supposed to get checksums but we don't have that. Maybe for now I'll just redownload.

src/main/java/com/google/cloud/tools/io/FileUtil.java
loosebazooka: #314 was related to putting eclipse functionality in appengine-plugins-core, and even though this is now used in plugins-core, perhaps the point is still valid and these should just be moved into the classes using them and FileUtil should be removed.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Extractor.java
```diff
@@ -42,24 +44,60 @@
...
   }
 
   @VisibleForTesting
   ExtractorProvider getExtractorProvider() {
     return extractorProvider;
   }
+
+  private void cleanUp(final Path target) throws IOException {
+    Preconditions.checkArgument(Files.isDirectory(target));
+    Files.walkFileTree(
```
coollog: What about using the google commons function? https://google.github.io/guava/releases/21.0/api/docs/com/google/common/io/MoreFiles.html#deleteDirectoryContents-java.nio.file.Path-com.google.common.io.RecursiveDeleteOption...-

loosebazooka: Needs java8. Will add note though.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Extractor.java
```diff
@@ -42,24 +44,61 @@
...
     messageListener.message("Extracting archive: " + archive);
 
-    extractorProvider.extract(archive, destination, messageListener);
+    try {
+      extractorProvider.extract(archive, destination, messageListener);
+    } catch (IOException ex) {
+      try {
+        messageListener.message("Extraction failed, cleaning up " + destination);
+        cleanUp(destination);
+      } catch (Exception exx) {
```
coollog: Should this be just catching an IOException?

loosebazooka: Yeah, probably.

PR #488, opened 2017-11-01
Glue together installer components into SDKInstaller

loosebazooka:
Finally got around to writing this PTAL. It ties together all the components and can install an SDK from start to finish.


src/main/java/com/google/cloud/tools/managedcloudsdk/OsInfo.java
```diff
@@ -0,0 +1,95 @@
...
+/** Enum for Operating System detection. */
+public class OsInfo {
+
+  public enum Name {
+    MAC,
+    WINDOWS,
+    LINUX;
+  }
+
+  public enum Arch {
```
elharo: Architecture?

src/main/java/com/google/cloud/tools/managedcloudsdk/install/FileDeleteVisitor.java
```diff
@@ -0,0 +1,43 @@
...
+package com.google.cloud.tools.managedcloudsdk.install;
+
+import java.io.IOException;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+
+class FileDeleteVisitor extends SimpleFileVisitor<Path> {
```
elharo: consider moving to c.g.cloud.tools.io

src/main/java/com/google/cloud/tools/managedcloudsdk/install/FileResourceProviderFactory.java
```diff
@@ -0,0 +1,98 @@
...
+    switch (os.name()) {
+      case WINDOWS:
+        return "windows-" + getArchitectureString() + ".zip";
+      case MAC:
+        return "darwin-" + getArchitectureString() + ".tar.gz";
+      case LINUX:
+        return "linux-" + getArchitectureString() + ".tar.gz";
+      default:
+        // we can't actually get here
+        throw new RuntimeException(new Throwable());
```
elharo: why chain the throwable?

src/main/java/com/google/cloud/tools/managedcloudsdk/install/SdkInstaller.java
```diff
@@ -0,0 +1,174 @@
...
+    this.installerFactory = installerFactory;
+  }
+
+  /**
+   * Download a Cloud SDK asynchronously.
+   *
+   * @param messageListener a listener on installer output
+   * @return a path to the Cloud SDK home directory (wrapped in a listenable future)
+   */
+  public ListenableFuture<Path> downloadSdk(final MessageListener messageListener) {
```
coollog: Should there be something that makes sure this is only called once, since it assumes a running executorService?

Or maybe throw an error if the submit is rejected?

src/main/java/com/google/cloud/tools/managedcloudsdk/install/SdkInstaller.java
```diff
@@ -0,0 +1,174 @@
...
+    ListenableFuture<Path> resultFuture =
+        executorService.submit(
+            new Callable<Path>() {
+              @Override
+              public Path call() throws Exception {
+                return downloadSdkSync(messageListener);
+              }
+            });
+    executorService.shutdown(); // shutdown executor after install
```
coollog: Wouldn't this shutdown the executor immediately before the download even begins?

loosebazooka: This will actually only shutdown the executor after all submitted jobs are completed.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/SdkInstaller.java
loosebazooka: added in an illegalstateexception

src/main/java/com/google/cloud/tools/managedcloudsdk/install/SdkInstaller.java
coollog: Ah okay that makes sense.

src/test/java/com/google/cloud/tools/managedcloudsdk/install/FileResourceProviderTest.java
```diff
@@ -0,0 +1,49 @@
...
+import java.net.URL;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class FileResourceProviderTest {
+
+  @Test
+  public void smokeTest() throws MalformedURLException {
```
coollog: Should this be in the form of testFileResourceProvider_*?

loosebazooka: Yeah, I guess that makes sense. Lemme update.

PR #493, opened 2017-11-13
Adds service to role mapping metadata in libraries.json

### LINKED ISSUE #485: Add service account role to the client lib metadata
Some APIs can be used with a service account with no role. For example, the translate API. 

Others (those that access project data) require a role to be set: GCS, Datastore etc.

I am proposing to add metadata to the library json that contains the minimum admin role that is needed for a given API. For example:
- GCS - > "Storage Admin"
- Datastore -> "Cloud Datastore Owner"

The consumers of this metadata could then decide to use (or not to use) this info. 

The complication with this proposed approach is with projects that consume multiple APIs. The service account then needs the union of all of these individual roles.

etanshaul:
fixes #485 

Adds mapping from service to set of roles. Note that some services do not have any roles (like the speech API).

It will be up to the client to determine how to consume this. For example, in IntelliJ, the plan is that when the user creates a new service account, the set of roles will be displayed allowing the user to choose which to add.


etanshaul:
unit tests are coming

etanshaul:
Some more info on the approach I took here:

The "master" list of roles was obtained by running a role listing query using the IAM API.

Using this list, I crossed-referenced the set of roles against the roles displayed in the web console:

image

- If a service had a category on the web console (like the "Datastore" category shown above), I mapped only the roles displayed in the web console.

- If a service had no mapping on the web console (like "Compute" for example), I mapped all the roles present in the IAM query.

- Lastly, only "GA" roles were mapped.

etanshaul:
@patflynn @elharo @loosebazooka, ptal

src/main/resources/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -28,6 +28,12 @@
     "name": "Google Cloud Resource Manager",
     "id": "googlecloudresourcemanager",
     "serviceName": "cloudresourcemanager.googleapis.com",
+    "serviceRoles": [
+      "roles/resourcemanager.organizationAdmin",
```
elharo: why always the prefix "roles/"?

etanshaul: because thats how they are identified by GCP (and how we identify them in IAM API calls).

It seems like this path (/) pattern is common to all resources in GCP (not just roles).

etanshaul:
thanks for reviewing @elharo 

@patflynn, @loosebazooka, any comments on this?

PR #494, opened 2017-11-13
Adds --server parameter for deploying configuration YAML.

coollog:
This is for https://github.com/GoogleCloudPlatform/app-maven-plugin/issues/231.

This adds the `DefaultDeployConfiguration#server` parameter to `DefaultDeployProjectConfigurationConfiguration`.

Will need to release a new version after this is pushed to be used in app-maven-plugin.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDeployment.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDeployment.java
@@ -139,6 +139,7 @@ public class CloudSdkAppEngineDeployment implements AppEngineDeployment {
     List<String> arguments = new ArrayList<>();
     arguments.add("deploy");
     arguments.add(deployable.toAbsolutePath().toString());
+    arguments.addAll(GcloudArgs.get("server", configuration.getServer()));
     arguments.addAll(GcloudArgs.get(configuration));
 
     try {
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDeployment.java
```diff
@@ -139,6 +139,7 @@ void deployConfig(String filename, DeployProjectConfigurationConfiguration confi
     List<String> arguments = new ArrayList<>();
     arguments.add("deploy");
     arguments.add(deployable.toAbsolutePath().toString());
+    arguments.addAll(GcloudArgs.get("server", configuration.getServer()));
```
elharo: what if this is not set, i.e. null?

coollog: GcloudArgs#get [calls](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/a028c181b7296148fb5d213c171f6ad475e31b55/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/GcloudArgs.java#L31) [Args#string](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/a028c181b7296148fb5d213c171f6ad475e31b55/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/Args.java#L37), which only adds the parameter if not null.

loosebazooka:
FYI this feature is primarily for users who are able to run their own appengine clusters (for QA for example)

coollog:
Also, as a side note, the fix for the Kokoro checks running will be up tomorrow.

PR #497, opened 2017-11-13
Command executor save output

loosebazooka:
Adds updated processing for command output.
For detecting which components are installed we save the output (to process later).
For showing better output progress, we allow non-full line out.


src/main/java/com/google/cloud/tools/managedcloudsdk/MessageListener.java
```diff
@@ -17,5 +17,18 @@
 package com.google.cloud.tools.managedcloudsdk;
 
 public interface MessageListener {
+  /**
+   * Process a full line. Implementers should add a newline to the end.
+   *
+   * @param line a full line (does not include newline character)
+   */
+  void messageLn(String line);
```
elharo: messageLine (no abbreviations)
maybe even writeLine or some such since verbs are more conventional

src/main/java/com/google/cloud/tools/managedcloudsdk/MessageListener.java
```diff
@@ -17,5 +17,18 @@
...
+
+  /**
+   * Process a raw message. Implementers should not add a newline to the end, it may contain newline
```
elharo: I'm confused as to why we need two methods here, or what this interface is doing.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Downloader.java
```diff
@@ -61,24 +61,24 @@ public void download() throws IOException, InterruptedException {
...
 
       try (BufferedOutputStream out =
           new BufferedOutputStream(
               Files.newOutputStream(destinationFile, StandardOpenOption.CREATE_NEW))) {
         int bytesRead;
         byte[] buffer = new byte[BUFFER_SIZE];
 
-        // Progress is updated every 1%
-        long updateThreshold = contentLength / 100;
+        // Progress is updated every Megabyte
```
elharo: not kilobyte?

src/main/java/com/google/cloud/tools/managedcloudsdk/process/AsyncStreamConsumer.java
```diff
@@ -0,0 +1,66 @@
...
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningExecutorService;
+import com.google.common.util.concurrent.MoreExecutors;
+import com.google.common.util.concurrent.SettableFuture;
+import java.io.InputStream;
+import java.util.concurrent.Executors;
+
+/**
+ * AsyncWrapper to handle stream consumption on a separate thread. Do not re-use this on streams -
+ * it can only handle one stream per instance.
```
elharo: We could probably rejigger this class such that it can't be reused.

src/main/java/com/google/cloud/tools/managedcloudsdk/process/ByteHandler.java
```diff
@@ -0,0 +1,28 @@
...
+
+package com.google.cloud.tools.managedcloudsdk.process;
+
+/**
+ * Handle bytes from a stream consumer.
+ *
+ * @param <T> use {@code Void} if you don't want to store the result
+ */
+public interface ByteHandler<T> {
+  void bytes(byte[] bytes, int length);
```
elharo: If length is > bytes.length what happens?

src/main/java/com/google/cloud/tools/managedcloudsdk/process/CollectingLineHandler.java
```diff
@@ -0,0 +1,36 @@
...
+public class CollectingLineHandler implements LineHandler<String> {
+
+  private final StringBuilder result = new StringBuilder("");
+
+  @Override
+  public void line(String line) {
+    if (line == null) {
+      return;
+    }
+    result.append(line).append(System.lineSeparator());
```
elharo: depending on what this is used for a literal line break character might be better.

src/main/java/com/google/cloud/tools/managedcloudsdk/process/StreamLineConsumer.java
```diff
@@ -0,0 +1,49 @@
...
+  private final LineHandler<T> lineHandler;
+
+  public StreamLineConsumer(InputStream inputStream, LineHandler<T> lineHandler) {
+    this.inputStream = inputStream;
+    this.lineHandler = lineHandler;
+  }
+
+  @Override
+  public T call() throws Exception {
+    try (BufferedReader br = new BufferedReader(new InputStreamReader(inputStream))) {
```
elharo: BufferedReader has some surprising behavior depending on line endings. It's safer to not use it.

src/test/java/com/google/cloud/tools/managedcloudsdk/process/CommandExecutorTest.java
```diff
@@ -76,93 +71,61 @@ public void testRun() throws IOException, InterruptedException, ExecutionExcepti
...
-
-    new CommandExecutor()
-        .setMessageListener(messageListener)
-        .setProcessBuilderFactory(processBuilderFactoryMock)
-        .run(command);
-
-    loggerInOrder.verify(messageListener).message("Running command : someCommand someOption");
-    loggerInOrder.verify(messageListener).message("IO Exception reading process output");
-  }
+  //  @Test
```
elharo: fix or delete

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Downloader.java
loosebazooka: my numbers are wrong, I want megabyte.

src/main/java/com/google/cloud/tools/managedcloudsdk/process/ByteHandler.java
loosebazooka: Presumably the caller will never do that as this is just a passthrough.

src/main/java/com/google/cloud/tools/managedcloudsdk/process/CollectingLineHandler.java
loosebazooka: makes sense.

src/test/java/com/google/cloud/tools/managedcloudsdk/process/CommandExecutorTest.java
loosebazooka: oops, will do.

src/main/java/com/google/cloud/tools/managedcloudsdk/MessageListener.java
loosebazooka: A provider of messages can either send lines or raw messages. This is mostly for having to deal with messages like
```
processing . . . . . . . done.
```
so the output shows progress.

loosebazooka: I think I can change this though, and just have "message" and inject newlines on my own into this when reading line by line.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/InstallerFactory.java
```diff
@@ -54,7 +58,12 @@ public InstallerFactory(OsInfo osInfo, boolean usageReporting) {
         getInstallScriptProvider(),
         usageReporting,
         messageListener,
-        new CommandExecutorFactory());
+        new CommandExecutorFactory(),
+        new AsyncStreamConsumer<>(
```
coollog: I feel like these could probably be simplified, or not even passed in, since the messageListener is already passed into the Installer construction.

src/main/java/com/google/cloud/tools/managedcloudsdk/process/StreamByteConsumer.java
```diff
@@ -0,0 +1,48 @@
...
+
+  /** Instantiated by {@link StreamByteConsumerFactory}. */
+  StreamByteConsumer(InputStream inputStream, ByteHandler<T> byteHandler) {
+    this.inputStream = inputStream;
+    this.byteHandler = byteHandler;
+  }
+
+  @Override
+  public T call() throws Exception {
+    byte[] byteBuffer = new byte[1024];
```
coollog: Maybe put this constant out as a static constant? And have it be the same as the BUFFER_SIZE used in Downloader?

src/main/java/com/google/cloud/tools/managedcloudsdk/process/StreamLineConsumer.java
loosebazooka: I understand when reading from somewhere remote the line endings may not match up, but this is data from the commandline on the same os on the same machine? As far as I can tell, this should be good to go. Would you say scanner is the safer choice here?

src/main/java/com/google/cloud/tools/managedcloudsdk/MessageListenerForwardingHandler.java
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  @Override
+  public void line(String line) {
+    messageListener.messageLn(line);
+  }
+
+  @Override
+  public void bytes(byte[] bytes, int length) {
+    messageListener.message(new String(bytes, 0, length, Charsets.UTF_8));
```
coollog: What's the reason for using UTF_8 encoding here?

src/main/java/com/google/cloud/tools/managedcloudsdk/install/InstallerFactory.java
loosebazooka: The message listener isn't necessary here. For example, you could collect and parse the output for use somewhere else.

loosebazooka: Oh never mind, I see what you're saying.

src/test/java/com/google/cloud/tools/managedcloudsdk/process/CollectingLineHandlerTest.java
```diff
@@ -19,23 +19,22 @@
 import org.junit.Assert;
 import org.junit.Test;
 
-/** Tests for {@link CollectingLineHandler}. */
+/** Tests for {@link CollectingByteHandler}. */
 public class CollectingLineHandlerTest {
```
coollog: CollectingByteHandlerTest?

src/test/java/com/google/cloud/tools/managedcloudsdk/process/CollectingLineHandlerTest.java
```diff
@@ -19,23 +19,22 @@
...
     Assert.assertEquals("line1\nline2\n", testHandler.getResult());
   }
 
   @Test
-  public void testLineHandler_nullIgnored() {
-    CollectingLineHandler testHandler = new CollectingLineHandler();
-    testHandler.line("line1");
-    testHandler.line(null);
-    testHandler.line("line2");
+  public void testLineHandler_junkIgnored() {
```
coollog: testByteHandler_*?

src/main/java/com/google/cloud/tools/managedcloudsdk/process/CollectingByteHandler.java
```diff
@@ -17,16 +17,13 @@
 package com.google.cloud.tools.managedcloudsdk.process;
 
 /** A line handler that collects lines and returns a single result string. */
```
coollog: byte*

src/main/java/com/google/cloud/tools/managedcloudsdk/process/CollectingByteHandler.java
```diff
@@ -17,16 +17,13 @@
...
-public class CollectingLineHandler implements LineHandler<String> {
+public class CollectingByteHandler implements ByteHandler<String> {
```
coollog: What would CollectingByteHandler be used for? It seems the same can be accomplished with just a MessageListenerForwardingHandler.

loosebazooka: It's for getting results from a command like when testing if gcloud has a component or something. It's just a convenience mechanism at this point, but it's value will be evident in a followup CL.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Downloader.java
```diff
@@ -87,15 +86,14 @@ public void download() throws IOException, InterruptedException {
           // update progress
           totalBytesRead += bytesRead;
           long bytesSinceLastUpdate = totalBytesRead - lastUpdated;
-          if (totalBytesRead == contentLength || bytesSinceLastUpdate > updateThreshold) {
-            messageListener.message(
-                String.valueOf(totalBytesRead) + "/" + String.valueOf(contentLength));
+          if (totalBytesRead == contentLength || bytesSinceLastUpdate > UPDATE_THRESHOLD) {
+            messageListener.message(".");
```
coollog: Wouldn't this write a ton of dots if the size is very large?

src/main/java/com/google/cloud/tools/managedcloudsdk/install/TarGzExtractorProvider.java
```diff
@@ -49,7 +49,7 @@ public void extract(Path archive, Path destination, MessageListener messageListe
       while ((entry = in.getNextTarEntry()) != null) {
         final Path entryTarget = destination.resolve(entry.getName());
         if (messageListener != null) {
-          messageListener.message(entryTarget.toString());
+          messageListener.message(entryTarget.toString() + "\n");
```
coollog: .toString() should be unnecessary now.

Ditto below

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Downloader.java
loosebazooka: Yeah, I think that's fine, we're just trying to show progress.

src/test/java/com/google/cloud/tools/managedcloudsdk/install/DownloaderTest.java
```diff
@@ -103,22 +103,15 @@ public void testDownload() throws IOException, InterruptedException {
...
-      long total = Long.valueOf(progressValues[1]);
-      Assert.assertTrue(read - lastRead > 0);
-      lastRead = read;
-      Assert.assertEquals(testFileSize, total);
+    Assert.assertEquals("Downloading " + fakeRemoteResource.toString() + "\n", values.get(0));
+    Assert.assertEquals("Downloading " + String.valueOf(testFileSize) + " bytes\n", values.get(1));
+    Assert.assertEquals("done.\n", values.get(values.size() - 1));
+
+    List<String> dots = values.subList(2, values.size() - 1);
+
```
coollog: Maybe also assert how many dots are expected to show up?

src/test/java/com/google/cloud/tools/managedcloudsdk/install/GenericArchivesVerifier.java
```diff
@@ -73,12 +73,12 @@ public static void assertListenerReceivedExtractionMessages(
 
     // tars allow for duplicate entries
     Mockito.verify(messageListener, Mockito.atLeastOnce())
-        .message(testRoot.resolve(ROOT).toString());
+        .message(testRoot.resolve(ROOT).toString() + "\n");
```
coollog: .toString() here as well

PR #500, opened 2017-11-15
Component Installer and Sdk Updater

loosebazooka:
Adds component installer code and updater code.


src/main/java/com/google/cloud/tools/managedcloudsdk/components/ComponentInstaller.java
```diff
@@ -0,0 +1,82 @@
...
+    // now configure parameters (not OS specific)
+    command.add(gcloud.toString()); // use full path
+    command.add("components");
+    command.add("install");
+    command.add(component.toString());
+    command.add("--quiet");
+
+    CommandExecutor commandExecutor = commandExecutorFactory.newCommandExecutor();
+
+    messageListener.message("Running command : " + Joiner.on(" ").join(command) + "\n");
```
elharo: What is this doing? Just logging to the console?

src/main/java/com/google/cloud/tools/managedcloudsdk/components/ComponentInstaller.java
```diff
@@ -0,0 +1,82 @@
...
+    int exitCode = commandExecutor.run(command, stdOutListener, stdErrListener);
+    if (exitCode != 0) {
+      throw new ExecutionException(
+          "Component Installer exited with non-zero exit code: " + exitCode, new Throwable());
```
elharo: Why new Throwable? It adds no extra information and will be flagged by some static analysis tools

src/main/java/com/google/cloud/tools/managedcloudsdk/components/ComponentInstaller.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+    try {
+      stdErrListener.getResult().get();
+      stdOutListener.getResult().get();
+    } catch (InterruptedException e) {
+      messageListener.message("Output interrupted...\n");
```
elharo: The \n's look wrong. It feels like this should be a stream of messages. Whether \n is needed depends on where they're shown to the user.

src/main/java/com/google/cloud/tools/managedcloudsdk/components/SdkComponentInstaller.java
```diff
@@ -0,0 +1,66 @@
+package com.google.cloud.tools.managedcloudsdk.components;
```
elharo: copyright

src/main/java/com/google/cloud/tools/managedcloudsdk/components/ComponentInstaller.java
loosebazooka: Sure, or wherever it is configured to output.

src/main/java/com/google/cloud/tools/managedcloudsdk/components/ComponentInstaller.java
loosebazooka: ExecutionException requires a throwable, it's possible I'm misusing it though.

src/main/java/com/google/cloud/tools/managedcloudsdk/components/ComponentInstaller.java
loosebazooka: They are expected to be shown to the user.

loosebazooka: It's also mixed mode process output (from gcloud) and messaging from the code.

src/main/java/com/google/cloud/tools/managedcloudsdk/update/SdkUpdater.java
```diff
@@ -0,0 +1,78 @@
...
+import com.google.cloud.tools.managedcloudsdk.MessageListener;
+import com.google.cloud.tools.managedcloudsdk.executors.SdkExecutorServiceFactory;
+import com.google.cloud.tools.managedcloudsdk.executors.SingleThreadExecutorServiceFactory;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningExecutorService;
+import java.nio.file.Path;
+import java.util.concurrent.Callable;
+
+/** Update an SDK. */
+public class SdkUpdater {
```
coollog: Since this class and the SdkComponentInstaller share a lot of common code it probably makes send to put all the common code in an abstract parent.

src/main/java/com/google/cloud/tools/managedcloudsdk/update/SdkUpdater.java
```diff
@@ -0,0 +1,78 @@
...
+
+  private final Path gcloud;
+  private final UpdaterFactory updaterFactory;
+  private final SdkExecutorServiceFactory executorServiceFactory;
+
+  /** Use {@link #newUpdater} to instantiate. */
+  SdkUpdater(
```
coollog: Maybe make this private if the instantiation is to be done with #newUpdater?

src/main/java/com/google/cloud/tools/managedcloudsdk/update/Updater.java
```diff
@@ -0,0 +1,78 @@
...
+final class Updater {
+
+  private final Path gcloud;
+  private final MessageListener messageListener;
+  private final CommandExecutorFactory commandExecutorFactory;
+  private final AsyncStreamHandler<Void> stdOutListener;
+  private final AsyncStreamHandler<Void> stdErrListener;
+
+  /** Instantiated by {@link UpdaterFactory}. */
+  Updater(
```
coollog: Might make sense to have this share an abstract parent with ComponentInstaller, maybe like GcloudWrapper?

src/main/java/com/google/cloud/tools/managedcloudsdk/components/ComponentInstaller.java
elharo: If we really want them to be shown to the end user, the console is the wrong place. For instance there could be a progress bar dialog that updates the message as new messages come in.

src/main/java/com/google/cloud/tools/managedcloudsdk/components/ComponentInstaller.java
elharo: Consulting the Javadoc it looks like you're correct. ExecutionException is only for asynchronous processes. We probably need a different exception type here, maybe  a custom one.

PR #502, opened 2017-11-17
Adds more abstraction into the process layer for convenience.

loosebazooka:


src/test/java/com/google/cloud/tools/managedcloudsdk/process/AsyncStreamHandlerTest.java
```diff
@@ -18,21 +18,25 @@
...
 import java.util.concurrent.Callable;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+import org.mockito.ArgumentCaptor;
 import org.mockito.Mock;
 import org.mockito.Mockito;
 import org.mockito.MockitoAnnotations;
 
 public class AsyncStreamHandlerTest {
```
coollog: This seems to be testing AsyncByteConsumer

src/main/java/com/google/cloud/tools/managedcloudsdk/command/AsyncCommandWrapper.java
```diff
@@ -14,52 +14,59 @@
...
-    this.commandFactory = commandFactory;
+  AsyncCommandWrapper(SdkExecutorServiceFactory executorServiceFactory) {
     this.executorServiceFactory = executorServiceFactory;
   }
 
-  /** Run a command and return a resultless control future. */
-  public ListenableFuture<Void> runCommand(
-      final List<String> parameters, final MessageListener messageListener) {
+  /** Wrap a command runner and return a resultless future. */
+  public ListenableFuture<Void> run(final CommandRunner commandRunner) {
```
coollog: This can probably be refactored into one function that just gets called with the different Callables.

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandFactory.java
```diff
@@ -0,0 +1,72 @@
...
+   * @param environment map of environment variables, can be {@code null}
+   * @param messageListener listener on command output
+   * @return a {@link CommandRunner} configured to run the command
+   */
+  public CommandRunner newRunner(
+      List<String> command,
+      Path workingDirectory,
+      Map<String, String> environment,
+      MessageListener messageListener) {
+    AsyncStreamHandler stdOut =
```
coollog: These duplicate messageListeners seem to have crept back in.

src/main/java/com/google/cloud/tools/managedcloudsdk/gcloud/Potato.java
```diff
@@ -18,19 +18,19 @@
...
-import com.google.cloud.tools.managedcloudsdk.process.CommandExecutorFactory;
-import com.google.cloud.tools.managedcloudsdk.process.CommandRunner;
+import com.google.cloud.tools.managedcloudsdk.process.ProcessExecutorFactory;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 
 /** {@link CommandRunner} factory for running gcloud commands. */
-public class GcloudCommandFactory {
+public class Potato {
```
coollog: Why is this called Potato?

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandCaller.java
```diff
@@ -0,0 +1,84 @@
...
+      AsyncStreamSaver stdErrListener) {
+    this.command = ImmutableList.copyOf(command);
+    this.workingDirectory = workingDirectory;
+    this.environment = environment;
+    this.processExecutorFactory = processExecutorFactory;
+    this.stdOutListener = stdOutListener;
+    this.stdErrListener = stdErrListener;
+  }
+
+  /** Runs the command and returns stdout as a result. */
```
elharo: Expand this comment with @return. I.e. explain that it's collecting stdout onto a string and returning that, not returning a PrintStream.

src/main/java/com/google/cloud/tools/managedcloudsdk/process/AsyncByteConsumer.java
```diff
@@ -0,0 +1,86 @@
...
+
+  /** Create a new instance. */
+  public AsyncByteConsumer(ByteHandler byteHandler) {
+    this(
+        byteHandler,
+        MoreExecutors.listeningDecorator(Executors.newSingleThreadExecutor()),
+        SettableFuture.<String>create());
+  }
+
+  AsyncByteConsumer(
```
elharo: Is this for testing?

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandCaller.java
```diff
@@ -0,0 +1,84 @@
...
+        logger.severe(stdErrListener.getResult().get());
+      } catch (InterruptedException ignored) {
+        // ignored
+      }
+      throw new CommandExitException("Process exited with non-zero exit code: " + exitCode);
+    }
+    try {
+      return stdOutListener.getResult().get();
+    } catch (InterruptedException ex) {
+      throw new ExecutionException("Interrupted obtaining result.", ex);
```
elharo: I wonder if it might make more sense to bubble the InterruptedException from here rather than wrapping it. Does wrapping in an ExecutionException help us out in some way?

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandCaller.java
```diff
@@ -0,0 +1,84 @@
...
+
+  /** Runs the command and returns stdout as a result. */
+  public String call() throws IOException, ExecutionException, CommandExitException {
+    ProcessExecutor processExecutor = processExecutorFactory.newCommandExecutor();
+
+    int exitCode =
+        processExecutor.run(command, workingDirectory, environment, stdOutListener, stdErrListener);
+    if (exitCode != 0) {
+      try {
+        // only log stdErr if we encounter an error.
```
elharo: You could log at info level if exitCode == 0

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandFactory.java
loosebazooka: We need a message listener for each stream.

src/main/java/com/google/cloud/tools/managedcloudsdk/MessageListenerForwardingHandler.java
```diff
@@ -19,7 +19,7 @@
 import com.google.cloud.tools.managedcloudsdk.process.ByteHandler;
 
 /** {@link ByteHandler} that redirects to {@link MessageListener}. */
-public class MessageListenerForwardingHandler implements ByteHandler<Void> {
+public class MessageListenerForwardingHandler implements ByteHandler {
```
elharo: Do we ever instantiate this directly or should it be an abstract class?

src/main/java/com/google/cloud/tools/managedcloudsdk/MessageListenerForwardingHandler.java
```diff
@@ -33,7 +33,7 @@ public void bytes(byte[] bytes, int length) {
   }
 
   @Override
-  public Void getResult() {
+  public String getResult() {
     return null;
```
elharo: This looks like it should be an abstract method.

src/main/java/com/google/cloud/tools/managedcloudsdk/command/AsyncCommandWrapper.java
```diff
@@ -0,0 +1,59 @@
...
+package com.google.cloud.tools.managedcloudsdk.command;
+
+import com.google.cloud.tools.managedcloudsdk.executors.SdkExecutorServiceFactory;
+import com.google.cloud.tools.managedcloudsdk.executors.SingleThreadExecutorServiceFactory;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningExecutorService;
+import java.util.concurrent.Callable;
+
+/** Wrapper around {@link CommandExecutor} to asynchronous execution. */
+public class AsyncCommandWrapper {
```
elharo: We've had some problems in the past with excess asynchronicity around commands.  I wonder if it should be left to the IDE to decide whether and how to execute a command asynchronously. E.g. in Eclipse we might want to use an Eclipse Job or bring up a progress monitor.

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandCaller.java
```diff
@@ -0,0 +1,87 @@
...
+import com.google.cloud.tools.managedcloudsdk.process.ProcessExecutorFactory;
+import com.google.common.collect.ImmutableList;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.logging.Logger;
+
+/** Execute a command and save and return stdout. */
```
elharo: synchronously, right?

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandCaller.java
```diff
@@ -0,0 +1,87 @@
...
+    ProcessExecutor processExecutor = processExecutorFactory.newCommandExecutor();
+
+    try {
+      int exitCode =
+          processExecutor.run(
+              command, workingDirectory, environment, stdOutListener, stdErrListener);
+      if (exitCode != 0) {
+        try {
+          // only log stdErr if we encounter an error
+          logger.severe(stdErrListener.getResult().get());
```
elharo: This info could go into the CommandExitException. As a client, I prefer to manage logging myself and not have libraries do it behind my back.

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandCaller.java
```diff
@@ -0,0 +1,87 @@
...
+        } catch (InterruptedException ignored) {
+          // ignored
+        }
+        throw new CommandExitException("Process exited with non-zero exit code: " + exitCode);
+      }
+      return stdOutListener.getResult().get();
```
elharo: It would also be useful to include this in the CommandExitException in the case of non-zero exit code.

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandRunner.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  /** Run the command and wait for completion. */
+  @Override
+  public Void execute()
+      throws CommandExecutionException, CommandExitException, InterruptedException {
+    ProcessExecutor processExecutor = processExecutorFactory.newCommandExecutor();
+
+    try {
+      int exitCode;
```
elharo: one line

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandRunner.java
```diff
@@ -0,0 +1,78 @@
...
+      exitCode =
+          processExecutor.run(
+              command, workingDirectory, environment, stdOutListener, stdErrListener);
+      if (exitCode != 0) {
+        throw new CommandExitException("Process exited with non-zero exit code: " + exitCode);
```
elharo: again, it would be useful to add stdout and stderr contents to the exceptions here

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandCaller.java
```diff
@@ -0,0 +1,87 @@
...
+   * @param command a command to run
+   * @param workingDirectory the working directory to run in, can be {@code null}
+   * @param environment map of environment variables, can be {@code null}
+   */
+  CommandCaller(
+      List<String> command,
+      Path workingDirectory,
+      Map<String, String> environment,
+      ProcessExecutorFactory processExecutorFactory,
+      AsyncStreamSaver stdOutListener,
```
elharo: It's not obvious to me that these two need to be injected. If they're always working on stdout and stderr, we could just create them as needed inside the class.

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandFactory.java
```diff
@@ -0,0 +1,72 @@
...
+        new AsyncByteConsumer(new MessageListenerForwardingHandler(messageListener));
+    AsyncStreamHandler stdErr =
+        new AsyncByteConsumer(new MessageListenerForwardingHandler(messageListener));
+
+    return new CommandRunner(
+        command, workingDirectory, environment, new ProcessExecutorFactory(), stdOut, stdErr);
```
elharo: If every object gets a new ProcessExecutorFactory, then there's no need to inject. Just create an instance field in the class.

src/main/java/com/google/cloud/tools/managedcloudsdk/command/AsyncCommandWrapper.java
loosebazooka: Yeah, I think I'm going to take out the async stuff and let it all run synchronously.

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandCaller.java
loosebazooka: Refactored to include outputs in exit exception.

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandRunner.java
loosebazooka: Runner doesn't collect output, it's redirecting to the streamHandler so it has no way of actually seeing the output.

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandFactory.java
loosebazooka: Injection is for testability.

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandCaller.java
loosebazooka: done

src/main/java/com/google/cloud/tools/managedcloudsdk/MessageListenerForwardingHandler.java
loosebazooka: It's an artifact of generalizing, but I moved this out and made it package private with the "command" package.

loosebazooka:
@elharo @coollog PTAL

src/main/java/com/google/cloud/tools/managedcloudsdk/command/AsyncStreamHandlerFactory.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private final MessageListener messageListener;
+
+    MessageListenerForwardingHandler(MessageListener messageListener) {
+      this.messageListener = messageListener;
+    }
+
+    @Override
+    public void bytes(byte[] bytes, int length) {
+      messageListener.message(new String(bytes, 0, length));
```
elharo: encoding?

src/main/java/com/google/cloud/tools/managedcloudsdk/command/AsyncStreamSaverFactory.java
```diff
@@ -0,0 +1,45 @@
...
+class AsyncStreamSaverFactory {
+
+  /** Create a new AsyncStreamSaver using the {@link CollectingByteHandler} implementation. */
+  AsyncStreamSaver newSaver() {
+    return new AsyncByteConsumer(new CollectingByteHandler());
+  }
+
+  static class CollectingByteHandler implements ByteHandler {
+
+    private final StringBuilder result = new StringBuilder("");
```
elharo: new StringBuilder();

src/main/java/com/google/cloud/tools/managedcloudsdk/command/AsyncStreamSaverFactory.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Override
+    public void bytes(byte[] bytes, int length) {
+      result.append(new String(bytes, 0, length));
```
elharo: encoding?

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandRunner.java
```diff
@@ -0,0 +1,66 @@
...
+      ProcessExecutorFactory processExecutorFactory,
+      AsyncStreamHandlerFactory streamHandlerFactory) {
+    this.processExecutorFactory = processExecutorFactory;
+    this.streamHandlerFactory = streamHandlerFactory;
+  }
+
+  /** Run the command and wait for completion. */
+  public void run(
+      List<String> command,
+      Path workingDirectory,
```
elharo: should probably document Nullability since you seem to be relying on it

src/main/java/com/google/cloud/tools/managedcloudsdk/process/ProcessExecutor.java
```diff
@@ -82,7 +73,7 @@ public int run(List<String> command, AsyncStreamHandler stdout, AsyncStreamHandl
       exitCode = process.waitFor();
     } catch (InterruptedException ex) {
       process.destroy();
-      throw new ExecutionException("Process cancelled.", ex);
+      throw new InterruptedException("Process cancelled.");
```
elharo: or just rethrow ex

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandRunner.java
```diff
@@ -0,0 +1,66 @@
...
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+
+/** Execute a command and redirect output to handlers. */
+public class CommandRunner {
+  private final ProcessExecutorFactory processExecutorFactory;
+  private final AsyncStreamHandlerFactory streamHandlerFactory;
+
+  CommandRunner(
```
coollog: @VisibleForTesting?

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandCaller.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+/** Execute a command synchronously and save and return stdout. */
+public class CommandCaller {
+  private final ProcessExecutorFactory processExecutorFactory;
+  private final AsyncStreamSaverFactory streamSaverFactory;
+
+  CommandCaller(
```
coollog: @VisibleForTesting?

src/main/java/com/google/cloud/tools/managedcloudsdk/process/AsyncByteConsumer.java
```diff
@@ -0,0 +1,89 @@
...
+            new Callable<String>() {
+              @Override
+              public String call() throws Exception {
+                return consumeBytes(inputStream, byteHandler);
+              }
+            }));
+    executorService.shutdown();
+  }
+
+  String consumeBytes(final InputStream inputStream, ByteHandler byteHandler) throws IOException {
```
coollog: `byteHandler` is probably not necessary to be passed in since it will be the same as the `byteHandler` field.

src/main/java/com/google/cloud/tools/managedcloudsdk/process/AsyncByteConsumer.java
```diff
@@ -0,0 +1,89 @@
...
+    byte[] byteBuffer = new byte[BUFFER_SIZE];
+    int bytesRead;
+    try (InputStream in = inputStream) {
+      while ((bytesRead = in.read(byteBuffer)) != -1) {
+        byteHandler.bytes(byteBuffer, bytesRead);
```
coollog: It might be good to make a note to not use `ByteHandler`s in multiple places (like two `AsyncByteConsumer`s both using the same `ByteHandler`) or to have `ByteHandler#bytes` be marked as needing to be thread-safe.

src/test/java/com/google/cloud/tools/managedcloudsdk/process/AsyncByteConsumerTest.java
```diff
@@ -59,12 +62,33 @@ public void testHandleStream_failIfReused() {
...
+
+    new AsyncByteConsumer(mockByteHandler, mockExecutorService, mockFuture)
+        .consumeBytes(fakeInputStream, mockByteHandler);
+
+    ArgumentCaptor<byte[]> bytes = ArgumentCaptor.forClass(byte[].class);
+    ArgumentCaptor<Integer> nBytes = ArgumentCaptor.forClass(Integer.class);
+    Mockito.verify(mockByteHandler, Mockito.atLeastOnce()).bytes(bytes.capture(), nBytes.capture());
+
+    int count = bytes.getAllValues().size();
+    StringBuilder result = new StringBuilder("");
```
coollog: The empty string here can be removed as well.

src/main/java/com/google/cloud/tools/managedcloudsdk/process/ProcessExecutor.java
```diff
@@ -48,20 +40,19 @@ ProcessBuilder createProcessBuilder() {
...
-  @VisibleForTesting
-  CommandExecutor setProcessBuilderFactory(ProcessBuilderFactory processBuilderFactory) {
-    this.processBuilderFactory = processBuilderFactory;
-    return this;
-  }
-
   /**
    * Runs the command.
    *
    * @param command the list of command line tokens
```
coollog: Probably good to document the other parameters as well.

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandCaller.java
```diff
@@ -0,0 +1,74 @@
...
+          stdOut = "stdout collection interrupted";
+        }
+        try {
+          stdErr = stdErrSaver.getResult().get();
+        } catch (InterruptedException ignored) {
+          stdErr = "stderr collection interrupted";
+        }
+        throw new CommandExitException(exitCode, stdOut + "\n" + stdErr);
+      }
+      return stdOutSaver.getResult().get();
```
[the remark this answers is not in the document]
coollog: Ah right

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandCaller.java
```diff
@@ -0,0 +1,74 @@
...
+import com.google.cloud.tools.managedcloudsdk.process.ProcessExecutor;
+import com.google.cloud.tools.managedcloudsdk.process.ProcessExecutorFactory;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+/** Execute a command synchronously and save and return stdout. */
+public class CommandCaller {
```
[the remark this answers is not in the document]
loosebazooka: it is for the SDK validation stuff. (in the followup cl)

src/main/java/com/google/cloud/tools/managedcloudsdk/process/ProcessExecutor.java
```diff
@@ -73,7 +77,7 @@ public int run(
       exitCode = process.waitFor();
     } catch (InterruptedException ex) {
       process.destroy();
-      throw new InterruptedException("Process cancelled.");
+      throw ex; // rethrow after cleanup
```
coollog: Can this just be like `finally { process.destroy(); }` (without the catch)?

loosebazooka: I think destroying the process before the streams have a chance to fully consume the output could lead to IOExceptions.

loosebazooka: for the case when the process has exited gracefully, and we still call "destroy"

coollog: Okay that makes sense.

PR #503, opened 2017-11-17
Provides an entry point into the Managed SDK mechanism.

loosebazooka:
Depends on #502, look at that first, then this. This is configured to merge into that branch (processHandlingAbstractions), I'll update the pull request after that's ready.

### DIFF
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
@@ -18,6 +18,7 @@ package com.google.cloud.tools.managedcloudsdk;
 
 import static com.google.cloud.tools.managedcloudsdk.OsInfo.Name.WINDOWS;
 
+import com.google.cloud.tools.appengine.cloudsdk.serialization.CloudSdkComponent;
 import com.google.cloud.tools.managedcloudsdk.command.CommandCaller;
 import com.google.cloud.tools.managedcloudsdk.command.CommandExecutionException;
 import com.google.cloud.tools.managedcloudsdk.command.CommandExitException;
@@ -105,11 +106,15 @@ public class ManagedCloudSdk {
             "components",
             "list",
             "--format=json",
-            "--filter=id:" + component + " AND state.name:Not Installed");
+            "--filter=id:" + component);
 
     try {
       String result = CommandCaller.newCaller().call(listComponentCommand, null, null);
-      return !result.contains(component.toString());
+      List<CloudSdkComponent> components = CloudSdkComponent.fromJsonList(result);
+      if (components.size() != 1) {
+        throw new ManagedSdkVerificationException("Invalid component" + component);
+      }
+      return !components.get(0).getState().getName().equals("Not Installed");
     } catch (CommandExecutionException | InterruptedException | CommandExitException ex) {
       throw new ManagedSdkVerificationException(ex);
     }
@@ -135,7 +140,12 @@ public class ManagedCloudSdk {
 
     try {
       String result = CommandCaller.newCaller().call(updateAvailableCommand, null, null);
-      return !result.contains("Update Available");
+      for (CloudSdkComponent component : CloudSdkComponent.fromJsonList(result)) {
+        if (component.getState().getName().equals("Update Available")) {
+          return false;
+        }
+      }
+      return true;
     } catch (CommandExecutionException | InterruptedException | CommandExitException ex) {
       throw new ManagedSdkVerificationException(ex);
     }
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
@@ -1,0 +1,174 @@
+/*
+ * Copyright 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.managedcloudsdk;
+
+import static com.google.cloud.tools.managedcloudsdk.OsInfo.Name.WINDOWS;
+
+import com.google.cloud.tools.managedcloudsdk.command.CommandCaller;
+import com.google.cloud.tools.managedcloudsdk.command.CommandExecutionException;
+import com.google.cloud.tools.managedcloudsdk.command.CommandExitException;
+import com.google.cloud.tools.managedcloudsdk.components.SdkComponent;
+import com.google.cloud.tools.managedcloudsdk.components.SdkComponentInstaller;
+import com.google.cloud.tools.managedcloudsdk.install.SdkInstaller;
+import com.google.cloud.tools.managedcloudsdk.update.SdkUpdater;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.List;
+
+/** A manager for installing, configuring and updating the Cloud SDK. */
+public class ManagedCloudSdk {
+
+  private final Version version;
+  private final Path managedSdkDirectory;
+  private final OsInfo osInfo;
+
+  /** Instantiated with {@link ManagedCloudSdk#newManagedSdk}. */
+  ManagedCloudSdk(Version version, Path managedSdkDirectory, OsInfo osInfo) {
+    this.version = version;
+    this.managedSdkDirectory = managedSdkDirectory;
+    this.osInfo = osInfo;
+  }
+
+  public Path getSdkHome() {
+    return managedSdkDirectory.resolve(version.getVersion()).resolve("google-cloud-sdk");
+  }
+
+  /** Returns a path to gcloud executable (operating system specific). */
+  public Path getGcloud() {
+    return getSdkHome()
+        .resolve("bin")
+        .resolve(osInfo.name().equals(WINDOWS) ? "gcloud.cmd" : "gcloud");
+  }
+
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdkFactory.java
```diff
@@ -0,0 +1,44 @@
...
+import java.nio.file.Paths;
+
+/** {@link ManagedCloudSdk} factory. */
+public class ManagedCloudSdkFactory {
+
+  private Path managedSdkDirectory;
+  private OsInfo osInfo;
+
+  /** Create a new factory with hardcoded managed sdk root directory. */
+  public ManagedCloudSdkFactory() throws UnsupportedOsException {
```
elharo: If the directory is always the same, should this be a singleton?

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
```diff
@@ -0,0 +1,128 @@
...
+      throws IOException, ExecutionException, CommandExitException {
+    if (!isInstalled()) {
+      return false;
+    }
+    List<String> listComponentCommand =
+        Arrays.asList(
+            "components",
+            "list",
+            "--format=json",
+            "--filter=id:" + component.toString() + " AND state.name:Not Installed");
```
coollog: nit: unnecessary toString()

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdkFactory.java
loosebazooka: The ManagedCloudSdkFactory?

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
```diff
@@ -0,0 +1,147 @@
...
+    List<String> listComponentCommand =
+        Arrays.asList(
+            getGcloud().toString(),
+            "components",
+            "list",
+            "--format=json",
+            "--filter=id:" + component + " AND state.name:Not Installed");
+
+    String result = new CommandFactory().newCaller(listComponentCommand, null, null).call();
+    return !result.contains(component.toString());
```
elharo: Using string search instead of parsing the JSON is risky. E.g. imagine at some point in the future there are python and python3 components. A search for python would return true if python3 is installed.

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  /** Query gcloud to see if sdk is up to date. Gcloud makes a call to the server to check this. */
+  public boolean isUpToDate() throws IOException, ExecutionException, CommandExitException {
```
elharo: ExecutionException feels like an implementation detail that shouldn't be exposed at this level. Though perhaps it's a lower level that shouldn't be throwing this?

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
```diff
@@ -0,0 +1,147 @@
...
+            "components",
+            "list",
+            "--format=json",
+            "--filter=state.name:Update Available");
+
+    String result = new CommandFactory().newCaller(updateAvailableCommand, null, null).call();
+    return !result.contains("Update Available");
+  }
+
+  // TODO : fix passthrough for useragent and client side usage reporting
```
elharo: Please file an issue for this so we don't forget about it.

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+    return true;
+  }
+
+  /**
+   * Query gcloud to see if component is installed. Gcloud makes a call to the server to check this,
+   * in future version we can explore the use of '--local-state-only' but that's a relatively new
+   * flag and may not work for all cases.
+   */
+  public boolean hasComponent(SdkComponent component)
```
elharo: ExecutionException feels like an implementation detail that shouldn't be exposed at this level. Though perhaps it's a lower level that shouldn't be throwing this?

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
```diff
@@ -0,0 +1,147 @@
...
+    String userAgentString = "google-cloud-tools-java";
+    return SdkInstaller.newInstaller(managedSdkDirectory, version, osInfo, userAgentString, false);
+  }
+
+  public SdkComponentInstaller newComponentInstaller() {
+    return SdkComponentInstaller.newComponentInstaller(getGcloud());
+  }
+
+  /** Return a new updater if a 'LATEST' sdk, returns {@code null} for versioned SDKs. */
+  public SdkUpdater newUpdater() {
```
elharo: I don't understand why we need this method. Can you elaborate?

src/test/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdkIntegrationTest.java
```diff
@@ -0,0 +1,145 @@
...
+    Assert.assertFalse(testSdk.isUpToDate());
+
+    ListenableFuture<Void> updateFuture = testSdk.newUpdater().update(testListener);
+    updateFuture.get();
+
+    Assert.assertTrue(testSdk.isInstalled());
+    Assert.assertTrue(testSdk.isUpToDate());
+  }
+
+  static class StdOutMessageListener implements MessageListener {
```
elharo: private?

src/test/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdkIntegrationTest.java
```diff
@@ -0,0 +1,145 @@
...
+    @Override
+    public void message(String rawString) {
+      System.out.print(rawString);
```
elharo: General principle that passing tests should generate no output. Consider using a non stdout PrintStream instead.

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
loosebazooka: So the latest SDK wont automatically be up to date all the time. If an update is available (isUpToDate() = false), then we need to start up a new updater to run `gcloud components update` or whatever.

elharo: Makes sense. Can you update the Javadoc to include this?

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
```diff
@@ -0,0 +1,147 @@
...
+    if (version != Version.LATEST) {
+      return null;
```
elharo: Returning null here feels iffy. I'd expect this method to return an object unconditionally and a separate check be made at a higher level to decide whether this method call is needed.

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
```diff
@@ -0,0 +1,184 @@
...
+    if (getSdkHome() == null) {
+      return false;
+    }
+    if (!Files.isDirectory(getSdkHome())) {
+      return false;
+    }
+    if (!Files.isRegularFile(getGcloud())) {
+      return false;
+    }
+    // verify the versions match up for fixed version installs
```
coollog: nit: Capital V

src/test/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdkIntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+    Assert.assertFalse(testSdk.hasComponent(testComponent));
+    Assert.assertTrue(testSdk.isUpToDate());
+
+    testSdk.newComponentInstaller().installComponent(testComponent, testListener);
+
+    Assert.assertTrue(testSdk.isInstalled());
+    Assert.assertTrue(testSdk.hasComponent(testComponent));
+    Assert.assertTrue(testSdk.isUpToDate());
+
+    // make sure we cant update a versioned cloud sdk
```
coollog: nit: Capital M and can't*

src/test/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdkIntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.Arrays;
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.junit.rules.TestWatcher;
+import org.junit.runner.Description;
+
+/** Integration tests for {@link ManagedCloudSdk} and all supporting classes. */
+public class ManagedCloudSdkIntegrationTest {
```
coollog: Is this the standard place to put integration tests? This test seems to take a very long time to complete.

loosebazooka: Yeah, it is a long test, and I don't know if it's an "integration test" exactly, but we do want it to run all time, maybe I'll rename it and leave it in.

loosebazooka:
ptal.

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
```diff
@@ -0,0 +1,189 @@
...
+      return false;
+    }
+    // Verify the versions match up for fixed version installs
+    if (version != Version.LATEST) {
+      try {
+        String versionFileContents =
+            new String(Files.readAllBytes(getSdkHome().resolve("VERSION"))).trim();
+        if (!versionFileContents.equals(version.getVersion())) {
+          throw new ManagedSdkVersionMismatchException(
+              "Installed sdk version: "
```
elharo: SDK

loosebazooka:
@elharo, sounds good, is there a strong preference for package-info.java in the managed SDK package or would README.md be sufficient?

elharo:
Either one works.

PR #504, opened 2017-11-21
fix a few warnings

elharo:
@loosebazooka


loosebazooka:
I'm in the middle of this code at the moment, this should be fixed by #502

PR #508, opened 2017-12-06
fix warnings

elharo:
@loosebazooka


loosebazooka:
There appears to be formatting issues with this.

PR #509, opened 2017-12-12
Update README for SDK Manager

loosebazooka:


README.md
```diff
@@ -57,3 +57,34 @@ configuration.setVersion("v1");
 // deploy
 deployment.deploy(deployConfiguration);
 ```
+
+## SDK Manager/Downloader
```
etanshaul: can you just call it the SDK Manager? downloading is part of what the manager does. I don't really care though.

loosebazooka: yeah

loosebazooka:
Hrmm, tests broken on windows? I'm taking a look.

README.md
```diff
@@ -57,3 +57,34 @@ configuration.setVersion("v1");
 // deploy
 deployment.deploy(deployConfiguration);
 ```
+
+## SDK Manager
+
+When using the Cloud SDK manager to provide a cloud SDK to users, the implementer must manage all operations. The operations are intended to run asynchronously, either on an executor or through mechanisms provided by an IDE.
```
elharo: I don't know what you mean by "must manage all operations."

README.md
```diff
@@ -57,3 +57,34 @@ configuration.setVersion("v1");
...
+
+```java
+// Create a new Managed SDK instance
+ManagedCloudSdk sdk = ManagedCloudSdk.newManagedSdk("123.123.123") // SDK fixed at version.
+ManagedCloudSdk sdk = ManagedCloudSdk.newManagedSdk() // 'LATEST' sdk, can be updated.
+
+// Implement the listener interface to listen to operation output
+MessageListener listener = new MessageListener() {...};
+
+// Always check if operations are needed before running them
```
elharo: If we always have to do this, we should push this into the install command itself. Just make it a noop if the SDK is already installed.

README.md
loosebazooka: Yeah, I guess what I mean is that it's not updated automatically... I'll change this text.

README.md
loosebazooka: Install's behavior is to write over a previous install. Although I don't see a library user using this in more than one place in their code, they could potentially wrap it on their own.

elharo: Again, if users really have to do this on every install it's safer and easier to just make that part of the call. However that doesn't need to block this PR.

PR #516, opened 2018-01-03
Swallow UnknownArchiveTypeException in SdkInstaller.install()

chanseokoh:
We construct `fileResourceProvider.getArchiveDestination()` so that it always returns either .tar.gz or .zip. See `FileResourceProviderFactory.newFileResourceProvider()` and `FileResourceProviderFactory.getVersionedFilename()`.

```java
      return new FileResourceProvider(
          new URL(LATEST_URL),
          downloads.resolve("google-cloud-sdk.tar.gz"),
```

```java
  private String getVersionedOsExtension() {
    switch (os.name()) {
      case WINDOWS:
        return "windows-" + getArchitectureString() + ".zip";
      case MAC:
        return "darwin-" + getArchitectureString() + ".tar.gz";
      case LINUX:
        return "linux-" + getArchitectureString() + ".tar.gz";
      default:
        // we can't actually get here
        throw new RuntimeException();
    }
  }
```

What do you think?

### DIFF
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/install/SdkInstaller.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/install/SdkInstaller.java
@@ -47,8 +47,8 @@ public class SdkInstaller {
 
   /** Download and install a new Cloud SDK. */
   public Path install(final MessageListener messageListener)
-      throws IOException, InterruptedException, SdkInstallerException, UnknownArchiveTypeException,
-          CommandExecutionException, CommandExitException {
+      throws IOException, InterruptedException, SdkInstallerException, CommandExecutionException,
+          CommandExitException {
 
     FileResourceProvider fileResourceProvider =
         fileResourceProviderFactory.newFileResourceProvider();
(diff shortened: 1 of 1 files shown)


loosebazooka:
Yeah makes sense.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/SdkInstaller.java
```diff
@@ -83,17 +83,23 @@ public Path install(final MessageListener messageListener)
...
+          .extract();
+      if (!Files.isDirectory(fileResourceProvider.getExtractedSdkHome())) {
+        throw new SdkInstallerException(
+            "Extraction succeeded but valid sdk home not found at "
+                + fileResourceProvider.getExtractedSdkHome());
+      }
+    } catch (UnknownArchiveTypeException e) {
+      // fileResourceProviderFactory.newFileResourceProvider() creates a fileResourceProvider that
+      // returns either .tar.gz or .zip for getArchiveDestination().
+      throw new RuntimeException();
```
elharo: chain the exception

elharo:
```
  private String getVersionedOsExtension() {
    switch (os.name()) {
      case WINDOWS:
        return "windows-" + getArchitectureString() + ".zip";
      case MAC:
        return "darwin-" + getArchitectureString() + ".tar.gz";
      case LINUX:
        return "linux-" + getArchitectureString() + ".tar.gz";
      default:
        // we totally can get here. Does this make a difference?
        throw new RuntimeException();
    }
  }
```

loosebazooka:
@elharo, it's an enum, and so far we've covered all the cases, it could be possible in the future that we do not (by adding more to the enum), but default is unreachable at the moment.

chanseokoh:
@elharo are we good on this? I've just chained the exception. For the switch you mentioned, only defined in the enum class are those 3 values, so the `default` case is technically unreachable.

elharo:
If os.name() returns an enum then can `getVersionedOsExtension` be moved into that enum? That will handle any future extensions to the list of OSs, simplify the code, and make it more OO.

PR #518, opened 2018-01-24
Higher fidelity feedback.

### LINKED ISSUE #519: Better progress monitoring for managed SDK operations
Latest iteration of the progress listener interface with child/subtask in it. 

Key changes : 
- workDone is now *not* cumulative.
- allow child-progressListeners. 
  - ProgressListener no long need to know any state except their own.
  - `normalizedProgress = currentProgress * allocation/totalWork`

```
public interface ProgressListener {
  void start(String message, long totalWork);
  void update(long workDone); // not cumulative
  void done();
  ProgressListener newChild(long allocation);
}
```

an example implementation would look something like this : 
```java
 public static class ParentListener implements ProgressListener {

    private final long totalWorkDone = 0;

    @Override
    public void start(String message, long totalWork) {
      SystemProgressBar.setMessage(message);
    }

    @Override
    public void update(long workDone) {
      totalWorkDone += workDone;
      SystemProgressBar.setWorkDone(totalWorkDone);
    }

    // this is not part of the progress api, but a convenience of this implementation
    public void update(String message) {
      SystemProgressBar.setMessage(message);
    }

    @Override
    public void done() {
      SystemProgressBar.done()
    }

    @Override
    public ProgressListener newChild(long allocation) {
      return new ChildListener(allocation, this);
    }
  }

  public static class ChildListener implements ProgressListener {

    private final long allocation;
    private final ParentListener parent;

    private long totalWork;
    private long totalWorkDone = 0;
    private long lastUpdate = 0;

    public ChildListener(long allocation, ParentListener parent) {
      this.allocation = allocation;
      this.parent = parent;
    }

    @Override
    public void start(String message, long totalWork) {
      this.totalWork = totalWork;
      parent.update(message);
    }

    @Override
    public void update(long workDone) {
      totalWorkDone += workDone;
      long normalizedtotalWorkDone = totalWorkDone * allocation / totalWork;

      if(normalizedtotalWorkDone > lastUpdate) {
        parent.update(normalizedtotalWorkDone - lastUpdate);
        lastUpdate = normalizedtotalWorkDone;
      }
    }

    @Override
    public void done() {
      parent.update(allocation - normalizedTotalWorkDone);
    }

    @Override
    public ProgressListener newChild(long allocation) {
      return null; // or whatever
    }
  }

}
```

loosebazooka:
Provide higher fidelity listener interface for feedback to managed sdk actions.
fixes #519 

Progress is only provided by the install action.

Component installer and sdk updater only provide console feedback: the programmer can provide a progress bar for it to display to the user, but the library doesn't provide any progress data.


src/main/java/com/google/cloud/tools/managedcloudsdk/textbars/TextProgressBar.java
```diff
@@ -0,0 +1,65 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.managedcloudsdk.textbars;
+
+import com.google.cloud.tools.managedcloudsdk.MessageListener;
+
+public class TextProgressBar {
```
coollog: Perhaps add a comment showing what the output looks like here?

src/main/java/com/google/cloud/tools/managedcloudsdk/textbars/TextProgressBar.java
```diff
@@ -0,0 +1,65 @@
...
+    messageListener.message("#=");
+  }
+
+  /**
+   * Update the progress bar.
+   *
+   * @param value change since last update call
+   */
+  public void update(long value) {
+    totalRead = totalRead + value;
```
coollog: += ?

src/main/java/com/google/cloud/tools/managedcloudsdk/textbars/TextProgressBar.java
```diff
@@ -0,0 +1,65 @@
...
+
+import com.google.cloud.tools.managedcloudsdk.MessageListener;
+
+public class TextProgressBar {
+
+  private static final long BARS = 58;
+
+  private final MessageListener messageListener;
+  private final double updateThreshold;
+  private double totalRead = 0;
```
coollog: Seems like `totalRead` and `progress` can both be `long`?

src/main/java/com/google/cloud/tools/managedcloudsdk/textbars/TextProgressBar.java
```diff
@@ -0,0 +1,65 @@
...
+  private double progress = 0;
```
coollog: Perhaps `barsAdded` or something similar would be a more descriptive name.

src/main/java/com/google/cloud/tools/managedcloudsdk/textbars/TextProgressBar.java
```diff
@@ -0,0 +1,65 @@
...
+
+  TextProgressBar(MessageListener messageListener, long total) {
+    this.messageListener = messageListener;
+    this.updateThreshold = ((double) total) / BARS;
```
coollog: Might be safer to avoid using a double here and just output new bars based on `totalRead/total`.

src/main/java/com/google/cloud/tools/managedcloudsdk/textbars/TextInfoBar.java
```diff
@@ -0,0 +1,44 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.managedcloudsdk.textbars;
+
+import com.google.cloud.tools.managedcloudsdk.MessageListener;
+
+public class TextInfoBar {
```
coollog: A comment here too showing an example output

src/main/java/com/google/cloud/tools/managedcloudsdk/textbars/TextProgressBar.java
loosebazooka: yeah I changed this.

src/main/java/com/google/cloud/tools/managedcloudsdk/textbars/TextProgressBar.java
loosebazooka: Good call, code is much simpler now.

src/main/java/com/google/cloud/tools/managedcloudsdk/textbars/TextProgressBar.java
```diff
@@ -18,18 +18,29 @@
...
+ */
 public class TextProgressBar {
 
   private static final long BARS = 58;
 
   private final MessageListener messageListener;
-  private final double updateThreshold;
-  private double totalRead = 0;
-  private double progress = 0;
+  // private final double updateThreshold;
```
coollog: Looks like a leftover comment.

src/main/java/com/google/cloud/tools/managedcloudsdk/textbars/TextProgressBar.java
```diff
@@ -43,22 +54,19 @@ public void start() {
...
-      int newBars = (int) (diff / updateThreshold);
-      for (int i = 0; i < newBars; i++) {
-        messageListener.message("=");
-      }
-      progress = progress + newBars;
+    currentProgress += value;
+    long targetBars = currentProgress * 58 / total;
+    while (currentBars < targetBars && currentBars < BARS) {
+      messageListener.message("=");
+      currentBars = currentBars + 1;
```
coollog: ++

briandealwis:
Eclipse has a reporting mechanism called `IProgressMonitor` that looks very similar to `TextProgressBar`:
  - `beginTask(String description, int amountOfWork)`: starting operation with a work budget
     - there's a constant `UNKNOWN = -1` for indeterminate work (like the `Extractor` class); the UI uses this as a hint to just pulse the progress bar
  - `worked(int amountWorked)`: called multiple times over the course of the work; the amountWorked accumulates, and total should be < `amountOfWork`
  - `subTask(String description)`: provides a UI hint that some major subpart is under way
  - `done()`: work is finished; any remaining work budget is consumed

`IProgressMonitor` also provides a helper method like `isCanceled()` (for polling) which could be useful though it's not necessary.

So if you could just pull out an interface `ProgressBar` then we could create a wrapper.

But in Eclipse, `IProgressMonitor`s are provided by the platform and not created by our code.  So rather than pass in a factory, the calling code to `Downloader` and `Extractor` should pass in an instance.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Downloader.java
```diff
@@ -48,14 +50,18 @@
 
   /** Download an archive, this will NOT overwrite a previously existing file. */
   public void download() throws IOException, InterruptedException {
+    download(new TextBarFactory());
+  }
+
+  @VisibleForTesting
+  void download(TextBarFactory textBarFactory) throws IOException, InterruptedException {
```
elharo: This API is very strange and feels like it's mixing things that shouldn't be mixed. I would not expect that a download method requires a piece of UI like textBarFactory.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Extractor.java
```diff
@@ -42,7 +43,14 @@
 
   /** Extract an archive. */
   public void extract() throws IOException, InterruptedException {
-    messageListener.message("Extracting archive: " + archive + "\n");
+    extract(new TextBarFactory());
+  }
+
+  @VisibleForTesting
+  void extract(TextBarFactory textBarFactory) throws IOException, InterruptedException {
```
elharo: ditto

src/main/java/com/google/cloud/tools/managedcloudsdk/textbars/TextBarFactory.java
```diff
@@ -0,0 +1,34 @@
...
+
+  public TextDividerBar newDividerBar(MessageListener messageListener) {
+    return new TextDividerBar(messageListener);
+  }
+
+  public TextInfoBar newInfoBar(MessageListener messageListener, String text) {
+    return new TextInfoBar(messageListener, text);
+  }
+
+  public TextProgressBar newProgressBar(MessageListener messageListener, Long contentSize) {
```
elharo: primitive `long` is usually preferred per Effective Java

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Downloader.java
loosebazooka: it's not part of the API, it's just a way to inject this.

loosebazooka:
@briandealwis this is a purely cosmetic change. The API to the managedsdk stuff doesn't actually have an entry point for progress monitoring at the moment except a text based listener.

We probably should provide one and one that is compatible with eclipse and intellij. #519 targets this?

briandealwis:
It is a cosmetic change, but the information it's putting out is useful.

loosebazooka:
PTAL, change the interface to the actions based on #519.

src/main/java/com/google/cloud/tools/managedcloudsdk/command/AsyncStreamHandlerFactory.java
```diff
@@ -16,32 +16,32 @@
...
-    MessageListenerForwardingHandler(MessageListener messageListener) {
-      this.messageListener = messageListener;
+    ConsoleListenerForwardingByteHandler(ConsoleListener consoleListener) {
+      this.consoleListener = consoleListener;
     }
 
     @Override
     public void bytes(byte[] bytes, int length) {
-      messageListener.message(new String(bytes, 0, length, Charsets.UTF_8));
+      consoleListener.console(new String(bytes, 0, length, Charsets.UTF_8));
```
coollog: nit: `0, length, ` can be omitted

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Installer.java
```diff
@@ -57,7 +61,10 @@ public void install()
     command.add("--quiet"); // don't accept user input during install
     command.add("--usage-reporting=" + usageReporting); // usage reporing passthrough
 
-    commandRunner.run(command, installedSdkRoot, null, messageListener);
+    progressListener.update("Installing Cloud SDK");
+    progressListener.update(250);
```
coollog: It looks like `progressListener` is initialized to start work of 300 in `SdkInstaller#install`. It might be good to add a comment here for where this number comes from or to have all the updates be done in the same scope as where the `totalWork` is set. (Using a different ProgressListener than one set to totalWork of 300 may cause unexpected behavior)

src/main/java/com/google/cloud/tools/managedcloudsdk/install/ZipExtractorProvider.java
```diff
@@ -67,14 +57,13 @@ public void extract(
         count++;
       }
 
-      TextProgressBar progressBar = textBarFactory.newProgressBar(messageListener, count);
-      progressBar.start();
+      // TextProgressBar progressBar = textBarFactory.newProgressBar(messageListener, count);
```
coollog: Is this intentional?

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Downloader.java
```diff
@@ -55,45 +57,38 @@ public void download() throws IOException, InterruptedException {
...
-          totalBytesRead += bytesRead;
-          long bytesSinceLastUpdate = totalBytesRead - lastUpdated;
-          if (totalBytesRead == contentLength || bytesSinceLastUpdate > UPDATE_THRESHOLD) {
-            messageListener.message(".");
-            lastUpdated = totalBytesRead;
-          }
+          out.write(buffer, 0, bytesRead);
+          totalRead = totalRead + bytesRead;
+          // update progress, normalize to 100
+          progressListener.update((int) (totalRead * 100 / contentLength));
```
coollog: Ditto here with the 100. (See comment in Installer.java)

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Extractor.java
```diff
@@ -16,42 +16,48 @@
...
     this.extractorProvider = extractorProvider;
-    this.messageListener = messageListener;
+    this.progressListener = progressListener;
   }
 
   /** Extract an archive. */
   public void extract() throws IOException, InterruptedException {
-    messageListener.message("Extracting archive: " + archive + "\n");
+    progressListener.update("Extracting archive: " + archive.getFileName());
+    progressListener.update(150);
```
coollog: Ditto (See comment in Installer.java)

src/main/java/com/google/cloud/tools/managedcloudsdk/install/ZipExtractorProvider.java
```diff
@@ -38,26 +38,32 @@
...
+  public void extract(Path archive, Path destination) throws IOException {
     // Use ZipFile instead of ZipArchiveInputStream so that we can obtain file permissions
     // on unix-like systems via getUnixMode(). ZipArchiveInputStream doesn't have access to
     // all the zip file data and will return "0" for any call to getUnixMode().
     try (ZipFile zipFile = new ZipFile(archive.toFile())) {
+      long count = 0;
+      Enumeration<ZipArchiveEntry> zipEntriesCounter = zipFile.getEntries();
+      while (zipEntriesCounter.hasMoreElements()) {
+        zipEntriesCounter.nextElement();
+        count++;
```
coollog: It seems like `count` is not being used anymore.

src/main/java/com/google/cloud/tools/managedcloudsdk/ProgressListener.java
```diff
@@ -0,0 +1,28 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.managedcloudsdk;
+
+public interface ProgressListener {
+  // progress
```
coollog: Is this a stray comment?

src/main/java/com/google/cloud/tools/managedcloudsdk/command/AsyncStreamHandlerFactory.java
loosebazooka: The content doesn't necessarily occupy the full array, it's a fixed buffer that's reused, so we need length in here.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Installer.java
briandealwis: This demonstrates why Eclipse has sub-monitors: you're scattering information across the implementation.  The Eclipse convention is that `start()`, `update()`, and `done()` must all be done in the same method.

But it's easy to create a `SubProgressListener` and give it some amount of work.  When its `update()` is called, it just calls update() on its parent.  On `done()`, it calls `update()` with whatever work remains.

loosebazooka: That's reasonable, I was trying to keep the implementer side simple. But I can do that.

src/test/java/com/google/cloud/tools/managedcloudsdk/NullProgressListener.java
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.managedcloudsdk;
+
+public class NullProgressListener implements ProgressListener {
```
briandealwis: It might simplify the other implementers to provide an incremental `ProgressListener` that pushes up updates as they come in.  I think something like the following should work (untested), taking in to account the `UNKNOWN` idea earlier.
```java
interface ProgressListener {
  /** Special value for situations where the total work cannot be determined in advance. */
  long UNKNOWN = -1;

  // etc.
}
class SubProgressListener implements ProgressListener {
  private ProgressListener parent;
  private long parentTotalWork;
  private long parentReported = 0; // accumulated
  private long totalWork;
  private long reported = 0; // accumulated

  public SubProgressListener(ProgressListener parent, long totalWork) {
    Preconditions.checkArgument(totalWork > 0);
    this.parent = parent;
    this.parentTotalWork = totalWork;
  }

  public void start(String message, long totalWork) {
    Preconditions.checkArgument(totalWork > 0 || totalWork == UNKNOWN)
    this.totalWork = totalWork;
  }

  public void update(long worked) {
    Preconditions.checkArgument(worked > 0);
    reported += worked;
    if(totalWork > 0) {
      Preconditions.checkState(reported <= totalWork);
      // ensure not affected by rounding (e.g., parentTotalWork == 1)
      long parentWorked = reported * parentTotalWork / totalWork - parentReported;
      if(parentWorked > 0) {
        parentReported += parentWorked;
        parent.update(parentWorked);
      }
    }
  }

  public void done() {
    if(parentReported < parentTotalWork) {
      update(parentTotalWork - parentReported);
    }
  }
}
```

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Extractor.java
```diff
@@ -16,42 +16,50 @@
...
     this.extractorProvider = extractorProvider;
-    this.messageListener = messageListener;
+    this.progressListener = progressListener;
   }
 
   /** Extract an archive. */
   public void extract() throws IOException, InterruptedException {
-    messageListener.message("Extracting archive: " + archive + "\n");
+    progressListener.start("Extracting archive: " + archive.getFileName(), 2);
+    // The extractor is responsible for progress from [100 to 200], if you plan on reusing this,
```
briandealwis: This comment is no longer true, right?

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Extractor.java
```diff
@@ -16,42 +16,50 @@
...
+    // pass the offset in as a parameter.
+    progressListener.update(1);
 
     try {
-      extractorProvider.extract(archive, destination, messageListener);
+      extractorProvider.extract(archive, destination);
```
briandealwis: It would be nice to provide a child ProgressListener here, but I know the zip and tar.gz extractors don't support obtaining a file count.  The Eclipse `IProgressMonitor`'s support an `UNKNOWN` (-1) as _totalWork_, in which case `update()`s are reported upwards using some kind of exponential backoff (IIRC).  It keeps the progress bar alive.  I incorporated this idea into an alternative child ProgressListener below.

src/test/java/com/google/cloud/tools/managedcloudsdk/NullProgressListener.java
loosebazooka: @briandealwis does this handle the `unknown` case with exponential backoff or were you saying that might be an option?

loosebazooka: Never mind I just reread what you said, that eclipse provides the implementation for -1.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Extractor.java
loosebazooka: I mean -1 could be written into the contract, so implements will presumably have to handle/deal with it if we chose to do it?

src/test/java/com/google/cloud/tools/managedcloudsdk/NullProgressListener.java
loosebazooka: I'll add an implementation in a follow up PR.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Extractor.java
briandealwis: Yeah: written into the contract.  TarGz and Zip extractor would call `start("extracting files", ProgressListener.UNKNOWN)` and just call `update(1)` after each file.  On a non-SSD (or in a VM), extracting the Cloud SDK is a visible chunk of work.

loosebazooka: https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/529

loosebazooka:
PTAL, I've got some follow up feature-bugs related to this filed at #529 #528, but I'd like to submit this.

loosebazooka:
@briandealwis @ivanporty @coollog ?

src/main/java/com/google/cloud/tools/managedcloudsdk/ConsoleListener.java
```diff
@@ -16,12 +16,12 @@
...
 package com.google.cloud.tools.managedcloudsdk;
 
-public interface MessageListener {
+public interface ConsoleListener {
   /**
    * Process a raw message. Implementers should not add a newline to the end, it may contain newline
    * characters of its own.
    *
-   * @param rawString a partial or full messages with all necessary newlines
+   * @param rawString a partial or full message with all necessary newlines
```
briandealwis: "_necessary newlines_" is still a bit odd sounding. "_any embedded newlines_"?

src/main/java/com/google/cloud/tools/managedcloudsdk/ProgressListener.java
```diff
@@ -0,0 +1,46 @@
...
+   * Start tracking progress for this task.
+   *
+   * @param message a message to display on the progress bar
+   * @param totalWork a total amount of work this progress listener handles, can special case to -1
+   *     where it is an unknown amount of work.
+   */
+  void start(String message, long totalWork);
+
+  /**
+   * Update the progress with an amount of work done since start or update was last called. It is
```
briandealwis: s/amount/increment/?  Would that be a bit more clear?

src/main/java/com/google/cloud/tools/managedcloudsdk/ProgressListener.java
```diff
@@ -0,0 +1,46 @@
...
+
+  /** Task is complete. */
+  void done();
+
+  /**
+   * Create a new progressListener child, allocation is how the amount of work of this child has
+   * been assigned by the parent. The implementer of child listeners must normalize their values to
+   * allocation.
+   *
+   * <pre>nomalizedTotalWorkDone = totalWorkDone * allocation / totalWork</pre>
```
briandealwis: Note that the implementer is expected to call `update()` with amounts that sum to the allocation.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Downloader.java
```diff
@@ -55,45 +57,37 @@ public void download() throws IOException, InterruptedException {
...
 
-      messageListener.message("Downloading " + address + "\n");
+      logger.info("Downloading " + address + " to " + destinationFile);
 
       try (BufferedOutputStream out =
           new BufferedOutputStream(
               Files.newOutputStream(destinationFile, StandardOpenOption.CREATE_NEW))) {
+
+        progressListener.start(
+            "Downloading " + String.valueOf(contentLength) + " bytes", contentLength);
```
briandealwis: Is it worth a comment that `getContentLengthLong()` returns -1 if unknown length?

_Downloading 180345724 bytes_ is a bit ugly…  `BinaryByteUnit.format(contentLength)`  from  https://github.com/JakeWharton/byteunits?

src/main/java/com/google/cloud/tools/managedcloudsdk/install/ExtractorProvider.java
```diff
@@ -28,8 +27,7 @@
    *
    * @param archive the archive to extract
    * @param destination the destination folder for extracted files
-   * @param messageListener a listener for extraction messages
    * @throws IOException if extractor fails
    */
-  void extract(Path archive, Path destination, MessageListener messageListener) throws IOException;
+  void extract(Path archive, Path destination) throws IOException;
```
briandealwis: Since you're changing this signature, might it be better to add the ProgressListener here now, even if the implementors do nothing with it?

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Downloader.java
loosebazooka: Added note for -1.

I'll add a bug to update formatting.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/ExtractorProvider.java
loosebazooka: This signature can change later without the implementers API changing.

PR #525, opened 2018-01-31
Unify line listeners

### LINKED ISSUE #524: Unify ProcessOutputLineListener and MessageListener
They're the same thing modulo a \n

elharo:
@loosebazooka @chanseokoh fix #524


loosebazooka:
I'm refactoring some of this for managed sdk.

chanseokoh:
> One is full lines (sdk) vs raw strings (mangedSdk). So at the moment they aren't really merge-able(?)

Yeah, I think they can't be merged at the current state. We have different contracts: for `ProcessOutputLineListener`, implementers are expected to print an additional "\n", while for `MessageListener`, print the given string as-is.

elharo:
I disagree. The line ending vs no line ending is purely cosmetic and not at all enforced by this interface. We shouldn't specify unenforceable constraints. Looking at the code there was no clear reason why line endings were required in MessageListener. It was just done one way in MessageListener and the opposite way in ProcessLineListener for no particular reason. It wasn't wrong. However it was needlessly inconsistent.

If the other refactoring is ready, then merge it. Otherwise we shouldn't wait. I let that go the last time, and it caused a lot more repeated work.

loosebazooka:
It's an implementation detail but not purely cosmetic, it's because of how the lower level reader was working. One (the CloudSdk based code) is reading them a line at a time. The other (managed sdk) is reading bytes and sending those over (so single line progress from the cloud sdk can be displayed). The newer one provides slightly better feedback to the user.

Having said that, they're both still redirecting output from a process, and we can be more consistent, but I wouldn't want to go back towards a line by line output for console feedback. I would prefer our older stuff provides raw feedback.

chanseokoh:
> I would prefer our older stuff provides raw feedback.

Just don't forget this: I *believe*, because the gcloud deploy output does some carriage return trick without line feed (the rotating bar `|` --> `/` --> `-` --> `\` --> ...), if you change the old CloudSdk stuff to provide raw feedback, output in IDEs would be messy.

loosebazooka:
The cloud sdk only uses carriage return on terminals if I'm not mistaken. We would see progress as a progression of dots, `........` (we don't actually see a progression since we're using line reader for stuff though, so we just show all the dots after the line ends).

chanseokoh:
> The cloud sdk only uses carriage return on terminals if I'm not mistaken.

Aha.

loosebazooka:
I'm going to close this for now.

PR #532, opened 2018-02-06
Remove deprecated compute api from libraries.json

loosebazooka:
For whatever reason, this compute api is deprecated : details at https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2732

Offline discussion was to just remove this. Also I removed a bunch of trailing white spaces from this file.


patflynn:
seems like the list should be updated to use whatever the new non-deprecated compute client is.

loosebazooka:
It does not appear like there is a non-deprecated one.

patflynn:
@garrettjonesgoogle what is the correct compute API for Java?

garrettjonesgoogle:
There is not a non-deprecated one yet. We are working on it - it will be a month or two.

Issue #528, opened 2018-02-02
Provide reasonable default for child progress listener


loosebazooka:
see #519 and https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/518#discussion_r165492493

PR #533, opened 2018-02-07
Add child progress listener and UNKNOWN const

### LINKED ISSUE #529: provide provision for unknown total work on progress listener
Provide a mechanism for writing an unknown amount of feedback. details in comments.

https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/518#discussion_r165496798

loosebazooka:
fixed #529 #528 

Provides a default implementation of a child progress listener. Implementers of progress listener should be able to just delegate to it for "newChild".

### DIFF
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/ChildProgressListener.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/ChildProgressListener.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2018 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.managedcloudsdk;
+
+/** Default implementation of a child listener. Should satisfy the simple use case. */
+public class ChildProgressListener implements ProgressListener {
+
+  private final ProgressListener parent;
+
+  private long totalWork;
+  private long totalWorkDone = 0;
+
+  private final long totalAllocatedWork;
+  private long totalAllocatedWorkDone = 0;
+
+  /**
+   * Create a child progress listener that performs a subtask for a parent.
+   *
+   * @param parent the parent ProgressListener that created this child
+   * @param totalAllocatedWork the amount of work from the parent this child is expected to report.
+   */
+  public ChildProgressListener(ProgressListener parent, long totalAllocatedWork) {
+    this.parent = parent;
+    this.totalAllocatedWork = totalAllocatedWork;
+  }
+
+  @Override
+  public void start(String message, long totalWork) {
+    parent.update(message);
+    this.totalWork = totalWork;
+  }
+
+  @Override
+  public void update(long workDone) {
+    totalWorkDone += workDone;
+
+    if (totalWork == UNKNOWN) {
+      // there is room here for an implementation of an unknown amount of work
+      // algorithm, but for now we just ignore any update.
+      return;
+    }
+
+    // normalized work done since last update
+    long normalizedWorkDone =
+        totalWorkDone * totalAllocatedWork / totalWork - totalAllocatedWorkDone;
+
+    if (normalizedWorkDone > 0) {
+      parent.update(normalizedWorkDone);
+      totalAllocatedWorkDone += normalizedWorkDone;
+    }
+  }
+
+  @Override
+  public void update(String message) {
+    parent.update(message);
+  }
+
+  @Override
+  public void done() {
+    if (totalAllocatedWorkDone < totalAllocatedWork) {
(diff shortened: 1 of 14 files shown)


src/main/java/com/google/cloud/tools/managedcloudsdk/install/Installer.java
```diff
@@ -61,10 +61,8 @@ public void install()
     command.add("--quiet"); // don't accept user input during install
     command.add("--usage-reporting=" + usageReporting); // usage reporing passthrough
 
-    progressListener.start("Installing Cloud SDK", 2);
-    progressListener.update(1);
+    progressListener.start("Installing Cloud SDK", -1);
```
briandealwis: Should use `UNKNOWN`

src/main/java/com/google/cloud/tools/managedcloudsdk/install/TarGzExtractorProvider.java
```diff
@@ -42,13 +43,20 @@
   TarGzExtractorProvider() {}
 
   @Override
-  public void extract(Path archive, Path destination) throws IOException {
+  public void extract(Path archive, Path destination, ProgressListener progressListener)
+      throws IOException {
+
+    progressListener.start("Extracting archive: " + archive.getFileName(), -1);
```
briandealwis: Should use `UNKNOWN`

src/main/java/com/google/cloud/tools/managedcloudsdk/install/ZipExtractorProvider.java
```diff
@@ -45,7 +46,11 @@
   ZipExtractorProvider() {}
 
   @Override
-  public void extract(Path archive, Path destination) throws IOException {
+  public void extract(Path archive, Path destination, ProgressListener progressListener)
+      throws IOException {
+
+    progressListener.start("Extracting archive: " + archive.getFileName(), -1);
```
briandealwis: Should use `UNKNOWN`

src/test/java/com/google/cloud/tools/managedcloudsdk/ChildProgressListenerTest.java
```diff
@@ -0,0 +1,138 @@
...
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+
+public class ChildProgressListenerTest {
+
+  @Mock ProgressListener mockParent;
+
+  @Before
+  public void initMocks() {
```
briandealwis: You prefer this over annotating the class with `@RunWith(MockitoJUnitRunner.class)`?

loosebazooka: Yeah, I don't know why though. I must have run into something weird in the past?

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Installer.java
loosebazooka: oops. yeah I'll update these.

src/main/java/com/google/cloud/tools/managedcloudsdk/ChildProgressListener.java
```diff
@@ -0,0 +1,83 @@
...
+    totalWorkDone += workDone;
+
+    if (totalWork == UNKNOWN) {
+      // there is room here for an implementation of an unknown amount of work
+      // algorithm, but for now we just ignore any update.
+      return;
+    }
+
+    // normalized work done since last update
+    long normalizedWorkDone =
```
coollog: nit: might be clearer to put ` - totalAllocatedWorkDone` on in the update on L62, so that the logic is

1) new total allocated work done = work done normalized to total allocated work
2) update parent with difference between new and old
3) update old to new

coollog: Or, perhaps the allocation shouldn't be visible to the child, but rather the parent manages that, since the parent is the one that creates the child via `newChild`. But, that would require changing the `ProgressListener` API.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Installer.java
```diff
@@ -61,10 +61,8 @@ public void install()
     command.add("--quiet"); // don't accept user input during install
     command.add("--usage-reporting=" + usageReporting); // usage reporing passthrough
 
-    progressListener.start("Installing Cloud SDK", 2);
-    progressListener.update(1);
+    progressListener.start("Installing Cloud SDK", ProgressListener.UNKNOWN);
```
coollog: It looks like this progressListener isn't being passed into anything or updated? Is this just going to be a 0/100 progress?

src/main/java/com/google/cloud/tools/managedcloudsdk/ChildProgressListener.java
loosebazooka: The logic as it is right now accounts for the differences in increments between `allocation` and child `total work`. So if `workDone < single unit of allocation` then we have to make sure we carry that over. But I'm not actually sure if that's what you're talking about.

I think the parent shouldn't care what the child is doing with its allocation. As long as it does `allocation` amount of work. I think if we move that translation up to the parent, it would have to keep state.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Installer.java
loosebazooka: Yeah. We can explore adding a watcher to the console, and maybe add an update(1) for every read we make from the console, but I don't know about doing that right now.

Issue #4, opened 2016-03-24
Implement standard GAE staging


joaoandremartins:
We're currently depending on the App Engine SDK to stage standard GAE apps. (i.e., AppCfg.main()).

We should implement staging in the library, cutting the App Engine SDK dependency and also the SdkResolver one (which makes us set a system property with the GAE SDK root folder).

patflynn:
@meltsufin here is the java library that supports standard staging:
https://mvnrepository.com/artifact/com.google.appengine/appengine-staging-sdk

elharo:
That artifact lives in google3 and is currently abadonware.

elharo:
see https://g3doc.corp.google.com/apphosting/g3doc/runtimes/java/maven_sdk.md?cl=head

loosebazooka:
obsolete

PR #535, opened 2018-02-12
More natural staging defaults

elharo:
@briandealwis


briandealwis:
I'd be in favour of this change if we renamed `DefaultStageStandardConfiguration` to something like `SensibleStageStandardConfiguration`.  `DefaultStageStandardConfiguration` to me should reflect the AppCfg defaults.

elharo:
appcfg and its defaults are an implementation detail that should not be exposed to an end user. if we ever manage to fix #4 we won't be depending on appcfg. Our API won't change in that case.

PR #536, opened 2018-02-13
Use OS specific conventions for managed cloud sdk home

### LINKED ISSUE #534: Store managed SDK in more appropriate locations
The managed Cloud SDK  currently installed the SDK in ~/.google-cloud-tools-java, which is contra to the guidance from the various OS specs. 

The [XDG basedir spec](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) would seem to guide using the $XDG_CACHE_HOME (e.g., `~/.cache/google-cloud-tools-java`). 
On MacOS  we should use something like `~/Library/Application Support/`and  on Windows `%LOCALAPPDATA%` (I think).

loosebazooka:
fixes #534


src/main/java/com/google/cloud/tools/managedcloudsdk/InvalidOsStateException.java
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.managedcloudsdk;
+
+/** Exception for when the OS is in a state that is incompatible with our assumptions. */
+public class InvalidOsStateException extends Exception {
```
loosebazooka: Maybe this exception should be CannotFindManagedSdkHomeException or something?

src/main/java/com/google/cloud/tools/managedcloudsdk/InvalidOsStateException.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.managedcloudsdk;
+
+/** Exception for when the OS is in a state that is incompatible with our assumptions. */
```
elharo: delete "Exception for when"

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
```diff
@@ -176,15 +181,49 @@ public SdkUpdater newUpdater() {
...
 
   /** Convenience method to obtain a new LATEST {@link ManagedCloudSdk} instance. */
-  public static ManagedCloudSdk newManagedSdk() throws UnsupportedOsException {
+  public static ManagedCloudSdk newManagedSdk()
+      throws UnsupportedOsException, InvalidOsStateException {
     return newManagedSdk(Version.LATEST);
   }
+
+  @VisibleForTesting
+  static Path getOsSpecificManagedSdkHome(
```
elharo: This feels like a no-args methods. The calls to System.getenv etc. can be made here.

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
```diff
@@ -176,15 +181,49 @@ public SdkUpdater newUpdater() {
...
+      OsInfo.Name osName, Properties systemProperties, Map<String, String> environment)
+      throws InvalidOsStateException {
+    Path userHome = Paths.get(systemProperties.getProperty("user.home"));
+    Path cloudSdkPartialPath = Paths.get("google-cloud-tools-java", "managed-cloud-sdk");
+
+    switch (osName) {
+      case WINDOWS:
+        String localAppData = environment.get("LOCALAPPDATA");
+        if (localAppData == null || localAppData.trim().isEmpty()) {
+          throw new InvalidOsStateException("LOCALAPPDATA environment is invalid or missing");
```
elharo: could we instead fall back to the directory we were using before?

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
```diff
@@ -176,15 +181,49 @@ public SdkUpdater newUpdater() {
...
+        if (!Files.exists(applicationSupport)) {
+          throw new InvalidOsStateException(applicationSupport.toString() + " does not exist");
+        }
+        return applicationSupport.resolve(cloudSdkPartialPath);
+
+      case LINUX:
+        return userHome.resolve(".cache").resolve(cloudSdkPartialPath);
+
+      default:
+        // we can't actually get here
```
elharo: add message

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
```diff
@@ -176,15 +181,49 @@ public SdkUpdater newUpdater() {
...
   }
+
+  @VisibleForTesting
+  static Path getOsSpecificManagedSdkHome(
+      OsInfo.Name osName, Properties systemProperties, Map<String, String> environment)
+      throws InvalidOsStateException {
+    Path userHome = Paths.get(systemProperties.getProperty("user.home"));
+    Path cloudSdkPartialPath = Paths.get("google-cloud-tools-java", "managed-cloud-sdk");
+
+    switch (osName) {
```
elharo: not in this cl but OS might be better as an enum

src/test/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdkTest.java
```diff
@@ -127,4 +133,72 @@ public void testManagedCloudSdk_latest()
     Assert.assertTrue(testSdk.hasComponent(testComponent));
     Assert.assertTrue(testSdk.isUpToDate());
   }
+
+  private static final Path cloudSdkPartialPath =
+      Paths.get("google-cloud-tools-java").resolve("managed-cloud-sdk");
+
+  @Test
+  public void testGetOsSpecificManagedSdk() throws InvalidOsStateException, IOException {
```
elharo: This should be three separate test methods, one per OS

src/test/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdkTest.java
```diff
@@ -127,4 +133,72 @@ public void testManagedCloudSdk_latest()
...
+      // pass
+    }
+  }
+
+  @Test
+  public void testGetOsSpecificManagedSdk_integrationTest()
+      throws UnsupportedOsException, InvalidOsStateException {
+    // this test should throw no exceptions, if it does, then our assumptions about the systems
+    // that we target are wrong.
+    ManagedCloudSdk.getOsSpecificManagedSdkHome(
```
elharo: assert not null or not empty

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
```diff
@@ -176,15 +181,49 @@ public SdkUpdater newUpdater() {
...
+          throw new InvalidOsStateException(applicationSupport.toString() + " does not exist");
+        }
+        return applicationSupport.resolve(cloudSdkPartialPath);
+
+      case LINUX:
+        return userHome.resolve(".cache").resolve(cloudSdkPartialPath);
+
+      default:
+        // we can't actually get here
+        throw new RuntimeException();
```
briandealwis: Couldn't this be `UnsupportedOsException`?

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
briandealwis: I agree:
  - I just tested on MacOS and creating a new user populates a host of new directories including `~/Library/Application Support` even before they ever login.  It's extremely unlikely for `~/Library/Application Support` to not exist.
  - `LOCALAPPDATA` is documented in MSDN as being present.  It's extremely unlikely to not exist.

So I think logging an error and falling back to the XDG-style `~/.cache/google-cloud-tools-java` should be great.  And I think we can then get rid of the `InvalidOsStateException`, I think.

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
briandealwis: I think the arguments are being passed to allow simulate testing for different OSs

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
ivanporty: +1, in favor of getting rid of another checked exception.

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
loosebazooka: This section is more programmer error, a consequence of me using a switch statement here?

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
loosebazooka: went to .cache fallback, and removed exception.

src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java
```diff
@@ -177,14 +185,48 @@ public SdkUpdater newUpdater() {
...
+    Path xdgPath = userHome.resolve(".cache").resolve(cloudSdkPartialPath);
+
+    switch (osName) {
+      case WINDOWS:
+        String localAppData = environment.get("LOCALAPPDATA");
+        if (localAppData == null || localAppData.trim().isEmpty()) {
+          logger.warning("LOCALAPPDATA environment is invalid or missing");
+          return xdgPath;
+        }
+        return Paths.get(localAppData).resolve(cloudSdkPartialPath);
```
briandealwis: Is it worth doing an exists-check on the LOCALAPPDATA path, similar to the MacOS case?

loosebazooka: yeah probably.

PR #543, opened 2018-02-23
Make downloader content length more readable

### LINKED ISSUE #542: Format bytes into human readable number for ProgressListener
When using `ProgressListener`, Managed SDK download process calls it with unformatted long bytes numbers, inconvenient to read, i.e.:

sdk-download

Would be great to have them formatted in locale-sensitive manner using standard Kb/Mb notations.

TadCordle:
Fixes #542

### DIFF
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/install/Downloader.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/install/Downloader.java
@@ -26,6 +26,7 @@ import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
+import java.util.Locale;
 import java.util.logging.Logger;
 
 /** Downloader for downloading a single Cloud SDK archive. */
@@ -71,7 +72,9 @@ final class Downloader {
               Files.newOutputStream(destinationFile, StandardOpenOption.CREATE_NEW))) {
 
         progressListener.start(
-            "Downloading " + String.valueOf(contentLength) + " bytes", contentLength);
+            String.format(
+                Locale.getDefault(), "Downloading %.2f MB", contentLength / 1024.0f / 1024.0f),
+            contentLength);
 
         int bytesRead;
         byte[] buffer = new byte[BUFFER_SIZE];
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/install/Downloader.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/install/Downloader.java
@@ -72,9 +72,7 @@ final class Downloader {
               Files.newOutputStream(destinationFile, StandardOpenOption.CREATE_NEW))) {
 
         progressListener.start(
-            String.format(
-                Locale.getDefault(), "Downloading %.2f MB", contentLength / 1024.0f / 1024.0f),
-            contentLength);
+            getDownloadStatus(contentLength, Locale.getDefault()), contentLength);
 
         int bytesRead;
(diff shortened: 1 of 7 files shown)


src/test/java/com/google/cloud/tools/managedcloudsdk/install/DownloaderTest.java
```diff
@@ -99,7 +100,9 @@ public void testDownload() throws IOException, InterruptedException {
     Assert.assertArrayEquals(Files.readAllBytes(destination), Files.readAllBytes(testSourceFile));
 
     ProgressVerifier.verifyProgress(
-        mockProgressListener, "Downloading " + String.valueOf(testFileSize) + " bytes");
+        mockProgressListener,
+        String.format(
```
elharo: This should be a string literal, not a simple replication of the logic. See the recent ToTT episode on keeping logic out of tests: https://big.corp.google.com/~jmcmaster/testing/2014/07/episode-338-dont-put-logic-in-tests.html

TadCordle: True, should be fixed now.

src/test/java/com/google/cloud/tools/managedcloudsdk/install/DownloaderTest.java
```diff
@@ -101,8 +101,7 @@ public void testDownload() throws IOException, InterruptedException {
 
     ProgressVerifier.verifyProgress(
         mockProgressListener,
-        String.format(
-            Locale.getDefault(), "Downloading %.2f MB", testFileSize / 1024.0f / 1024.0f));
+        String.format(Locale.getDefault(), "Downloading %.2f MB", 0.078125954f));
```
elharo: again, this can go all the way to a single string literal. I don't think you need String.format here.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Downloader.java
```diff
@@ -71,7 +72,9 @@ public void download() throws IOException, InterruptedException {
               Files.newOutputStream(destinationFile, StandardOpenOption.CREATE_NEW))) {
 
         progressListener.start(
-            "Downloading " + String.valueOf(contentLength) + " bytes", contentLength);
+            String.format(
+                Locale.getDefault(), "Downloading %.2f MB", contentLength / 1024.0f / 1024.0f),
```
briandealwis: [java.util.Formatter](https://docs.oracle.com/javase/8/docs/api/java/util/Formatter.html)'s javadoc says you need to have the comma flag to get the numeric grouping: `%,.2f`.

src/test/java/com/google/cloud/tools/managedcloudsdk/install/DownloaderTest.java
TadCordle: I mostly did this to keep the locale specification in, should I just remove it from the test and assume US?

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Downloader.java
elharo: And this bug would be caught with a test that used a string literal showing the desired output rather than a test that duplicates the incomplete logic.

src/test/java/com/google/cloud/tools/managedcloudsdk/install/DownloaderTest.java
elharo: You don't have to assume the locale. You can specify it at the test suite level. I.e. in .travis.yaml

PR #549, opened 2018-02-26
Update to guava21, cleanup todos

### LINKED ISSUE #548: Update Guava to 21.0 (not 24.0-jre)
@etanshaul what version of guava is released with intellij? We don't want to set this too incompatible with intellij or we'll end up with some method missing exceptions if intellij loads in an older version of guava.

@elharo eclipse is presumably safe from these kinds of issues because of osgi?

loosebazooka:
fixed #548

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,7 +80,7 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>20.0</version>
+      <version>21.0</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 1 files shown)


loosebazooka:
Apparently the behavior of guava recursive delete is a little strange on windows/macOS, I'll punt fixing that to another time.

etanshaul:
does Eclipse need to confirm that 21 will work prior to merging? IntelliJ is OK.

loosebazooka:
Yeah I'll wait for @elharo to also approve.

PR #552, opened 2018-02-27
Update windows sdk download path to use bundled python

### LINKED ISSUE #551: Managed Cloud SDK installation requires python
Trying to install the Managed Cloud SDK on a fresh Windows box fails as the installation requires Python.  Here's what we see on our console:
```
Installing/upgrading the Cloud SDK... (may take several minutes)
[Installing Google Cloud SDK]
Welcome to the Google Cloud SDK!

To use the Google Cloud SDK, you must have Python installed and on your PATH.
As an alternative, you may also set the CLOUDSDK_PYTHON environment variable
to the location of your Python executable.
Google Cloud SDK installer will now exit.
Press any key to continue . . . 
```
and the process then hangs.

loosebazooka:
Fixes #551.


chanseokoh:
There is a test failure on Kokoro Windows.

loosebazooka:
oops, fixing that.

chanseokoh:
Can you test if this actually works, e.g., by modifying `continuous.bat` like below?
   1. delete `python.exe` or the entire directory found on `%PATH%`
   1. verify that you get "command not found" when typing `python`
   1. verify the SDK is installed

chanseokoh:
Probably also verify `CLOUDSDK_PYTHON` is not set.

chanseokoh:
https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2855#issuecomment-369025541

What do you think of backporting this fix to 0.4.4 and make a 0.4.5 release? Eclipse can avoid having to upgrade Guava.

briandealwis:
I'm just running the tests on my Windows VM that doesn't have python installed (at least, not on the PATH).

briandealwis:
Uh oh:
```
testManagedCloudSdk_fixedVersion(com.google.cloud.tools.managedcloudsdk.ManagedCloudSdkTest)

ERROR: Cannot use bundled Python installation to update Cloud SDK in
non-interactive mode. Please run again in interactive mode.



If you really want to run in non-interactive mode, please run the
following command before re-running this one:



  FOR /F "delims=" %i in ( '""C:\Users\IEUser\AppData\Local\Temp\junit2343958914746932095\178.0.0\google-cloud-sdk\bin\gcloud.cmd"" components copy-bundled-python'
  ) DO (
    SET CLOUDSDK_PYTHON=%i
  )

(Substitute `%%i` for `%i` if in a .bat script.)

Feb 27, 2018 1:15:30 PM com.google.cloud.tools.managedcloudsdk.ManagedCloudSdk getOsSpecificManagedSdkHome
WARNING: C:\Users\IEUser\AppData\Local\Temp\junit2835618866901702073\Library\Application Support does not exist
Feb 27, 2018 1:15:31 PM com.google.cloud.tools.managedcloudsdk.install.Downloader download
```

chanseokoh:
Looking at the message, maybe a workaround is to set `CLOUDSDK_PYTHON` to the bundled python directory first. Hope it is a working python home when unzipped.

chanseokoh:
Sorry, forget about backporting. We can wait and upgrade Guava in Eclipse.

briandealwis:
> Looking at the message, maybe a workaround is to set CLOUDSDK_PYTHON to the bundled python directory first. Hope it is a working python home when unzipped.

Nope. `gcloud components copy-bundled-python` literally copies the included Python into a temp directory, and outputs the temp location:
```
C:\Users\IEUser> gcloud components copy-bundled-python
c:\users\ieuser\appdata\local\temp\tmpjmkt_z\python\python.exec
```
Setting the CLOUDSDK_PYTHON to that copied directory then allows replacing the bundled python found in 1google-cloud-sdk/platform/bundledpython`.

loosebazooka:
Ugh, this is the worst. Let me see if I can make the kokoro test fail.

loosebazooka:
I mean I think we're going to have to do just that, call the copy command, save the param and then run update with that env set.

elharo:
Can we ask the Cloud SDK team to improve the support for this use case?

loosebazooka:
@briandealwis can you test this? I still need to update the kokoro build to ensure we catch actually build failures.

@elharo yeah. I talked to them. This is the solution they have for windows, given the inability to replace files that are currently in use.

loosebazooka:
@briandealwis hold off on that, I think I need to just make a new Windows VM and mess around with this some more.

chanseokoh:
I just tested the latest commit on Windows and noticed that `SdkComponentInstaller.installComponent()` fails:

```
The syntax of the command is incorrect.
Exception in thread "main" com.google.cloud.tools.managedcloudsdk.command.CommandExitException: Process failed with exit code: 255
	at com.google.cloud.tools.managedcloudsdk.command.CommandRunner.run(CommandRunner.java:59)
	at com.google.cloud.tools.managedcloudsdk.components.SdkComponentInstaller.installComponent(SdkComponentInstaller.java:66)
	at Main.main(Main.java:46)
```

briandealwis:
I think it would be worth while to have `CommandRunner` log the commands to be executed.

chanseokoh:
Tested installing an SDK, installing a component, and updating an out-of-date SDK on Windows. They all work.

src/main/java/com/google/cloud/tools/managedcloudsdk/components/BundledPythonCopier.java
```diff
@@ -0,0 +1,30 @@
...
+package com.google.cloud.tools.managedcloudsdk.components;
+
+import com.google.cloud.tools.managedcloudsdk.command.CommandExecutionException;
+import com.google.cloud.tools.managedcloudsdk.command.CommandExitException;
+import java.util.Map;
+
+public interface BundledPythonCopier {
+  /**
+   * Copy a cloud sdk bundled python into a temporary directory and return an environment map with
+   * the location of that bundled python.
```
chanseokoh: `map defining {@code CLOUDSDK_PYTHON} set to the location of that copied python`?

BTW, from the interface point of view, it feels a bit weird that the method "copyPython" returns a map. What do think of returning a path set let callers to do whatever they want to do?

loosebazooka: Yeah that's true, I can move this around.

loosebazooka: I tried moving it around, but I think as a shared piece of code, the result is a little more easily consumable like this.

src/main/java/com/google/cloud/tools/managedcloudsdk/components/SdkUpdater.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Google Inc.
+ * Copyright 2018 Google Inc.
```
briandealwis: Why did you change the year?

src/test/java/com/google/cloud/tools/managedcloudsdk/components/WindowsBundledPythonCopierTestHelper.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Google Inc.
+ * Copyright 2018 Google Inc.
```
briandealwis: Ditto year

src/main/java/com/google/cloud/tools/managedcloudsdk/components/WindowsBundledPythonCopier.java
```diff
@@ -0,0 +1,49 @@
...
+    this.commandCaller = commandCaller;
+  }
+
+  @Override
+  public Map<String, String> copyPython()
+      throws InterruptedException, CommandExitException, CommandExecutionException {
+    List<String> copyPythonCommand =
+        Arrays.asList(gcloud.toString(), "components", "copy-bundled-python");
+    // A trim() required to remove newlines from call result. Using new lines in windows
+    // environment passed through via ProcessBuilder will result in crpytic : "The syntax of
```
briandealwis: nit: cryptic -> cryptic

loosebazooka:
Maybe this PR can be submitted? I'm out till tuesday though and there is some follow up work.
- removed copied python
- add or fix kokoro tests to run ManagedSdkTests on windows in a new shell rather than directly through the script (try cmd /c or something?)

If you want to submit this and create new PRs for those you can. I will also just come back and deal with these on tuesday.

src/test/java/com/google/cloud/tools/managedcloudsdk/components/WindowsBundledPythonCopierTestHelper.java
loosebazooka: I think this one is fine, It's a new file that for some reason git thinks is a version of that other file that was removed.

loosebazooka:
@chanseokoh if we do remove python, we probably want to change the copybundledpython to just return a  path as you said.

briandealwis:
I suspect we need to copy the bundled python on every update too…

chanseokoh:
> I suspect we need to copy the bundled python on every update too…

Looking at the code, I believe we are copying the bundle on every SDK update and component update. We are covered, right?

briandealwis:
Oops, I submitted that by accident before checking!

PR #556, opened 2018-03-02
Delete copied bundled python

### LINKED ISSUE #554: Delete copied python on windows

chanseokoh:
Fixes #554.

Looks like `Files.deleteRecursively()` is still not available in Guava 21. I reused `FileDeleteVisitor`.

To my surprise, there was no good way to remove a last segment in a `Path` or `File`. Using `..` did not work. `Path.subpath()` returns a relative path, and I have no confidence in correctly converting it to an absolute path on Windows. (I've previously encountered a problem on Windows with `"/C:/some/path"`, for which Windows doesn't interprets `C:` as a drive letter.) So I decided to manipulate the string itself.


src/main/java/com/google/cloud/tools/managedcloudsdk/components/WindowsBundledPythonCopier.java
```diff
@@ -44,6 +49,32 @@
     // environment passed through via ProcessBuilder will result in cryptic : "The syntax of
     // the command is incorrect."
     String tempPythonLocation = commandCaller.call(copyPythonCommand, null, null).trim();
+
+    Runtime.getRuntime()
+        .addShutdownHook(
+            new Thread() {
```
briandealwis: Could reduce this to `new Thread(() -> deleteCopiesPython(tempPythonLocation))`

src/main/java/com/google/cloud/tools/managedcloudsdk/components/WindowsBundledPythonCopier.java
```diff
@@ -44,6 +49,32 @@
     // environment passed through via ProcessBuilder will result in cryptic : "The syntax of
     // the command is incorrect."
     String tempPythonLocation = commandCaller.call(copyPythonCommand, null, null).trim();
```
briandealwis: I think it's worth a comment here describing what's returned (i.e., a path to the exe in a temp directory `c:\users\ieuser\appdata\local\temp\tmpjmkt_z\python\python.exec`) and maybe renaming the variable to `tempPythonExeLocation`.

src/main/java/com/google/cloud/tools/managedcloudsdk/components/WindowsBundledPythonCopier.java
```diff
@@ -44,6 +48,32 @@
...
+
+    Runtime.getRuntime()
```
briandealwis: To be safe, it might be better to check that the path is in the user's `%APPLOCALDATA%\temp`.

src/main/java/com/google/cloud/tools/managedcloudsdk/components/WindowsBundledPythonCopier.java
chanseokoh: Done.

src/main/java/com/google/cloud/tools/managedcloudsdk/components/WindowsBundledPythonCopier.java
chanseokoh: +1

briandealwis:
It's `%LOCALAPPDATA%`, but more importantly is check that its prefix is `%TEMP%`.

src/main/java/com/google/cloud/tools/managedcloudsdk/components/WindowsBundledPythonCopier.java
chanseokoh: I am not sure if it will always be `%APPLOCALDATA%`. I remember other env variables like `%TEMP%` or `%TMP%, and we can't be certain how this will work exactly. I think checking this is an overkill. We are the one that made a copy behind the scene to a temp location, and it is our responsibility to remove it.

The only reason I am testing if it ends with "python.exe" is to prevent a possible future gcloud change that returns a python home instead of the binary. We don't want to accidentally delete the the whole Windows temp directory itself.

briandealwis: That's kinda my point: it's an internal command, it's entirely possible someone may say "hey, let's check if there's already a python and just return that instead"

briandealwis:
My fear is that we may not end up using the bundled python and will be returned the installed python, and blow it away.

chanseokoh:
Huh, didn't know it may not copy.

chanseokoh:
Maybe we should just not attempt a cleanup. It is copied to a temp directory, let the OS handle it. I don't actually find #554 is needed.

chanseokoh:
The reason is not clear, but `MoreFiles.deleteRecursively()` doesn't seem to work on kokoro-macos and kokoro-windows. Not sure why.

loosebazooka:
Might make sense to ask the cloud sdk people how they handle it and why
they made those decisions

chanseokoh:
I reverted `MoreFiles.deleteRecursively()`. Looks like it may try to be on the safe side and back off deleting.  For now, I'll use the previous `FileDeleteVisitor`.

PR #557, opened 2018-03-04
Replace FileDeleteVisitor with Guava MoreFiles.deleteRecursively()

elharo:
Replace the hand-crafted directory deletion code with Guava 21.0's `MoreFiles.deleteRecursively()`.

@chanseokoh @loosebazooka I think I found the missing piece that caused this to fail on Windows the last time we tried this.

tldr; deleting directories has been a hole in Java's cross-platform story for 20 years and I trust the Guava team to have put more time and effort into thinking out the corner cases and working around them than we have.


patflynn:
@elharo is there a bug to link this PR to?

elharo:
I don't think so, just a TODO

patflynn:
In that case perhaps just add more info about what this PR fixes to the change description?

chanseokoh:
FTR: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2880#issuecomment-370256930

briandealwis:
Updated description

elharo:
tldr; deleting directories has been a hole in Java's cross-platform story for 20 years and I trust the Guava team to have put more time and effort into thinking out the corner cases and working around them than we have.

PR #566, opened 2018-03-14
Add auth login,activate-service-account

loosebazooka:


loosebazooka:
@etanshaul @chanseokoh the IDEs don't deal with application-default-credentials, right?

chanseokoh:
Users might be using gcloud application-default-credentials for their own reason, but AFAIU, we encourage people to use the env var ATM. At least our CT4E implementation won't deal with the gcloud application-default-credentials.

etanshaul:
same answer for IntelliJ: the user could use ADC on their own, but we do nothing to support or encourage it in the plugin.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAuth.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.regex.Pattern;
+
+/** Cloud SDK based implementation of {@link Auth}. */
+public class CloudSdkAuth implements Auth {
+  private final CloudSdk cloudSdk;
+  private final Pattern emailPattern =
```
coollog: nit: static?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAuth.java
```diff
@@ -0,0 +1,86 @@
...
+  /**
+   * Logs into the cloud sdk with a specific user (does not retrigger auth flow if user is already
+   * configured for the system).
+   *
+   * @param user a user email
+   */
+  @Override
+  public void login(String user) {
+    Preconditions.checkNotNull(user);
+    Preconditions.checkArgument(emailPattern.matcher(user).find(), "Invalid email: " + user);
```
coollog: Do we want users of this API to handle the `IllegalArgumentException`?

src/main/java/com/google/cloud/tools/appengine/api/auth/Auth.java
```diff
@@ -0,0 +1,27 @@
...
+package com.google.cloud.tools.appengine.api.auth;
+
+import java.nio.file.Path;
+
+public interface Auth {
+  void login(String user);
+
+  void login();
+
+  void activateServiceAccount(Path jsonFile);
```
patflynn: I'm wondering whether we need this.  If we have the path to the json file can't we just pass it as a flag to commands like we do in IDEs?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAuth.java
loosebazooka: That's a good question. I guess I should change this.

chanseokoh:
Got curious. What specifically is this gcloud login for in the context of build tools? For Google libraries running in the App Engine local server?

src/main/java/com/google/cloud/tools/appengine/api/auth/Auth.java
loosebazooka: I think we might end up doing that.

loosebazooka:
gcloud auth login would be mostly necessary in the case that the user does not have a cloud sdk installed except through cloud-tools-sdk-downloader. Since they do not have gcloud on the path, the user would have to do something like `~/.cache/google-cloud-tools-java......./bin/gcloud auth login`

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAuth.java
```diff
@@ -0,0 +1,89 @@
...
+   */
+  @Override
+  public void login(String user) {
+    Preconditions.checkNotNull(user);
+    if (!emailPattern.matcher(user).find()) {
+      throw new AppEngineException("Invalid email address: " + user);
+    }
+    try {
+      cloudSdk.runAuthCommand(Arrays.asList("login", user));
+    } catch (ProcessRunnerException e) {
```
coollog: nit: should we be avoiding single letter variable names?

loosebazooka: yes, I'll update.

PR #574, opened 2018-03-28
Exit with failure if the upload fails for some reason

loosebazooka:


kokoro/release_publish.sh
```diff
@@ -46,12 +46,6 @@ GetSessionID() {
 	eval $__resultvar="'$nxsessionid'"
 }
 
-# Usage: UploadJAR <session ID> <file>
```
loosebazooka: Not sure why this method was here, so took it out.

chanseokoh:
Did it not exit in some previous execution? With the following, it should exit anyway.
```
 # Fail on any error.
 set -e
```

loosebazooka:
@chanseokoh that's a good point, let me go recheck the logs.

loosebazooka:
Looks like the error was actually in the response? but curl succeeded, I'm going to recreate this locally if I can

loosebazooka:
@chanseokoh yeah I saw that, it does work with 401's in my testing in our environments, my hope at least is that this improves things a little.

PR #575, opened 2018-03-28
Order APIs

### LINKED ISSUE #526: Sensible order for cloud libraries
Figure out what order makes sense in the json. E.g. alphabetic? GA before beta before alpha?

elharo:
@loosebazooka @etanshaul fix #526 

I cleaned up the names, alphabetized what was left, and updated some versions.

### DIFF
--- a/src/test/java/com/google/cloud/tools/libraries/LibrariesTest.java
+++ b/src/test/java/com/google/cloud/tools/libraries/LibrariesTest.java
@@ -163,7 +163,7 @@ public class LibrariesTest {
       assertReachable(repo);
     }
   }
-  
+
   @Test
   public void testOrder() throws IOException {
     List<String> names = new ArrayList<>();
--- a/src/test/java/com/google/cloud/tools/libraries/LibrariesTest.java
+++ b/src/test/java/com/google/cloud/tools/libraries/LibrariesTest.java
@@ -24,6 +24,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import javax.json.Json;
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/cloud/tools/libraries/LibrariesTest.java
```diff
@@ -160,4 +163,17 @@ public void testVersionExists() throws IOException {
       assertReachable(repo);
     }
   }
+  
+  @Test
+  public void testOrder() throws IOException {
+    List<String> names = new ArrayList<>();
```
loosebazooka: Sorry, I'm on a little bit of a java8 stuff kick, but can we just check sorted vs unsorted?

```java
    List<String> namesLowerCase = apis.stream().map(api -> api.getString("name").toLowerCase(Locale.US)).collect(Collectors.toList());

    Assert.assertTrue(Comparators.isInOrder(namesLowerCase, String::compareTo));
    //OR
    Assert.assertEquals(namesLowerCase.stream().sorted().collect(Collectors.toList()), namesLowerCase) // if you want to see differences?
```

chanseokoh: Hmm... whatever way it is, I think the final comparison of `.compareTo()` is simply based on the Unicode value, so it may not work as intended even if the strings were lower-cased with a proper locale? Should we be really bothered with a locale in this case?

elharo:
JSON parsers maintain the order of arrays in the data, not quite the same as enforcing sort order.

src/test/java/com/google/cloud/tools/libraries/LibrariesTest.java
elharo: Yes, lower casing always needs a locale. Even basic ASCII letters such as i and I change case differently in different locales. In this case, all names are English and ASCII so Locale. US works. 

Let me play with Java 8.

elharo: The Java 8 suggestions work well as long as the test doesn't fail. However when it does fail (and this did fail initially, and will likely fail again when we add libraries) the more detailed error message from the individual non-streamed checks makes the specific error more apparent. 

This test is what made me realize lower casing was needed. Otherwise Datastore sorted after DNS, to my surprise.

PR #577, opened 2018-04-04
remove inherit option for io

loosebazooka:
No one is using this, cleaning it up. This will help with further refactoring.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/DefaultProcessRunner.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/DefaultProcessRunner.java
@@ -16,8 +16,6 @@
 
 package com.google.cloud.tools.appengine.cloudsdk.internal.process;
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/process/DefaultProcessRunner.java
```diff
@@ -99,14 +80,6 @@ public void run(String[] command) throws ProcessRunnerException {
       // Configure process builder.
       final ProcessBuilder processBuilder = new ProcessBuilder();
 
-      // If there are no listeners, we might still want to redirect stdout and stderr to the parent
-      // process, or not.
-      if (stdOutLineListeners.isEmpty() && inheritProcessOutput) {
```
loosebazooka: I just realized that if someone was using a start listener to get directly at the process's output, this would break them.

loosebazooka:
covered in larger rewrite.

PR #600, opened 2018-05-07
Errorprone

elharo:
Upgrade errorprone; not clear why null issues don't seem to be firing.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -158,8 +158,8 @@
             </path>
           </annotationProcessorPaths>
           <compilerArgs>
-            <arg>-Xep:NullAway:ERROR</arg>
-            <arg>-XepOpt:NullAway:AnnotatedPackages=com.uber</arg>
+            <arg>-Xep:NullAway:ERROR</arg>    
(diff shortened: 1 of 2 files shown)


loosebazooka:
@coollog are you familiar with maven error prone config with nullaway?

coollog:
It looks like the null issues seem to be firing in the [Kokoro builds](https://fusion.corp.google.com/runanalysis/test/prod%3Aappengine-plugins-core%2Fmacos_external%2Fpresubmit/prod%3Aappengine-plugins-core%2Fmacos_external%2Fpresubmit/KOKORO/709637fe-3bab-4726-9e43-2d7102496566/1525715057427/prod%3Aappengine-plugins-core%2Fmacos_external%2Fpresubmit%20Build%20%23182/Target%20Transitions?target=appengine-plugins-core/macos_external/presubmit). Is it just not firing locally?

coollog:
I'm running `./mvnw clean compile` locally and it seems to fail as expected.

PR #604, opened 2018-05-08
Code cleanup

chanseokoh:
I'm fine backing off some if you don't like them.

### DIFF
--- a/src/test/java/com/google/cloud/tools/managedcloudsdk/install/DownloaderTest.java
+++ b/src/test/java/com/google/cloud/tools/managedcloudsdk/install/DownloaderTest.java
@@ -28,23 +28,19 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Locale;
 import org.junit.Assert;
-import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
+import org.junit.runner.RunWith;
 import org.mockito.Mock;
 import org.mockito.Mockito;
-import org.mockito.MockitoAnnotations;
+import org.mockito.junit.MockitoJUnitRunner;
 
+@RunWith(MockitoJUnitRunner.class)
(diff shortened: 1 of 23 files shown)


src/test/java/com/google/cloud/tools/managedcloudsdk/install/DownloaderTest.java
```diff
@@ -156,29 +152,26 @@ public void testDownload_failIfExists() throws IOException, InterruptedException
...
-                Thread.currentThread().interrupt();
-                try {
-                  downloader.download();
-                  Assert.fail("InterruptedException expected but not thrown.");
-                } catch (InterruptedException ex) {
-                  Assert.assertEquals("Download was interrupted", ex.getMessage());
-                } catch (IOException e) {
-                  Assert.fail("Test failed due to IOException");
-                }
+            () -> {
```
elharo: don't do this. JavaNCSS can't handle it.

src/test/java/com/google/cloud/tools/managedcloudsdk/install/ExtractorTest.java
```diff
@@ -21,46 +21,37 @@
...
 
     Mockito.doAnswer(
-            new Answer<Void>() {
-              @Override
-              public Void answer(InvocationOnMock invocationOnMock) throws Throwable {
-                Files.createDirectory(extractionDestination.resolve("some-dir"));
-                Files.createFile(extractionDestination.resolve("some-file"));
-                return null;
-              }
+            unused -> {
```
elharo: ditto; don't introduce lambdas. Maybe it works in tests though if JavaNCSS doesn't have to instrument that?

src/test/java/com/google/cloud/tools/managedcloudsdk/install/DownloaderTest.java
chanseokoh: That's very unfortunate.

PR #607, opened 2018-05-10
autoformat

### LINKED ISSUE #601: Make formatting automatic or remove it
I.e. `$ mvn com.coveo:fmt-maven-plugin:format`

Possibly we can bind this to the validate or compile phase. Otherwise it's wasting too much dev time. We should not run this manually. 

I.e. change the goal from check to format.

elharo:
fix #601 @briandealwis @loosebazooka

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -197,7 +197,6 @@
         <groupId>com.coveo</groupId>
         <artifactId>fmt-maven-plugin</artifactId>
         <version>2.4.0</version>
-        <!-- use fmt:format to auto format -->
         <dependencies>
           <dependency>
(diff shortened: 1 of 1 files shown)


loosebazooka:
I'm not sure this is the right way forward with formatting. There's no way to guarantee formatted code is included with a commit? Especially if the user doesn't build from the command line before committing.

elharo:
I'm OK with that. The problems this catches are trivial. They are not comparable with the consistent and repeated delays because some code didn't get autoformatted. This check has cost us multiple SWE-days by this point. It's not worth it.

PR #613, opened 2018-05-14
Prepare for 0.6.0 release

chanseokoh:


chanseokoh:
I think we're good, the release process already created a tag called `v0.6.0`. That said, I'll re-tag it to include #612.

PR #620, opened 2018-05-15
throw NullPointerException

### LINKED ISSUE #603: FilePermissions.verifyDirectoryCreatable is effectively nullable
and probably shouldn't be

elharo:
@loosebazooka fix #603

### DIFF
--- a/src/main/java/com/google/cloud/tools/io/FilePermissions.java
+++ b/src/main/java/com/google/cloud/tools/io/FilePermissions.java
@@ -18,7 +18,6 @@ package com.google.cloud.tools.io;
 
 import com.google.common.annotations.Beta;
 import com.google.common.base.Preconditions;
-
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/cloud/tools/io/FilePermissions.java
```diff
@@ -41,6 +42,7 @@
   public static void verifyDirectoryCreatable(Path path)
       throws AccessDeniedException, NotDirectoryException {
 
+    Preconditions.checkNotNull(path, "Null directory path");
```
chanseokoh: Could add `@throws NPE` to the Javadoc.

elharo: We normally assume this by default and only document it if the argument is Nullable.

PR #628, opened 2018-05-15
Fix DownloaderTest

### LINKED ISSUE #626: Broken DownloaderTest with non-main thread throwing exceptions
Note that the following test will fail, even though the `AssertionException` will be logged to the output.
```java
  @Test
  public void test() throws InterruptedException {
    Thread t = new Thread(() -> {
      Assert.fail();  // throws AssertionError()
    });
    t.start();
    t.join();
  }
```
That said, the `DownloaderTest.testDownload_interruptTriggersCleanup` is broken.

chanseokoh:
Fixes #626. (`Future.get()` is a blocking call, FYI.)

Also changing to always use UTF-8. What `Charset.defaultCharset()` returns is platform-dependent.


chanseokoh:
Sorry, the approval automatically dismissed by the new commit.

PR #633, opened 2018-05-21
Google java format should check not format.

loosebazooka:
By not checking, this would allow code to submitted unformatted and
uncaught by the test infrastucture.


loosebazooka:
@elharo I understand this is a point of contention, but our test infrastructure cannot correct formatting. The developer must do it on their own. A git presubmit hook to run it can be configured if you like.

loosebazooka:
checkFormat fails very fast AND it is run by `./mvnw clean verify` which everyone should be running anyway before a commit+push. It's the same thing as errorprone or checkstyle or anything else failing. 

So I fail to see what the cost is. The failure happens super early in the build, the new error message in 2.0+ actually tell you the exact command to run to fix the problem (reducing that cognitive load). At most it's like 10 seconds.

If you are talking about build checks on github(like kokoro), then the developer is to blame for not properly following procedure and running verify before pushing. If you want it to happen automatically, you can write a git-hook into your local configuration that does something like https://gist.github.com/arnobroekhof/9454645.

loosebazooka:
So a simple presubmit hook that will autoformat looks like this:

`<project-root>/.git/hooks/pre-submit` and `chmod +x` the file
```bash
#!/bin/bash

echo "Running pre-commit hooks..."

PROJECT_ROOT=$(git rev-parse --show-toplevel)

echo "Running google java format..."
MVN_RESULT=$(cd $PROJECT_ROOT; ./mvnw fmt:format)
# optional, see the output of google java format
echo "$MVN_RESULT"

if [$? -ne 0]; then
  echo "Error running google java format"
  exit 1
fi
```

PR #639, opened 2018-05-24
More null checks

### LINKED ISSUE #112: use @javax.annotation.Nullable for nullable parameters and return values

elharo:
fix #112 @loosebazooka

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
@@ -96,23 +96,26 @@ public class CloudSdkAppEngineFlexibleStaging implements AppEngineFlexibleStagin
   static void copyDockerContext(
       StageFlexibleConfiguration config, CopyService copyService, @Nullable String runtime)
       throws IOException, AppEngineException {
-    if (config.getDockerDirectory() != null && config.getDockerDirectory().exists()) {
+    File dockerDirectory = config.getDockerDirectory();
+    if (dockerDirectory != null && dockerDirectory.exists()) {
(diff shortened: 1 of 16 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
```diff
@@ -96,23 +96,26 @@ static String findRuntime(StageFlexibleConfiguration config)
...
+        if (!Files.isRegularFile(dockerDirectory.toPath().resolve("Dockerfile"))) {
           throw new AppEngineException(
-              "Docker directory "
-                  + config.getDockerDirectory().toPath()
-                  + " does not contain Dockerfile.");
+              "Docker directory " + dockerDirectory.toPath() + " does not contain Dockerfile.");
         } else {
-          copyService.copyDirectory(
-              config.getDockerDirectory().toPath(), config.getStagingDirectory().toPath());
+          File stagingDirectory = config.getStagingDirectory();
```
loosebazooka: Can we enforce non-nullability on the interface `StageFlexibleConfiguration` directly? Since getStagingDirectory cannot be null.

loosebazooka:
Maybe we really should be using Optional more?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
elharo: We definitely should take a deeper look at this and other things that are currently marked Nullable. We may be able to do some refactoring to guarantee non-nullability. Here we'd need to shift from using setter methods to a constructor or perhaps a builder.

PR #644, opened 2018-05-25
Use builder to avoid null fields

elharo:
@loosebazooka


chanseokoh:
`[ERROR] ...: You have 1 Checkstyle violation.`

PR #645, opened 2018-05-25
Fix charset warning in test class

chanseokoh:
```
[DefaultCharset] Implicit use of the platform default charset, which can result in differing behaviour between JVM executions or incorrect behavior if the encoding of the data source doesn't match expectations.
    (see http://errorprone.info/bugpattern/DefaultCharset)
  Did you mean 'result.append(new String(bytes.getAllValues().get(i), 0, nBytes.getAllValues().get(i), UTF_8));' or 'result.append(new String(bytes.getAllValues().get(i), 0, nBytes.getAllValues().get(i), Charset.defaultCharset()));'?
```

### DIFF
--- a/src/test/java/com/google/cloud/tools/managedcloudsdk/command/AsyncByteConsumerTest.java
+++ b/src/test/java/com/google/cloud/tools/managedcloudsdk/command/AsyncByteConsumerTest.java
@@ -82,7 +82,9 @@ public class AsyncByteConsumerTest {
     int count = bytes.getAllValues().size();
     StringBuilder result = new StringBuilder();
     for (int i = 0; i < count; i++) {
-      result.append(new String(bytes.getAllValues().get(i), 0, nBytes.getAllValues().get(i)));
+      byte[] rawBytes = bytes.getAllValues().get(i);
+      int length = nBytes.getAllValues().get(i);
+      result.append(new String(rawBytes, 0, length, StandardCharsets.UTF_8));
     }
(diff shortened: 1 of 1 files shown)


src/test/java/com/google/cloud/tools/managedcloudsdk/command/AsyncByteConsumerTest.java
```diff
@@ -82,7 +82,9 @@ public void testConsumeBytes() throws Exception {
     int count = bytes.getAllValues().size();
     StringBuilder result = new StringBuilder();
     for (int i = 0; i < count; i++) {
-      result.append(new String(bytes.getAllValues().get(i), 0, nBytes.getAllValues().get(i)));
+      byte[] rawBytes = bytes.getAllValues().get(i);
+      int length = nBytes.getAllValues().get(i);
+      result.append(new String(rawBytes, 0, length, StandardCharsets.UTF_8));
```
briandealwis: Although this is likely to always work in practice, I don't see any guarantee that the consumer will receive bytes that align on a character boundary.  In which case, this `new String()` could fail.  Both `AsyncStreamSaverFactory.CollectingByteHandler` and `AsyncStreamHandlerFactory.ConsoleListenerForwardingByteHandler` suffer from this in particular.

chanseokoh: At least we're OK in tests.

PR #652, opened 2018-06-06
Add IoT Core and promote Pubsub to GA

### LINKED ISSUE #649: Add IOT API and update to 0.50.0-alpha
In libraries.json https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-clients/google-cloud-iot

elharo:
fix #649 @etanshaul @loosebazooka 

FYI service roles are documented at https://cloud.google.com/iot/docs/how-tos/iam and in Cloud Console on **IAM & Admin > Roles** https://pantheon.corp.google.com/iam-admin/roles

### DIFF
--- a/src/main/resources/com/google/cloud/tools/libraries/libraries.json
+++ b/src/main/resources/com/google/cloud/tools/libraries/libraries.json
@@ -531,6 +531,42 @@
       }
     ]
   },
+{
+    "name": "IoT Core",
+    "id": "iotcore",
+    "serviceName": "cloudiot.googleapis.com",
+    "serviceRoles": [
+      "roles/cloudiot.admin",
+      "roles/owner",
+      "roles/cloudiot.deviceController",
+      "roles/cloudiot.editor",
+      "roles/editor",
+      "roles/cloudiot.provisioner",
+      "roles/cloudiot.viewer",
+      "roles/viewer",
+      "roles/cloudiot.serviceAgent"
+    ],
+    "documentation": "https://cloud.google.com/iot-core",
+    "description": "A fully managed service that allows you to easily and securely connect, manage, and ingest data from millions of globally dispersed devices.",
+    "transports": ["grpc"],
+    "clients": [
+      {
+      "name": "Java Client Libraries for IoT Core",
+      "language": "java",
+      "site": "https://cloud.google.com/iot-core",
+      "apireference": "https://googlecloudplatform.github.io/google-cloud-java/google-cloud-clients/apidocs/index.html?com/google/cloud/iot/v1/package-summary.htmlocs/index.html?com/google/cloud/dialogflow/v2beta1/package-summary.html",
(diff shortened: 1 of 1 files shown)


src/main/resources/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -526,7 +526,43 @@
...
+      }
+    ]
+  },
+{
+    "name": "IoT Core",
+    "id": "iotcore",
+    "serviceName": "cloudiot.googleapis.com",
+    "serviceRoles": [
+      "roles/cloudiot.admin",
+      "roles/owner",
```
etanshaul: this role looks suspiciously out of place for IoT. like a general owner role?

etanshaul: same for `roles/editor`, `roles/viewer`

etanshaul: the docs you linked to say you can use `roles/cloudiot.editor` instead of `roles\editor`. similar for the others.

elharo: Is this supposed to be an exhaustive list? I simply added what was on that page.

etanshaul: right, but I'm just saying we should used the cloudiot scoped roles and not the general ones like `roles/editor` since those probably grant much broader permissions.

etanshaul: from the first link:

"Note that the roles roles/owner, roles/editor, and roles/viewer include permissions for other Google Cloud Platform services as well. "

I'm not sure why they even listed those

elharo: What's the purpose of including these roles in the metadata? (Eclipse doesn't use them.)

etanshaul: IntelliJ has the feature, during client library addition, to create new service accounts and assign specific roles to them.

For example, if the user selects to add the datastore client library, we present the option to create a new service account and display the associated datastore roles (with the option to pick and choose how many or how little to apply)

elharo: So the question becomes, should IntelliJ offer these roles or not? It seems like we might want them on a service account, or are they deprecated somehow? How do they apply to other APIs where we don't currently list these roles? 

If they shouldn't be used, maybe we should file a bug to correct the  https://cloud.google.com/iot/docs/how-tos/iam page.

etanshaul: lgtm now. I think filing a bug on that page is a good idea. It doesn't seem appropriate to have those roles there. I'll do taht.

PR #653, opened 2018-06-06
Set root as working directory when updating Cloud SDK

### LINKED ISSUE #651: Set root directory as working directory when launching gcloud process for updating
To avoid https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/3131.

chanseokoh:
Fixes #651.

This is like executing the `gcloud` binary from the root directory.

### DIFF
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/components/SdkComponentInstaller.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/components/SdkComponentInstaller.java
@@ -65,9 +65,10 @@ public class SdkComponentInstaller {
       environment = pythonCopier.copyPython();
     }
 
+    Path workingDirectory = gcloud.getRoot();
     List<String> command =
         Arrays.asList(gcloud.toString(), "components", "install", component.toString(), "--quiet");
-    commandRunner.run(command, null, environment, consoleListener);
+    commandRunner.run(command, workingDirectory, environment, consoleListener);
     progressListener.done();
   }
 
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/components/SdkComponentInstaller.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/components/SdkComponentInstaller.java
@@ -34,15 +34,16 @@ import javax.annotation.Nullable;
 /** Install an SDK component. */
 public class SdkComponentInstaller {
 
-  private final Path gcloud;
+  private final Path gcloudPath;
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/cloud/tools/managedcloudsdk/components/SdkComponentInstaller.java
```diff
@@ -65,9 +65,10 @@ public void installComponent(
       environment = pythonCopier.copyPython();
     }
 
+    Path workingDirectory = gcloud.getRoot();
```
elharo: Can this be null? The method can return null.

Also, the variable `gclould` was a little hard to read. I didn't initially realize it was a path and not a CloudSdk object. Should we consider renaming this?

chanseokoh: This will not return null as long as the path is absolute. In our case, the path is an absolute path to the gcloud executable, e.g., `/home/user/.cache/google-cloud-tools-java/managed-cloud-sdk/LATEST/google-cloud-sdk/bin/gcloud`. We don't carry around a relative path whose behavior relies on the current working directory. I've added a precondition check:
```
    Preconditions.checkArgument(gcloudPath.isAbsolute());
```
(BTW, passing `null` does not break things, which is the current behavior.)

I've renamed `gcloud` to `gcloudPath`.

src/main/java/com/google/cloud/tools/managedcloudsdk/components/SdkComponentInstaller.java
```diff
@@ -65,27 +66,29 @@ public void installComponent(
       environment = pythonCopier.copyPython();
     }
 
+    Path workingDirectory = gcloudPath.getRoot();
     List<String> command =
-        Arrays.asList(gcloud.toString(), "components", "install", component.toString(), "--quiet");
-    commandRunner.run(command, null, environment, consoleListener);
+        Arrays.asList(
+            gcloudPath.toString(), "components", "install", component.toString(), "--quiet");
+    commandRunner.run(command, workingDirectory, environment, consoleListener);
```
elharo: should we make the workingDirectory argument in commandRunner.run not nullable? (It is now.)

chanseokoh: `CommandRunner` is very generic and can run any process with arbitrary arguments, so I think it's fine to accept null for `workingDirectory`.

Issue #641, opened 2018-05-25
No setter methods in StageFlexibleConfiguration et al


elharo:
StageFlexibleConfiguration could be less Nullable if we stop using setter methods in the implementation classes. I think this came up in several other places too including `DefaultStageStandardConfiguration`. A constructor or a builder should work here.

Other classes that need this treatment:

* DefaultGenRepoInfoFileConfiguration
* DefaultDeployConfiguration
* DefaultVersionsSelectionConfiguration
* DefaultRunConfiguration
* DefaultDeployProjectConfigurationConfiguration
* DefaultConfiguration

PR #654, opened 2018-06-06
staging and source directories must be specified

elharo:
more work on #641 @loosebazooka @chanseokoh @briandealwis


src/main/java/com/google/cloud/tools/appengine/api/deploy/StageStandardConfiguration.java
```diff
@@ -16,45 +16,129 @@
...
 import java.io.File;
 import javax.annotation.Nullable;
 
-/** Configuration for {@link AppEngineStandardStaging#stageStandard(StageStandardConfiguration)}. */
-public interface StageStandardConfiguration {
+/**
+ * Arguments needed to stage an App Engine standard environment application. Null return values
+ * indicate that the configuration was not set, and thus assumes the tool default value.
+ */
+public class StageStandardConfiguration {
```
loosebazooka: It seems like if FlexibleStaging is a builder, all configurations should be builders.

elharo: done

elharo: We can. My thought was the nullable fields that still have setters didn't need the builder, but as long as they're primarily for initializers we can put them all in the builder.

I'd rather not allow these fields to be null at all, but that's a more far reaching change and decision for another PR.

src/main/java/com/google/cloud/tools/appengine/api/deploy/DefaultStageStandardConfiguration.java
```diff
@@ -1,157 +0,0 @@
-/*
```
loosebazooka: I would like for the interfaces to remain in, in some way for the build tools at the moment.

elharo: Interfaces can't enforce non-nullness. Do the build tools currently implement this interface?

loosebazooka: yes, to expose it directly to the build config. Oh is it because the annotation doesn't pass through?

elharo: No, it's not the annotation. An implementation can return anything. A class can enforce preconditions and postconditions. An interface can't.

src/main/java/com/google/cloud/tools/appengine/api/deploy/DefaultStageStandardConfiguration.java
```diff
@@ -35,123 +39,165 @@
...
+
+    /** Builds the configuration. */
+    public StageStandardConfiguration build() {
+      if (sourceDirectory == null || stagingDirectory == null) {
+        throw new NullPointerException("Incomplete configuration");
+      }
+      DefaultStageStandardConfiguration stageStandardConfiguration =
+          new DefaultStageStandardConfiguration(sourceDirectory, stagingDirectory);
+
+      stageStandardConfiguration.dockerfile = dockerfile;
```
loosebazooka: I think the pattern, in general, calls for the members of the class to be final, and everything passed through the contructor (as nullable?) Instead of setting them here via scope privilege.

elharo: That is one way to do it, but it's a private implementation detail, not a fundamental feature of builders. As long as a reference to the object doesn't escape before the method returns, there's no reason not to do this. There's also one strong reason to do this. This approach is far less susceptible to accidentally swapping arguments. If we just passed them all into a constructor, would you notice if  jarSplittingExcludes was being passed where deleteJsps was expected and vice versa? Private or not, very long argument lists of the same type are risky.

loosebazooka: I mean you would probably catch that testing your builder. It's just a little strange to me to see it like this.

src/main/java/com/google/cloud/tools/appengine/api/deploy/DefaultStageStandardConfiguration.java
```diff
@@ -35,123 +39,165 @@
...
+      this.sourceDirectory = Preconditions.checkNotNull(sourceDirectory);
+      return this;
+    }
+
+    public Builder setStagingDirectory(File stagingDirectory) {
+      this.stagingDirectory = Preconditions.checkNotNull(stagingDirectory);
+      return this;
+    }
+
+    public void setDockerfile(@Nullable File dockerfile) {
```
briandealwis: Why are the rest of these builder methods terminal (i.e., not returning the `Builder` instance)?

src/main/java/com/google/cloud/tools/appengine/api/deploy/DefaultStageStandardConfiguration.java
briandealwis: Or pass the builder instance into the constructor?

src/main/java/com/google/cloud/tools/appengine/api/deploy/DefaultStageStandardConfiguration.java
elharo: done

src/main/java/com/google/cloud/tools/appengine/api/deploy/DefaultStageStandardConfiguration.java
elharo: I'm not confident enough in the existing test coverage to rely on it exclusively.
Passing the builder instance into the constructor is an interesting trick I hadn't thought of. It seems to flip flop the existing approach.

PR #655, opened 2018-06-07
Verify files to extract are under destination

chanseokoh:
We've been OK since we extract trusted archives (Cloud SDK zips) , but it's better to fix it correctly from the security point of view.

https://snyk.io/research/zip-slip-vulnerability#what-action-should-you-take

I've done actual installation on Linux and Windows.

### DIFF
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/install/Extractor.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/install/Extractor.java
@@ -20,6 +20,7 @@ import com.google.cloud.tools.managedcloudsdk.ProgressListener;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.io.MoreFiles;
 import com.google.common.io.RecursiveDeleteOption;
+import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.logging.Logger;
@@ -80,5 +81,11 @@ final class Extractor {
 
   private void cleanUp(final Path target) throws IOException {
     MoreFiles.deleteRecursively(target, RecursiveDeleteOption.ALLOW_INSECURE);
+  }
+
+  static boolean isTargetInsideDestination(Path destination, Path target) throws IOException {
+    String canonicalDestination = destination.toFile().getCanonicalPath();
+    String canonicalTarget = target.toFile().getCanonicalPath();
+    return canonicalTarget.startsWith(canonicalDestination + File.separator);
   }
 }
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/install/TarGzExtractorProvider.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/install/TarGzExtractorProvider.java
@@ -53,7 +53,11 @@ final class TarGzExtractorProvider implements ExtractorProvider {
     try (TarArchiveInputStream in = new TarArchiveInputStream(gzipIn)) {
       TarArchiveEntry entry;
       while ((entry = in.getNextTarEntry()) != null) {
-        final Path entryTarget = destination.resolve(entry.getName());
+        Path entryTarget = destination.resolve(entry.getName());
+
+        if (!Extractor.isTargetInsideDestination(destination, entryTarget)) {
+          throw new IOException("Blocked unzipping files outside destination: " + entry.getName());
+        }
 
         progressListener.update(1);
         logger.fine(entryTarget.toString());
--- a/src/main/java/com/google/cloud/tools/managedcloudsdk/install/ZipExtractorProvider.java
+++ b/src/main/java/com/google/cloud/tools/managedcloudsdk/install/ZipExtractorProvider.java
@@ -62,6 +62,10 @@ final class ZipExtractorProvider implements ExtractorProvider {
         ZipArchiveEntry entry = zipEntries.nextElement();
         Path entryTarget = destination.resolve(entry.getName());
 
+        if (!Extractor.isTargetInsideDestination(destination, entryTarget)) {
+          throw new IOException("Blocked unzipping files outside destination: " + entry.getName());
+        }
+
         progressListener.update(1);
         logger.fine(entryTarget.toString());
 
--- a/src/test/java/com/google/cloud/tools/managedcloudsdk/install/ExtractorTest.java
+++ b/src/test/java/com/google/cloud/tools/managedcloudsdk/install/ExtractorTest.java
@@ -104,9 +104,10 @@ public class ExtractorTest {
 
   @Test
   public void testIsTargetInsideDestination_targetInside() throws IOException {
-    Assert.assertTrue(Extractor.isTargetInsideDestination(
-        Paths.get("/tmp/../home//user/./destination"),
-        Paths.get("/./bin/../home/user///cache/./../destination/yes")));
+    Assert.assertTrue(
+        Extractor.isTargetInsideDestination(
+            Paths.get("/tmp/../home//user/./destination"),
+            Paths.get("/./bin/../home/user///cache/./../destination/yes")));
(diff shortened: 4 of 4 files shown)


chanseokoh:
I'll add tests for extracting actual zips that exploits this.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Extractor.java
```diff
@@ -81,4 +82,10 @@ ExtractorProvider getExtractorProvider() {
   private void cleanUp(final Path target) throws IOException {
     MoreFiles.deleteRecursively(target, RecursiveDeleteOption.ALLOW_INSECURE);
   }
+
+  static boolean isTargetInsideDestination(Path destination, Path target) throws IOException {
+    String canonicalDestination = destination.toFile().getCanonicalPath();
+    String canonicalTarget = target.toFile().getCanonicalPath();
+    return canonicalTarget.startsWith(canonicalDestination + File.separator);
```
briandealwis: String prefix checking won't catch "/usr/libexec" vs "/usr/lib".  Convert to a `Path` and use `Path.startsWith(Path)`?

src/main/java/com/google/cloud/tools/managedcloudsdk/install/TarGzExtractorProvider.java
```diff
@@ -53,7 +53,11 @@ public void extract(Path archive, Path destination, ProgressListener progressLis
     try (TarArchiveInputStream in = new TarArchiveInputStream(gzipIn)) {
       TarArchiveEntry entry;
       while ((entry = in.getNextTarEntry()) != null) {
-        final Path entryTarget = destination.resolve(entry.getName());
+        Path entryTarget = destination.resolve(entry.getName());
+
+        if (!Extractor.isTargetInsideDestination(destination, entryTarget)) {
+          throw new IOException("Blocked unzipping files outside destination: " + entry.getName());
```
briandealwis: I wondered about using `java.nio.file.AccessDeniedException` rather than the very generic `IOException`.  Or `ZipException` and `ArchiveException`?

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Extractor.java
chanseokoh: We are good because we add `File.separator`. I just consulted the example code for fix coming from the very folks who discovered this vulnerability, so I think we're safe. You might still argue using `Path` seems ideal, but when I tried it, I realized that `Path.startsWith()` will work differently for the case where two paths are identical. So, I guess following the example code is reasonable.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/TarGzExtractorProvider.java
chanseokoh: I think I'll go for `ZipException` for both. `AccessDeniedException` seemed suitable actually, but its constructor expects the file in problem rather than a general error message.

chanseokoh:
I'm not sure why the test fails on Mac. I'll have to push some temporary diagnostic commits to figure out what is happening.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/TarGzExtractorProvider.java
chanseokoh: However, I won't expose `ZipException`, as this is an unusual case that shouldn't happen normally. `ZipException` is an `IOException`, so we don't need to change the method signature.

chanseokoh:
Turns out `/tmp` was expanding to `/Volumes/BuildData/tmpfs/` on Mac.

chanseokoh:
@koto CC'ed per Bug 109861633

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Extractor.java
briandealwis: Ok, I missed the `File.separator`.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/TarGzExtractorProvider.java
briandealwis: `ZipException` doesn't seem right for the TarGz case.  I guess I'd keep `IOException`, or create a new specific exception — `InvalidArchiveEntryException`?  Your call.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Extractor.java
```diff
@@ -81,4 +82,10 @@ ExtractorProvider getExtractorProvider() {
   private void cleanUp(final Path target) throws IOException {
     MoreFiles.deleteRecursively(target, RecursiveDeleteOption.ALLOW_INSECURE);
   }
+
+  static boolean isTargetInsideDestination(Path destination, Path target) throws IOException {
```
briandealwis: Seems odd that arguments are in a different order from what's implied in the method name.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/TarGzExtractorProvider.java
chanseokoh: `IOException` doesn't seem totally out of place at the top level, generically stating the extraction operation failed. The TarGz API throws it if an archive is malformed. If I were to create `InvalidArchiveEntryException`, I'd make it a subclass of `IOException` as well, primarily to avoid exposing it to callers. This exceptional case is only about zips with malicious intent, which I don't want callers to be prepared for handling it in a special way.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Extractor.java
chanseokoh: Switched the order.

src/main/java/com/google/cloud/tools/managedcloudsdk/install/Extractor.java
```diff
@@ -81,4 +82,10 @@ ExtractorProvider getExtractorProvider() {
   private void cleanUp(final Path target) throws IOException {
     MoreFiles.deleteRecursively(target, RecursiveDeleteOption.ALLOW_INSECURE);
   }
+
+  static boolean isTargetInsideDestination(Path target, Path destination) throws IOException {
```
briandealwis: I realized that this is re-canonicalizing the destination for every tar/zip entry.  Extraction already takes some time.

chanseokoh: That's a good point. I'll see what I can do.

chanseokoh: So I've been adding a Javadoc after making this accept a canonicalized destination path like this, and now I feel I would just remove this method and embed the code directly. We are basically talking about the necessity of trading simplicity for optimization, so embedding the code is reasonable and actually better than having a util method with a weird Javadoc.

```java
  /**
   * Check if {@code target} is inside {@code canonicalDestination}
   *
   * @param target path to check if inside {@code canonicalDestination}
   * @param canonicalDestination canonicalized destination path; must be canonicalized beforehand
   *     for this method to work correctly (behavior undefined otherwise)
   */
  // For optimization reasons, this method does not check if "canonicalDestination" is already
  // canonicalized.
  static boolean isTargetInsideDestination(Path target, String canonicalDestination)
```

chanseokoh: Optimized performance. This removes small tests for the method, but we still have end-to-end tests with actual Zip-Slip samples.

PR #657, opened 2018-06-08
0.6.2

loosebazooka:


elharo:
Before cutting a new release we should finish the prs that are waiting for
review.

--
Elliotte Rusty Harold
on a mobile phone so please excuse weird autocorrects

PR #658, opened 2018-06-08
0.6.1

loosebazooka:
Alright, this is the correct version.


loosebazooka:
It just has some features I need in the build plugins, we can always do another release after.

PR #661, opened 2018-06-10
allow omitting docker directory

### LINKED ISSUE #660: StageFlexibleConfiguration docker directory should be nullable
It's not required for flexible staging.

elharo:
fix #660 @loosebazooka @chanseokoh


loosebazooka:
I realize we took out the interfaces earlier (interface StageFlexConfig) in a previous PR, but I think we should leave the interfaces in and have the builders. Obviously this can go in without that, but taking the interfaces out at this point forces a major change in the build tools design which might be more time intensive than we have.

PR #663, opened 2018-06-11
Enhance AppYaml

briandealwis:
The PR enhances the `AppYaml` class to provide roughly the same functionality as the `AppEngineDescriptor` class.  
  - I broke `AppYaml`'s constructor-based loader since it assumes the input comes from a file.  And active constructors aren't great.
    - the only consumers of `AppYaml` outside of `appengine-plugins-core` is CT4E
  - I wondered about moving `AppYaml` into `com.google.cloud.tools.appengine` to be alongside `AppEngineDescriptor`.  The `app.yaml` format doesn't seem to be used outside of App Engine, correct?

(Ideally `AppYaml` and `AppEngineDescriptor` would be a single POJO model object that could be loaded with either SnakeYAML or Jackson/Gson, but that involves quite a bit more work.  And surely there must be some code we can leverage somewhere?)

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
@@ -25,6 +25,7 @@ import com.google.cloud.tools.io.FileUtil;
 import com.google.cloud.tools.project.AppYaml;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import com.google.common.io.MoreFiles;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
@@ -32,9 +33,6 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
-import org.yaml.snakeyaml.error.YAMLException;
-import org.yaml.snakeyaml.parser.ParserException;
-import org.yaml.snakeyaml.scanner.ScannerException;
 
 /** Cloud SDK based implementation of {@link AppEngineFlexibleStaging}. */
 public class CloudSdkAppEngineFlexibleStaging implements AppEngineFlexibleStaging {
@@ -69,7 +67,7 @@ public class CloudSdkAppEngineFlexibleStaging implements AppEngineFlexibleStagin
       copyDockerContext(config, copyService, runtime);
       copyAppEngineContext(config, copyService);
       copyArtifact(config, copyService);
-    } catch (IOException | YAMLException ex) {
+    } catch (IOException ex) {
       throw new AppEngineException(ex);
     }
   }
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/cloud/tools/project/AppYaml.java
```diff
@@ -16,54 +16,113 @@
...
-    }
+  @SuppressWarnings("unchecked")
+  public static AppYaml parse(InputStream input) {
+    // our needs are simple so just load using primitive objects
+    Yaml yaml = new Yaml(new SafeConstructor());
+    Map<String, ?> contents = (Map<String, ?>) yaml.load(input);
+    return new AppYaml(contents);
+  }
+
+  private AppYaml(Map<String, ?> yamlMap) {
```
chanseokoh: `@Nullable` for the parameter?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineFlexibleStaging.java
```diff
@@ -84,7 +85,9 @@ static String findRuntime(StageFlexibleConfiguration config)
         throw new AppEngineException("Invalid Staging Configuration: missing App Engine directory");
       }
       Path appYaml = appEngineDirectory.toPath().resolve(APP_YAML);
-      return new AppYaml(appYaml).getRuntime();
+      try (InputStream input = Files.newInputStream(appYaml)) {
```
elharo: might want to buffer this

src/main/java/com/google/cloud/tools/project/AppYaml.java
```diff
@@ -16,54 +16,113 @@
...
 
-/** Tools for reading app.yaml */
+/** Tools for reading {@code app.yaml}. */
 public class AppYaml {
 
-  private final Map<String, ?> yamlMap;
-
+  private static final String ENVIRONMENT_TYPE_KEY = "env";
   private static final String RUNTIME_KEY = "runtime";
+  private static final String API_VERSION_KEY = "api_version";
```
elharo: Personally, I've never seen the point of replacing these types of strings with named constants. The field names are not more descriptive than the strings themselves

src/main/java/com/google/cloud/tools/project/AppYaml.java
```diff
@@ -16,54 +16,113 @@
...
+
+  private final Map<String, ?> yamlMap;
 
   /**
    * Parse an app.yaml file to an AppYaml object.
    *
-   * @param appYaml the app.yaml file
-   * @throws IOException if reading app.yaml fails due to I/O errors
+   * @param input the input, typically the contents of an {@code app.yaml} file
    * @throws org.yaml.snakeyaml.scanner.ScannerException if reading app.yaml fails while scanning
```
elharo: This should be a checked exception. We shouldn't bubble this.

src/main/java/com/google/cloud/tools/project/AppYaml.java
```diff
@@ -16,54 +16,113 @@
...
    *     due to malformed YAML (undocumented {@link RuntimeException} from {@link Yaml#load})
    * @throws org.yaml.snakeyaml.parser.ParserException if reading app.yaml fails while parsing due
```
elharo: ditto this one. Perhaps not in this PR, but we do need to replace these with a checked exception from our own API. SnakeYAML should not be part of our interface contract. A custom subclass of IOException might be appropriate.

src/main/java/com/google/cloud/tools/project/AppYaml.java
```diff
@@ -16,54 +16,113 @@
...
+  private String getString(String key) {
+    Object value = yamlMap.get(key);
+    return value instanceof String ? (String) value : null;
+  }
+
+  @Nullable
+  @SuppressWarnings("unchecked")
+  private Map<String, ?> getStringMap(String key) {
+    Object value = yamlMap.get(key);
+    return value instanceof Map<?, ?> ? (Map<String, ?>) value : null;
```
chanseokoh: Probably `instanceof Map<String, ?>`.

briandealwis: That causes an error in Eclipse since you can't actually check the generic types.

src/main/java/com/google/cloud/tools/project/AppYaml.java
```diff
@@ -16,54 +16,118 @@
...
-      }
+  @SuppressWarnings("unchecked")
+  public static AppYaml parse(InputStream input) throws AppEngineException {
+    try {
+      // our needs are simple so just load using primitive objects
+      Yaml yaml = new Yaml(new SafeConstructor());
+      Map<String, ?> contents = (Map<String, ?>) yaml.load(input);
+      return new AppYaml(contents);
+    } catch (ScannerException | ParserException ex) {
+      throw new AppEngineException("Malformed 'app.yaml'.", ex);
```
elharo: Might be worth doing

"Malformed app.yaml: " + ex.getMessage()

depending on how informative the SnakeYaml messages are.

briandealwis: They're multi-line and quite verbose.  For example:
```
while parsing a block mapping
 in 'reader', line 1, column 1:
    : m a l f o r m e d !
    ^
expected <block end>, but found ':'
 in 'reader', line 1, column 1:
    : m a l f o r m e d !
    ^
```
I'd expect that these issues would normally be surfaced to the user in some other way.

PR #664, opened 2018-06-11
add in stage flex interface

loosebazooka:


src/main/java/com/google/cloud/tools/appengine/api/deploy/DefaultStageFlexibleConfiguration.java
```diff
@@ -0,0 +1,116 @@
+/*
+ * Copyright 2017 Google Inc.
```
elharo: 2018

loosebazooka:
I guess this needs to be re-approved?

PR #670, opened 2018-06-14
Accept "HEAD" as CloudSdkVersion

### LINKED ISSUE #561: Deal with "HEAD" in cloud SDK versions.
Perhaps throwing an exception for this known case of the SDK isn't the ideal response.

chanseokoh:
Fixes #561.

I should say this may not be an ideal way to deal with HEAD. I've wondered what `getMajorVersion()` etc should return for HEAD, and after some tries, I ended up assigning the max int limit to the individual elements. This has the following consequences.

- A HEAD is just another regular version, nothing special, only having high numbers that are unsurpassable.
- When comparing recency, HEAD is always later (greater) than any non-HEAD version, although in reality this will not hold in most cases.
- One HEAD version is semantically equal to another HEAD version, although this will not be true in reality in almost all cases.

And I also need to verify if the content of the `VERSION` file is exactly `HEAD` without any surplus space.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
@@ -34,7 +34,7 @@ public class CloudSdkVersion implements Comparable<CloudSdkVersion> {
 
   private final String version;
   private final int majorVersion;
-  private final int minorVerion;
+  private final int minorVersion;
   private final int patchVersion;
 
   @Nullable
@@ -52,22 +52,29 @@ public class CloudSdkVersion implements Comparable<CloudSdkVersion> {
     Preconditions.checkNotNull(version, "Null version");
     Preconditions.checkArgument(!version.isEmpty(), "empty version");
 
-    Matcher matcher = SEMVER_PATTERN.matcher(version);
-    if (!matcher.matches()) {
-      throw new IllegalArgumentException(
-          String.format("Pattern \"%s\" is not a valid CloudSdkVersion.", version));
+    if ("HEAD".equals(version)) {
+      majorVersion = Integer.MAX_VALUE;
+      minorVersion = Integer.MAX_VALUE;
+      patchVersion = Integer.MAX_VALUE;
+      preRelease = null;
+      buildIdentifier = null;
+    } else {
+      Matcher matcher = SEMVER_PATTERN.matcher(version);
+      if (!matcher.matches()) {
+        throw new IllegalArgumentException(
+            String.format("Pattern \"%s\" is not a valid CloudSdkVersion.", version));
+      }
+
+      majorVersion = Integer.parseInt(matcher.group("major"));
+      minorVersion = Integer.parseInt(matcher.group("minor"));
+      patchVersion = Integer.parseInt(matcher.group("patch"));
+
+      preRelease =
+          matcher.group("prerelease") != null
+              ? new CloudSdkVersionPreRelease(matcher.group("prerelease"))
+              : null;
+      buildIdentifier = matcher.group("build");
     }
-
-    majorVersion = Integer.parseInt(matcher.group("major"));
-    minorVerion = Integer.parseInt(matcher.group("minor"));
-    patchVersion = Integer.parseInt(matcher.group("patch"));
-
-    preRelease =
-        matcher.group("prerelease") != null
-            ? new CloudSdkVersionPreRelease(matcher.group("prerelease"))
-            : null;
-    buildIdentifier = matcher.group("build");
-
     this.version = version;
   }
 
@@ -114,9 +121,9 @@ public class CloudSdkVersion implements Comparable<CloudSdkVersion> {
     Preconditions.checkNotNull(other);
 
     // First, compare required fields
-    List<Integer> mine = ImmutableList.of(majorVersion, minorVerion, patchVersion);
+    List<Integer> mine = ImmutableList.of(majorVersion, minorVersion, patchVersion);
     List<Integer> others =
-        ImmutableList.of(other.getMajorVersion(), other.getMinorVerion(), other.getPatchVersion());
+        ImmutableList.of(other.getMajorVersion(), other.getMinorVersion(), other.getPatchVersion());
     for (int i = 0; i < mine.size(); i++) {
       int result = mine.get(i).compareTo(others.get(i));
       if (result != 0) {
@@ -142,7 +149,7 @@ public class CloudSdkVersion implements Comparable<CloudSdkVersion> {
 
   @Override
   public int hashCode() {
-    return Objects.hash(majorVersion, minorVerion, patchVersion, preRelease, buildIdentifier);
+    return Objects.hash(majorVersion, minorVersion, patchVersion, preRelease, buildIdentifier);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


loosebazooka:
You could trim the string if you're worried about spaces

chanseokoh:
I added `trim()` just in case (although https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/561#issuecomment-397367152 shows it's precisely `HEAD`).

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
```diff
@@ -52,22 +52,29 @@ public CloudSdkVersion(String version) {
...
-    if (!matcher.matches()) {
-      throw new IllegalArgumentException(
-          String.format("Pattern \"%s\" is not a valid CloudSdkVersion.", version));
-    }
-
-    majorVersion = Integer.parseInt(matcher.group("major"));
-    minorVerion = Integer.parseInt(matcher.group("minor"));
-    patchVersion = Integer.parseInt(matcher.group("patch"));
+    if ("HEAD".equals(version.trim())) {
+      majorVersion = Integer.MAX_VALUE;
```
elharo: Instead of faking a version number we could change the comparable logic to understand that HEAD is later than other versions.

chanseokoh:
> Do we know why/when the VERSION file contains the string "HEAD"?

An internal dogfood version is one case, I'm reasonable sure. I don't know about any other case. If it is HEAD, at least it indicates that the SDK was not installed from official channels. So I believe this is for internal support.

So when first installing a dogfood version, it will be the latest one, but as new official releases roll out, the dogfood version will be out-dated until the user manually updates it. Currently, the user can't use our tools with a dogfood version at all, since we throw an exception with a version parsing error.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
chanseokoh: It's not a big deal to have such a dedicated logic, but we still have the question of which integer value to return when clients call `getMajorVersion()`, `getMinorVersion()`, and `getPatchVersion()`. And, we should also consider that clients may use these method to determine minimal versions, for example. That is,

```java
if (version.getMajorVersion() > 198
    || version.getMajorVersion() > 197 && version.getMinorVersion() > 3) {
   // minimum version requirement
}
```

Returning large values for HEAD seems not a terrible idea, as the main purpose as I understand is to unblock the internal users from using our tools by making the check for the minimum SDK version requirement pass.

chanseokoh: That said, another conceptually ideal way is to make `getMajorVersion()` and such return another layer of abstraction on top of `int` to include a special value for the HEAD notion, but considering that this is mainly for internal support (and maybe an SDK manually built from source, which would extremely rare), I don't think it's worth the complexity.

briandealwis:
Maybe we should just get rid of the methods that expose the version fields.  So it's comparable, but that's it.

patflynn:
That sounds pretty good to me. +1.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
```diff
@@ -142,7 +155,8 @@ public int compareTo(CloudSdkVersion other) {
 
   @Override
   public int hashCode() {
-    return Objects.hash(majorVersion, minorVerion, patchVersion, preRelease, buildIdentifier);
+    return Objects.hash(
```
briandealwis: Hmm, should we be hashing `version`?  Consider "0.0.1" vs "0.0.01": we previously considered them the same.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
```diff
@@ -164,24 +178,12 @@ public boolean equals(Object obj) {
     CloudSdkVersion otherVersion = (CloudSdkVersion) obj;
 
     return Objects.equals(majorVersion, otherVersion.majorVersion)
```
briandealwis: One advantage of using HEAD = { -1, -1, -1 } is that we can't create a version from "-1.-1.-1" whereas we could from "999999.999999.999999".

src/test/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersionTest.java
```diff
@@ -71,49 +79,34 @@ public void testConstructor_missingRequiredNumbers() {
   @Test
   public void testConstructor_requiredNumbersOnly() {
     CloudSdkVersion version = new CloudSdkVersion("0.1.0");
-    assertEquals(0, version.getMajorVersion());
-    assertEquals(1, version.getMinorVerion());
-    assertEquals(0, version.getPatchVersion());
```
briandealwis: It's still useful for the tests to verify these fields though, right? (and ditto for the rest)

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
chanseokoh: I was thinking of a possibility in the future another version string such as "unknown" or a commit SHA hash could be associated with the value -1. But considering that `new CloudSdkVersion("0.0.1").compareTo(new CloudSdkVersion("0.0.01")) == true`, it makes sense to drop `version` from computing the hash.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersion.java
chanseokoh: Yup.

src/test/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersionTest.java
chanseokoh: That's true, but I don't really want to make the fields package-protected only for testing. Making a method package-protected, I'm a little bit more lenient, OTOH. We have `compareTo()` tests, so we are somewhat covered.

chanseokoh:
Not sure why the Travis push build is failing all of a sudden. Seems unrelated.

```
[ERROR] Failed to execute goal org.jacoco:jacoco-maven-plugin:0.8.1:report (generation) on project appengine-plugins-core: Execution generation of goal org.jacoco:jacoco-maven-plugin:0.8.1:report failed: A required class was missing while executing org.jacoco:jacoco-maven-plugin:0.8.1:report: org/apache/commons/collections/ExtendedProperties
```

src/test/java/com/google/cloud/tools/appengine/cloudsdk/serialization/CloudSdkVersionTest.java
briandealwis: Problem is that you don't know that the parsing is correct.  Especially since `toString()` just returns the original version string.  Maybe if `toString()` computed the result then you could compare that.

chanseokoh: But the test will depend on the implementation details of `toString()`, which is exactly the current situation.

briandealwis: With your patch as is, there is no way to verify that the version parsing has correctly identified the version components.  Having `toString()` as a proxy seems ok to me if you don't want to expose `*Version` and friends as `@VisibleForTesting`.

chanseokoh: I can think of some ways to verify the version parsing indirectly. I'll add some dedicated tests.

chanseokoh:
I still need to figure out why the Travis PR build is failing. Seems like a bug of some plugin.

chanseokoh:
It's passing now. @briandealwis PTAL

chanseokoh:
Fair enough. I still think the new tests are much stronger than the old test that suffers from the same pitfalls, but I've restored the old test with `@VisibleForTesting`. Not a big deal, after all.

PR #673, opened 2018-06-15
Update maven-bundle-plugin to 3.5.0

briandealwis:
Fixes missed package references as seen in https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/3165


briandealwis:
Using 3.5.0 fixes the issue (https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/3165#issuecomment-397710785): we could continue to use the multi-catch, and ignore #674.

chanseokoh:
Sounds good. Let's ignore #674 then.

PR #674, opened 2018-06-15
Cause AppYaml#parse() to wrap all YAMLException as AppEngineException

briandealwis:
Wrap all SnakeYAML exceptions in our `AppEngineException`.

### DIFF
--- a/src/main/java/com/google/cloud/tools/project/AppYaml.java
+++ b/src/main/java/com/google/cloud/tools/project/AppYaml.java
@@ -23,8 +23,7 @@ import java.util.Map;
 import javax.annotation.Nullable;
 import org.yaml.snakeyaml.Yaml;
 import org.yaml.snakeyaml.constructor.SafeConstructor;
-import org.yaml.snakeyaml.parser.ParserException;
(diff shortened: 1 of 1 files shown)


briandealwis:
This change isn't actually required with the fix in #673. But SnakeYAML's exceptions are all `RuntimeException`s so it's hard to be sure that we're catching all relevant exceptions here.  Just looking at some of the possible YAMLExceptions and tracing them back, I found:

  - `DuplicateKeyException`, thrown by `SafeConstructor`, which we use in [`AppYaml`](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/master/src/main/java/com/google/cloud/tools/project/AppYaml.java#L54)
  - `ComposerException` which is on the call-chain of the [`Yaml#load(InputStream)`](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/master/src/main/java/com/google/cloud/tools/project/AppYaml.java#L55)

And some of the others also appear possible.

chanseokoh:
Hmm... May not be a bad idea to catch them all.

PR #679, opened 2018-06-20
fix source config command

loosebazooka:


chanseokoh:
There are formatting errors.

loosebazooka:
ah my pre-commit hook wasn't working as intended

PR #689, opened 2018-07-20
update libraries to 1.37.1

elharo:
@etanshaul


elharo:
need to update; 1.37.1 just came out

PR #694, opened 2018-08-22
update libraries

### LINKED ISSUE #691: Upgrade google-cloud-java to 0.59.0
I think there are several new beta APIs in this release we don't have yet including Google Cloud Java Client for Container Analysis, KMS, and Kubernetes

elharo:
@loosebazooka @coollog part of  #691

### DIFF
--- a/src/test/java/com/google/cloud/tools/libraries/LibrariesTest.java
+++ b/src/test/java/com/google/cloud/tools/libraries/LibrariesTest.java
@@ -170,7 +170,8 @@ public class LibrariesTest {
               + "/"
               + coordinates.getString("artifactId")
               + "/"
-              + coordinates.getString("version");
+              + coordinates.getString("version")
+              + "/";
       assertReachable(repo);
     }
   }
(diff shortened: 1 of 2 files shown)


loosebazooka:
added in @chanseokoh @etanshaul as users of this part of the lib

src/test/java/com/google/cloud/tools/libraries/LibrariesTest.java
```diff
@@ -170,7 +170,8 @@ public void testVersionExists() throws IOException {
               + "/"
               + coordinates.getString("artifactId")
               + "/"
-              + coordinates.getString("version");
+              + coordinates.getString("version")
+              + "/";
```
briandealwis: Would it make sense to change `assertReachable()` to call `setFollowRedirects(true)` on the HTTP connection?  Probably not — we want these URLs to be canonical.

elharo: I was getting 404s for some reason, not 300's

PR #709, opened 2018-09-26
don't die on bad version string

### LINKED ISSUE #708: Pattern found in the Cloud SDK version file could not be parsed: 218.0.0.213974472
com.google.cloud.tools.appengine.cloudsdk.CloudSdkVersionFileParseException: Pattern found in the Cloud SDK version file could not be parsed: 218.0.0.213974472
	at com.google.cloud.tools.appengine.cloudsdk.CloudSdk.getVersion(CloudSdk.java:95)
	at com.google.cloud.tools.appengine.cloudsdk.CloudSdk.validateCloudSdkVersion(CloudSdk.java:187)
	at com.google.cloud.tools.appengine.cloudsdk.CloudSdk.validateCloudSdk(CloudSdk.java:181)
	at com.google.cloud.tools.appengine.cloudsdk.GcloudRunner.run(GcloudRunner.java:75)
	at com.google.cloud.tools.appengine.cloudsdk.CloudSdkAppEngineDeployment.deploy(CloudSdkAppEngineDeployment.java:88)
	at com.google.cloud.tools.intellij.appengine.java.cloud.AppEngineDeploy.deploy(AppEngineDeploy.java:152)
	at com.google.cloud.tools.intellij.appengine.java.cloud.executor.AppEngineStandardDeployTask.lambda$deploy$0(AppEngineStandardDeployTask.java:126)
	at com.google.cloud.tools.appengine.cloudsdk.process.LegacyProcessHandler.syncRun(LegacyProcessHandler.java:144)
	at com.google.cloud.tools.appengine.cloudsdk.process.LegacyProcessHandler.access$200(LegacyProcessHandler.java:31)
	at com.google.cloud.tools.appengine.cloudsdk.process.LegacyProcessHandler$3.run(LegacyProcessHandler.java:163)
Caused by: java.lang.IllegalArgumentException: Pattern "218.0.0.213974472" is not a valid CloudSdkVersion.
	at com.google.cloud.tools.appengine.cloudsdk.serialization.CloudSdkVersion.<init>(CloudSdkVersion.java:66)
	at com.google.cloud.tools.appengine.cloudsdk.CloudSdk.getVersion(CloudSdk.java:91)
	... 9 more
Exception class name:
java.lang.IllegalArgumentException
Exception message:
Pattern found in the Cloud SDK version file could not be parsed: 218.0.0.213974472
Throw file name:
CloudSdk.java
Throw line number:
95
Throw class name:
com.google.cloud.tools.appengine.cloudsdk.CloudSdk
Throw method name:
getVersion

elharo:
fix #708 @loosebazooka @patflynn In this case we were correctly noticing a bad release that was pulled. I'm not 100% sure we want to fix this. Arguably we did exactly what we should have done when faced with a bad push.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -189,7 +189,7 @@ public class CloudSdk {
         throw new CloudSdkOutOfDateException(version, MINIMUM_VERSION);
       }
     } catch (CloudSdkVersionFileParseException ex) {
-      // unusual string in version file. Continue onward and hope.
+      logger.log(Level.WARNING, ex.getMessage(), ex);
     } catch (CloudSdkVersionFileNotFoundException ex) {
       // this is likely a version of the Cloud SDK prior to when VERSION files were introduced
       throw new CloudSdkOutOfDateException(MINIMUM_VERSION);
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
@@ -188,6 +188,8 @@ public class CloudSdk {
       if (version.compareTo(MINIMUM_VERSION) < 0) {
         throw new CloudSdkOutOfDateException(version, MINIMUM_VERSION);
       }
+    } catch (CloudSdkVersionFileParseException ex) {
+      // unusual string in version file. Continue onward and hope.
     } catch (CloudSdkVersionFileNotFoundException ex) {
       // this is likely a version of the Cloud SDK prior to when VERSION files were introduced
       throw new CloudSdkOutOfDateException(MINIMUM_VERSION);
--- a/src/test/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkTest.java
+++ b/src/test/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkTest.java
@@ -28,7 +28,6 @@ import static org.mockito.Mockito.when;
 
 import com.google.cloud.tools.appengine.cloudsdk.serialization.CloudSdkVersion;
 import com.google.common.io.Files;
-
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
@@ -95,7 +94,7 @@ public class CloudSdkTest {
   public void testGetVersion_fileContentInvalid() throws IOException {
     String fileContents = "this is not a valid version string";
     writeVersionFile(fileContents);
-    try { 
+    try {
       sdk.getVersion();
       fail();
     } catch (CloudSdkVersionFileException ex) {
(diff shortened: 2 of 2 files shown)


etanshaul:
I'm with you that this may not be the fix we want. The tools (IDEs, build) would still want to know about this so they can inform the users (i.e. show a warning in IntelliJ, Eclipse).

@loosebazooka

elharo:
It would be easy to bubble the `CloudSdkVersionFileParseException` so clients can message the user if they choose. That said, I'm not sure it's worth the extra work on the IDE side to handle this. This was a one off mistake by the Cloud SDK, not a regular occurrence.

etanshaul:
Isn't it already bubbled up to the clients? 

The user report that started this discussion came from the IDE calling `CloudSdkAppEngineDeployment#deploy` which did not catch the `CloudSdkVersionFileParseException` causing the IDE to blow up.

I guess I'm suggesting that we don't make the fix here, and rather have the clients handle the exception:

- Show are error to the user, perhaps with a suggestion to manually install the SDK
- Notify us somehow (e.g. via an analytics ping in IntelliJ)

@elharo @patflynn @loosebazooka wdyt?

etanshaul:
well there is the mechanism in IntelliJ that provided this user report, but currently its only set up to report when an exception bubbles up to the IDE and the user clicks a button to "report error to google". So in its current form its not ideal when we are trying to prevent the IDE error in the first place. 

I'm not sure what Eclipse provides.

patflynn:
I'm not convinced we should throw an exception during a Cloud SDK operation because we can't parse the version string. 

Ideally clients would be able to do version validation through some separate API.

ivanporty:
As far as I can remember, IJ checks if Managed SDK is installed, and this involves checking the version file, but it does not rely on `validate` family of methods. If version is incorrect or unexpected, it will attempt to re-install. But this method is not used by Managed SDK installer/updater.

patflynn:
@etanshaul no I'm saying we shouldn't throw an exception for non-critical validation errors such as a bad version string.  Clients that want to do version validation should be able to inquire into the version through some other mechanism.

etanshaul:
aha ok makes sense to me. In that case lgtm on this PR which effectively accomplishes that. 

As @ivanporty mentioned, we do check the version via other means already in IJ - so we should look into that and make sure we handle these cases properly on the client side.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
```diff
@@ -188,6 +188,8 @@ private void validateCloudSdkVersion()
       if (version.compareTo(MINIMUM_VERSION) < 0) {
         throw new CloudSdkOutOfDateException(version, MINIMUM_VERSION);
       }
+    } catch (CloudSdkVersionFileParseException ex) {
+      // unusual string in version file. Continue onward and hope.
```
etanshaul: could we log a message here?

chanseokoh:
There is another method that does the check: [CloudSdk.getVersion()](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/v0.6.6/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java#L76)

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java
elharo: done

Issue #717, opened 2018-10-10
Run nightly tests to catch URL moves and out-of-date artifacts


briandealwis:

briandealwis:
I've configured the Travis job to rebuild master nightly.

chanseokoh:
> I've configured the Travis job to rebuild master nightly.

Interesting. How/where did you do it? Is there a doc?

chanseokoh:
Also, will we notified if the nightly builds fail?

loosebazooka:
This should also help us catch the sdk breakages.

briandealwis:
Go to the [Travis page](https://travis-ci.org/GoogleCloudPlatform/appengine-plugins-core) and open _More options > Settings_ and look at the bottom for _Cron Jobs_.

JoeWang1127:
close as not planned

PR #718, opened 2018-10-10
Verify that libraries.json artifacts are latest available

briandealwis:
Compares listed versions with the latest listed on Maven Central and fails listing the out-of-date artifacts.

Towards #717


loosebazooka:
Could the up to date check perhaps be run in another profile?

briandealwis:
Or could use the maven-failsafe-plugin.

briandealwis:
More useful would be a bot that would propose PRs when newer dependencies are available.

Issue #723, opened 2018-10-26
Change Config classes to Value classes with builders.


loosebazooka:
Replace and simplify Config Classes. Use `Path` instead of `File`
- [x] Infra to support it (#724) 
- [x] DeployConfiguration (#724)
- [x] DeployProjectConfigurationConfiguration, StageStandardConfiguration, StageFlexibleConfiguration (#727)
- [x] Update Deploy classes to use Path instead of File (#728)
- [x] Update Run/Stop Config (also use Path in those classes) (#729)
- [x] Update Versions Config (#731)
- [x] Remove "set" from setters on the builder (#732)
- [x] Update GenRepoInfo Config (#733) 
- [x] Remove last of `File` references from appengine.api package (#733)

DONE!

elharo:
I prefer not doing this. It adds a dependency for code we can easily provide ourselves. In general I find magic generation frameworks to make code more opaque and harder to follow.

loosebazooka:
The generated code is at compile time and is packaged with the Jar, but yeah, someone will have to look in `target/generated-sources/` to see them. These are value/builder classes with no logic, which is exactly where this framework is useful is removing excess boilerplate code.

loosebazooka:
Update, we didn't use autobuilder. But all config's are now final value classes with builders. Lists might still be mutable. Might have to do something about that in the future.

PR #727, opened 2018-10-30
Update more deployment configs

loosebazooka:
depends on #724, needs to be rebased on that when merged.
part of #723


src/main/java/com/google/cloud/tools/appengine/api/deploy/DeployProjectConfigurationConfiguration.java
```diff
@@ -16,16 +16,71 @@
...
 
 /** Configuration for {@link AppEngineDeployment} project-level yaml deployments. */
-public interface DeployProjectConfigurationConfiguration extends Configuration {
+public class DeployProjectConfigurationConfiguration {
 
+  private final File appEngineDirectory;
+  @Nullable private final String projectId;
+  @Nullable private final String server;
+
+  private DeployProjectConfigurationConfiguration(
```
elharo: We might want to use a Path here.

src/main/java/com/google/cloud/tools/appengine/api/deploy/StageFlexibleConfiguration.java
```diff
@@ -20,21 +20,82 @@
 import javax.annotation.Nullable;
 
 /** Configuration for {@link AppEngineFlexibleStaging#stageFlexible(StageFlexibleConfiguration)}. */
-public interface StageFlexibleConfiguration {
+public class StageFlexibleConfiguration {
+
+  private final File appEngineDirectory;
```
elharo: Could use paths here.

loosebazooka:
Changing to `path`, it turns out, is a pretty big change, gonna post it as a separate PR.

PR #730, opened 2018-10-31
Add dev profile to pom.xml

loosebazooka:
This allows a developer to run
```
./mvnw verify -P dev
```
and it will autoformat and skip long tests.


loosebazooka:
I'd rather not ignore all tests by default. We can discuss a better default though.

PR #731, opened 2018-10-31
Update VersionsConfig, remove Base Config class

loosebazooka:
more of #723


src/main/java/com/google/cloud/tools/appengine/api/versions/VersionsListConfiguration.java
```diff
@@ -16,15 +16,65 @@
 
 package com.google.cloud.tools.appengine.api.versions;
 
-import com.google.cloud.tools.appengine.api.Configuration;
 import javax.annotation.Nullable;
 
-/** Configuration for {@link AppEngineVersions#list(VersionsListConfiguration)}. */
-public interface VersionsListConfiguration extends Configuration {
+/** Plain Java bean implementation of {@link VersionsListConfiguration}. */
```
briandealwis: It's an implementation of itself?

src/main/java/com/google/cloud/tools/appengine/api/versions/VersionsListConfiguration.java
```diff
@@ -16,15 +16,65 @@
...
+  }
+
+  public static class Builder {
+    @Nullable private Boolean hideNoTraffic;
+    @Nullable private String projectId;
+    @Nullable private String service;
+
+    private Builder() {}
+
+    public Builder setHideNoTraffic(@Nullable Boolean hideNoTraffic) {
```
etanshaul: just wondering (in part so I know for other projects too), but is it the convention for our builders to use `[set]HideNoTraffic` instead of `hideNoTraffic`? Most implementations I've seen use the later.

src/main/java/com/google/cloud/tools/appengine/api/versions/VersionsListConfiguration.java
loosebazooka: gah, naming

src/main/java/com/google/cloud/tools/appengine/api/versions/VersionsListConfiguration.java
loosebazooka: It's what we've been using, also since it's a `Boolean` it can be null (which then delegates to the command line default for that flag).

src/main/java/com/google/cloud/tools/appengine/api/versions/VersionsSelectionConfiguration.java
```diff
@@ -16,16 +16,65 @@
...
+      this.service = service;
+      return this;
+    }
+
+    public Builder setProjectId(@Nullable String projectId) {
+      this.projectId = projectId;
+      return this;
+    }
+
+    public VersionsSelectionConfiguration build() {
```
etanshaul: is projectId a required field? if so maybe add a Precondition?

loosebazooka: no, it's optional. Technically the code can delegate to gcloud. How we enforce this in the tooling is a different matter.

src/main/java/com/google/cloud/tools/appengine/api/versions/VersionsListConfiguration.java
loosebazooka: I misread the question. But I'm not opposed to that style, let me see what the styleguide recommendation is.

PR #733, opened 2018-10-31
Update GenRepoInfo configs and remove last of the File references

loosebazooka:
part of #723


src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/Args.java
```diff
@@ -95,26 +94,14 @@
...
-    if (file != null) {
-      return path(name, file.toPath());
-    }
-    return Collections.emptyList();
-  }
-
   /**
    * Produces the flag form of a path object, using {@link Path#toString()}.
    *
    * @return {@code [--name, path.toString()]} or {@code []} if path is null or not set.
```
etanshaul: javadoc updated too to match changes?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/Args.java
```diff
@@ -95,26 +94,14 @@
...
    */
   static List<String> path(String name, @Nullable Path path) {
     if (path != null && !path.toString().isEmpty()) {
-      return Arrays.asList("--" + name, path.toString());
+      return Arrays.asList("--" + name, path.toAbsolutePath().toString());
```
etanshaul: could this cause any behavior changes for the callers or is it the same thing?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/Args.java
loosebazooka: ah, yeah keep missing those.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/internal/args/Args.java
loosebazooka: I looked through it, and it seemed fine. But I'll go through and check again. There actually weren't too many callers.

loosebazooka: so it looks like this should be okay, the callers (in tests) were all provided with absolute paths. And all commands that execute with this parameter should be taking absolute paths anyway.

loosebazooka: I've pushed an update to the tests to make it explicit.

PR #736, opened 2018-11-08
Add staging for standard builds artifact style

loosebazooka:
Uses same path as flex for deployment (app.yaml). This also
renames the *FlexibleStaging operations/config ~~*ArtifactStaging~~ *ArchiveStaging
cc: @ludoch 
The implementation of this has changed (Nov 19th) :
1. extraFiles are configured through extraFilesDirectory parameter in the ArtifactStagingConfig object
2. "src/main/appengine" still only contains yamls and staging only pulls app.yaml out of there.

A user might use this (for example in gradle) like this:

```
appengine {
  staging {
    extraFilesDirectory = "src/main/my-extra-files"
  }
}
```

and put a bunch of stuff in there that will be staged and deployed with their application.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/api/deploy/AppEngineArtifactStaging.java
+++ b/src/main/java/com/google/cloud/tools/appengine/api/deploy/AppEngineArtifactStaging.java
@@ -22,7 +22,7 @@ import com.google.cloud.tools.appengine.api.AppEngineException;
  * Main interface for staging Java application for App Engine flexible environment before
  * deployment.
  */
-public interface AppEngineFlexibleStaging {
+public interface AppEngineArtifactStaging {
 
   /**
    * Stages a Java JAR/WAR artifact for App Engine flexible environment deployment. Copies app.yaml,
@@ -31,5 +31,5 @@ public interface AppEngineFlexibleStaging {
    * @param config Specifies artifacts and staging destination
    * @throws AppEngineException When staging fails
    */
-  void stageFlexible(StageFlexibleConfiguration config) throws AppEngineException;
+  void stageFlexible(StageArtifactConfiguration config) throws AppEngineException;
 }
--- a/src/main/java/com/google/cloud/tools/appengine/api/deploy/StageArtifactConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/appengine/api/deploy/StageArtifactConfiguration.java
@@ -20,15 +20,15 @@ import com.google.common.base.Preconditions;
 import java.nio.file.Path;
 import javax.annotation.Nullable;
 
-/** Configuration for {@link AppEngineFlexibleStaging#stageFlexible(StageFlexibleConfiguration)}. */
-public class StageFlexibleConfiguration {
+/** Configuration for {@link AppEngineArtifactStaging#stageFlexible(StageArtifactConfiguration)}. */
+public class StageArtifactConfiguration {
 
   private final Path appEngineDirectory;
   @Nullable private final Path dockerDirectory;
   private final Path artifact;
   private final Path stagingDirectory;
 
-  private StageFlexibleConfiguration(
+  private StageArtifactConfiguration(
       Path appEngineDirectory,
       @Nullable Path dockerDirectory,
       Path artifact,
@@ -83,14 +83,14 @@ public class StageFlexibleConfiguration {
       this.stagingDirectory = stagingDirectory;
     }
 
-    public StageFlexibleConfiguration.Builder dockerDirectory(@Nullable Path dockerDirectory) {
+    public StageArtifactConfiguration.Builder dockerDirectory(@Nullable Path dockerDirectory) {
       this.dockerDirectory = dockerDirectory;
       return this;
     }
 
-    /** Build a {@link StageFlexibleConfiguration}. */
-    public StageFlexibleConfiguration build() {
-      return new StageFlexibleConfiguration(
+    /** Build a {@link StageArtifactConfiguration}. */
+    public StageArtifactConfiguration build() {
+      return new StageArtifactConfiguration(
           this.appEngineDirectory, this.dockerDirectory, this.artifact, this.stagingDirectory);
     }
   }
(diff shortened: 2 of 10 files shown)


src/main/java/com/google/cloud/tools/appengine/api/deploy/AppEngineArtifactStaging.java
```diff
@@ -22,7 +22,7 @@
  * Main interface for staging Java application for App Engine flexible environment before
  * deployment.
  */
-public interface AppEngineFlexibleStaging {
+public interface AppEngineArtifactStaging {
```
elharo: artifact sounds like a Maven term, not an App Engine term. Maybe it makes sense here, but is there a more standard App Engine term for this thing? or even just rop it completely so `AppEngineStaging`? I'm not sure.

src/main/java/com/google/cloud/tools/appengine/api/deploy/StageArtifactConfiguration.java
```diff
@@ -20,15 +20,15 @@
 import java.nio.file.Path;
 import javax.annotation.Nullable;
 
-/** Configuration for {@link AppEngineFlexibleStaging#stageFlexible(StageFlexibleConfiguration)}. */
-public class StageFlexibleConfiguration {
+/** Configuration for {@link AppEngineArtifactStaging#stageFlexible(StageArtifactConfiguration)}. */
+public class StageArtifactConfiguration {
```
elharo: AppEngineStagingConfiguration or perhaps just StagingConfiguration?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineArtifactStaging.java
```diff
@@ -19,35 +19,43 @@
 import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
 
 import com.google.cloud.tools.appengine.api.AppEngineException;
-import com.google.cloud.tools.appengine.api.deploy.AppEngineFlexibleStaging;
-import com.google.cloud.tools.appengine.api.deploy.StageFlexibleConfiguration;
+import com.google.cloud.tools.appengine.api.deploy.AppEngineArtifactStaging;
+import com.google.cloud.tools.appengine.api.deploy.StageArtifactConfiguration;
```
elharo: Staging sounds more right than Stage here

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineArtifactStaging.java
```diff
@@ -61,19 +69,46 @@ public void stageFlexible(StageFlexibleConfiguration config) throws AppEngineExc
...
-      copyDockerContext(config, copyService, runtime);
-      copyAppEngineContext(config, copyService);
-      copyArtifact(config, copyService);
+      if ("flex".equals(env)) {
+        // stage for flex
+        copyDockerContext(config, copyService, runtime);
+        copyAppEngineContextFlex(config, copyService);
+        copyArtifact(config, copyService);
+      } else if ("java11".equals(runtime)) {
+        // stage for appengine standard in artifact mode
```
elharo: is this "artifact mode" something we define in docs?

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineArtifactStaging.java
```diff
@@ -124,7 +159,23 @@ static void copyAppEngineContext(StageFlexibleConfiguration config, CopyService
...
+  static void copyAppEngineContextStandard(
+      StageArtifactConfiguration config, CopyService copyService)
+      throws IOException, AppEngineException {
+    Path appYaml = config.getAppEngineDirectory().resolve(APP_YAML);
+    if (!Files.exists(appYaml)) {
+      throw new AppEngineException(APP_YAML + " not found in the App Engine directory.");
+    }
+    Path appEngineDirectory = config.getAppEngineDirectory();
+    Path stagingDirectory = config.getStagingDirectory();
+    System.out.println(appEngineDirectory + " " + stagingDirectory);
```
elharo: should this be a logger? I'm thinking about the usage of this in an IDE environment where the main UI is not System.out.

src/test/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineArtifactStagingTest.java
```diff
@@ -89,7 +94,7 @@ public void testCopyDockerContext_runtimeJavaNoWarning() throws AppEngineExcepti
   public void testCopyDockerContext_noDocker() throws AppEngineException, IOException {
     when(config.getDockerDirectory()).thenReturn(null);
 
-    CloudSdkAppEngineFlexibleStaging.copyDockerContext(config, copyService, "java");
+    CloudSdkAppEngineArtifactStaging.copyDockerContext(config, copyService, "java");
```
elharo: What ArtifactStaging do we have that is not CloudSDK based?

loosebazooka: This is a legacy thing I guess. There's the Interfaces which can be removed
```
CloudSdkAppEngineArtifactStaging implements AppEngineArtifactStaging
```
but I'm not covering that in this PR.

src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineArtifactStaging.java
loosebazooka: oh this is a mistake, I was testing something.

loosebazooka:
Need to fix the kokoro build at what appears to be the kokoro config level.

elharo:
This does look like something we should dig into and fix first:

+ sudo /opt/google-cloud-sdk/bin/gcloud components update
ERROR: gcloud crashed (TypeError): unsupported operand type(s) for +=: 'int' and 'NoneType'

briandealwis:
That's https://issuetracker.google.com/issues/119096137

patflynn:
@loosebazooka it'd be nice to provide a little more context to this PR.  What behavior is it enabling? Why?

src/test/java/com/google/cloud/tools/appengine/api/deploy/StageArchiveConfigurationTest.java
```diff
@@ -34,12 +34,12 @@ public void setUp() {
     // files are files and directories are directories?
     // should we use paths instead?
     configuration =
-        StageFlexibleConfiguration.builder(file, file, file).dockerDirectory(file).build();
+        StageArchiveConfiguration.builder(file, file, file).dockerDirectory(file).build();
```
elharo: The arguments here are a little weird, the same thing four times? but I guess that's not new

PR #737, opened 2018-11-19
0.7.0

loosebazooka:
Mostly just has updates to the "Config" objects. Want to get a headstart on updating the plugins.


loosebazooka:
Waiting till build is fixed to run this update.

PR #738, opened 2018-11-19
Update gcp-ubuntu build with gcloud update workaround

loosebazooka:


loosebazooka:
Also might need to update the release build script. Will do in a followup.

loosebazooka:
yeah also apparently fine on mac too.

PR #744, opened 2018-12-04
Update getGaeRuntime to handle java11

### LINKED ISSUE #745: inject devappserver environment correctly (like KickStart.java)
Replace https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/master/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java#L280

Perhaps it should look something like this (@ludoch):
```java
static String getGaeRuntimeJava(boolean isSandboxEnforced) {
    String gaeRuntime = "java7";
    if (System.getProperty("java.specification.version").equals("11")) {
      gaeRuntime = "java11";
    } else if (!isSandboxEnabled) {
      gaeRuntime = "java8";
    }
    return gaeRuntime;
}
```

loosebazooka:
Minor fix to injected dev appserver variables.

Looks like kickstart might be doing this a little differently.

fixes #745

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
+++ b/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
@@ -137,7 +137,7 @@ public class CloudSdkAppEngineDevServer1 implements AppEngineDevServer {
               + Joiner.on(",").withKeyValueSeparator("=").join(appEngineEnvironment));
     }
 
-    String gaeRuntime = getGaeRuntimeJava(!isSandboxEnforced);
+    String gaeRuntime = getGaeRuntimeJava(config.getServices());
     appEngineEnvironment.putAll(getLocalAppEngineEnvironmentVariables(gaeRuntime));
 
     Map<String, String> configEnvironment = config.getEnvironment();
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdkAppEngineDevServer1.java
```diff
@@ -271,14 +271,29 @@ boolean isSandboxEnforced(List<Path> services) throws AppEngineException {
...
+  static String getGaeRuntimeJava(List<Path> services) throws AppEngineException {
+    boolean hasJava8 = false;
+    for (Path serviceDirectory : services) {
+      Path appengineWebXml = serviceDirectory.resolve("WEB-INF/appengine-web.xml");
+      try (InputStream is = Files.newInputStream(appengineWebXml)) {
+        String serviceRuntime = AppEngineDescriptor.parse(is).getRuntime();
+        if (serviceRuntime.startsWith("java11")) {
+          return "java11";
+        } else if (serviceRuntime.startsWith("java8")) {
+          hasJava8 = true;
```
chanseokoh: Looks like this can be `return "java8";`, and we can remove `hasJava8`.

loosebazooka: well if the first service is java8 and the second is java11 then we have an issue where java8 shortcuts the search.

loosebazooka:
@chanseokoh, yeah was discussing offline with ludo the correct way to do this.

PR #748, opened 2018-12-12
better travis comments

loosebazooka:


loosebazooka:
image

@briandealwis I don't think the travis yaml parser is fully up to spec :/, so gonna abandon this.

loosebazooka:
Abandonned.

.travis.yml
```diff
@@ -16,8 +16,9 @@ install:
   - unset _JAVA_OPTIONS
 
   # check if cloud SDK was installed (/bin exists)
-  # travis auto creates cache directories, so remove it before install
   - if [ ! -d $HOME/google-cloud-sdk/bin ]; then
+  # travis auto creates cache directories, so remove it before install
```
chanseokoh: I think there's a possibility this may work if we give a 2-space indent, so that "# travis ..." aligns with the "if" above,

loosebazooka: tried it on the first commit in this PR.

loosebazooka: https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/748/commits/b1751621771851e43230dd3d7ee4ab562f375294

chanseokoh: Ah, now I get it. The YAML parser will have all these in a singe line.

PR #750, opened 2018-12-17
Remove kokoro workaround for cloud sdk

loosebazooka:


loosebazooka:
The workaround is now breaking the build. What a life we live.

PR #759, opened 2019-01-03
Set up Travis build matrix to include OpenJDK 11 build / upgrade plugins

chanseokoh:
Trying out our build tools on Java 11. (Lot of changes based on https://github.com/GoogleContainerTools/jib/pull/1386)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -131,8 +131,8 @@
         <version>3.8.0</version>
         <configuration>
           <!-- Use <release> instead when migrating to JDK 11. -->
-          <source>8</source>
-          <target>8</target>
+          <source>1.8</source>
+          <target>1.8</target>
           <showDeprecation>true</showDeprecation>
           <showWarnings>true</showWarnings>
           <annotationProcessorPaths>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -128,44 +128,30 @@
...
-        <version>3.6.1</version>
+        <version>3.8.0</version>
         <configuration>
-          <compilerId>javac-with-errorprone</compilerId>
-          <forceJavacCompilerUse>true</forceJavacCompilerUse>
-          <!-- maven-compiler-plugin defaults to targeting Java 5, but our javac
-          only supports >=6 -->
-          <source>1.8</source>
-          <target>1.8</target>
+          <!-- Use <release> instead when migrating to JDK 11. -->
```
loosebazooka: does release work with java8 as well? or do you have to use source/target for java8?

chanseokoh: `--release` flag is unknown to Java 8 compilers, so unfortunately, we will have to retain `--source` and `--target` for a very long time.

loosebazooka: Oh I didn't realize it was a pass-through directly to compiler.

PR #764, opened 2019-01-23
Revert "Update getGaeRuntime to handle java11 (#744)"

loosebazooka:
This reverts commit 12794c28717e46748951191bf1336abf348b049c.

Java11 is not available in **compat** mode.


src/main/java/com/google/cloud/tools/appengine/operations/DevServerV1.java
```diff
@@ -272,29 +272,14 @@ boolean isSandboxEnforced(List<Path> services) throws AppEngineException {
...
-        }
-      } catch (IOException | SAXException ex) {
-        throw new AppEngineException(ex);
-      }
-    }
-    if (hasJava8) {
-      return "java8";
-    }
-    return "java7";
+  static String getGaeRuntimeJava(boolean isSandboxEnforced) {
```
chanseokoh: The argument seems incorrect, or confusing at least, because you pass the argument after negating it: `getGaeRuntimeJava(!isSandboxEnforced)`. Indeed, the sandbox is _not_ enforced for Java 8 runtimes. Chould be `noSandbox` or `sandboxNotEnforced`, but I don't like them either honestly.

loosebazooka: Hrmmm... yeah I was reverting to previous state. However, I might as well clean this up.

loosebazooka: @chanseokoh I need to file a followup that removes java7 support anyway. I'd like to leave this PR as a simple revert if that's okay.

chanseokoh: Yeah, sounds good.

PR #767, opened 2019-01-25
allow empty services

### LINKED ISSUE #766: toBuilder in RunConfiguration
Eclipse needs to add jvmFlags after a run config is created:

```
    jvmFlags.add("-Xdebug"); //$NON-NLS-1$
    jvmFlags.add("-Xrunjdwp:transport=dt_socket,server=n,suspend=y,quiet=y,address=" + debugPort); //$NON-NLS-1$
    devServerRunConfiguration.setJvmFlags(jvmFlags);
```

Also we have cases where a RunConfiguration does not have deployables.

elharo:
@loosebazooka @briandealwis fix #766 

It's looking like 0.7.3 doesn't quite work for Eclipse.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/configuration/RunConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/appengine/configuration/RunConfiguration.java
@@ -182,7 +182,8 @@ public class RunConfiguration {
 
   /** @return a mutable list that exposes the internal state */
   public List<String> getJvmFlags() {
-    return jvmFlags;
+    ArrayList<String> copy = new ArrayList<>(jvmFlags);
+    return copy;
   }
 
   @Nullable
@@ -355,7 +356,8 @@ public class RunConfiguration {
 
     public Builder jvmFlags(@Nullable List<String> jvmFlags) {
       if (jvmFlags != null) {
-        this.jvmFlags = jvmFlags;
+        this.jvmFlags.clear();
+        this.jvmFlags.addAll(jvmFlags);
       }
       return this;
     }
--- a/src/main/java/com/google/cloud/tools/appengine/configuration/RunConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/appengine/configuration/RunConfiguration.java
@@ -180,9 +180,7 @@ public class RunConfiguration {
     return pythonStartupArgs;
   }
 
-  /**
-   * @return a mutable list that exposes the internal state
-   */
+  /** @return a mutable list that exposes the internal state */
   public List<String> getJvmFlags() {
     return jvmFlags;
   }
--- a/src/main/java/com/google/cloud/tools/appengine/configuration/RunConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/appengine/configuration/RunConfiguration.java
@@ -18,6 +18,7 @@ package com.google.cloud.tools.appengine.configuration;
 
 import com.google.common.base.Preconditions;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import javax.annotation.Nullable;
@@ -37,7 +38,7 @@ public class RunConfiguration {
   @Nullable private final String threadsafeOverride;
   @Nullable private final String pythonStartupScript;
   @Nullable private final String pythonStartupArgs;
-  @Nullable private final List<String> jvmFlags;
+  private List<String> jvmFlags = new ArrayList<>();
   @Nullable private final String customEntrypoint;
   @Nullable private final String runtime;
   @Nullable private final Boolean allowSkippedFiles;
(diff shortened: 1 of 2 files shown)


briandealwis:
We can't pass around the Builder?

src/main/java/com/google/cloud/tools/appengine/configuration/RunConfiguration.java
```diff
@@ -354,7 +354,9 @@ public Builder pythonStartupArgs(@Nullable String pythonStartupArgs) {
     }
 
     public Builder jvmFlags(@Nullable List<String> jvmFlags) {
```
chanseokoh: `jvmFlags(null)` will be a no-op, and it may feel confusing. I think just taking out `@Nullable` so that people can't call `jvmFlags(null)`? The down side is that we pass the null check down to the clients. Also it may not easy to find places in client code where null can get passed.

But is returning a mutable internal state is the only best way? Can't we have like `addJvmFlags()`?

chanseokoh: Or, maybe passing around the builder like @briandealwis said?

elharo: It's not the only way. It's the way with least code change, here and in Eclipse. Passing a builder really doesn't look right in Eclipse, and it's not a safe change. Note that right now this method does expose mutable internal state. This PR makes that explicit.

Te alternative is adding a toBuilder method that converts a RunConfiguration back into a Builder. It's a lot more code though.

src/main/java/com/google/cloud/tools/appengine/configuration/RunConfiguration.java
```diff
@@ -179,7 +180,7 @@ public String getPythonStartupArgs() {
     return pythonStartupArgs;
   }
 
-  @Nullable
+  /** @return a mutable list that exposes the internal state */
```
loosebazooka: I mean I the solution might be to make a copy of this class into the builder if you want to edit again. I understand that it was never immutable, but it feels weird to do this.

```
RunConfiguration.builder(runConfiguration)
```

wdyt?

elharo: We can do that. It is a lot more work. Protobufs have this automatically. Here we have to write the code to copy everything. We also get hit because the builders don't have getter methods, though that's a little simpler to work around. There are some places we'll have to carry around both a builder and a run configuration.

loosebazooka: Why do the builders need getters? I'm a little confused, can you point me to the eclipse code?

loosebazooka: But I think using a getting on the builder might be the more meaningful solution if that's what you need

elharo: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/3391 is where some of this work is happening. Look particularly at LocalAppEngineServerLaunchConfigurationDelegate and maybe LocalAppEngineServerBehavior. The main reason to have getters on builders ifs if we're passing around builders instead of RunConfigurations between methods, because we also need to query into values. Having all the setters in one class and the getters in another means we need to have two objects when we have code that needs to both read and write.

loosebazooka:
gonna delegate to @briandealwis  + @chanseokoh, since they're more familiar with the eclipse flow.

chanseokoh:
I get that this PR is not about exposing an internal `List`. It has been possible to get the internal object and modify it. This PR just makes `RunConfiguration` start with an empty `List` rather than `null`.

That said, I wonder if it isn't actually safer to set an empty `List` on the Eclipse side, and the Eclipse code will later retrieve that internal `List` and modify it (as it currently does). I admit setting an empty `List` at the beginning on the client side for the purpose of being able to modify it later looks super weird, but for the safety's sake, I fear if there won't be a side-effect of changing the default value from `null` to an empty `List` at appengine-plugins-core may break something. Code-wise, this actually doesn't look bad on the Eclipse side, because you'll get away from the null checks. That is, you'll do
```java
    List<String> vmArguments = Arrays.asList(DebugPlugin.parseArguments(vmArgumentString));
    builder.jvmFlags(vmArguments);  // Always pass an list, which can be empty.
```
instead of
```java
    List<String> vmArguments = Arrays.asList(DebugPlugin.parseArguments(vmArgumentString));
    if (!vmArguments.isEmpty()) {
      builder.jvmFlags(vmArguments);
    }
```
and just do
```java
  // We know it's never null, so just add flags.
  devServerRunConfiguration.getJvmFlags().add("-Xdebug");
```
instead of
```java
    List<String> jvmFlags = new ArrayList<>();
    if (devServerRunConfiguration.getJvmFlags() != null) {
      jvmFlags.addAll(devServerRunConfiguration.getJvmFlags());
    }
    jvmFlags.add("-Xdebug");
```
WDYT?

chanseokoh:
But if a lot of test code have to change without changing the default value to an empty list, it can be a different story...

elharo:
need to close this and reopen to fix some git mess

PR #778, opened 2019-02-22
no more devappserverv2

elharo:
@loosebazooka @ludoch


loosebazooka:
There's a little more to it than this. We need to remove the unused parameters for Dev App Server 2, I'm currently working on a larger PR that includes all the changes fully. But if you want, we can submit this first and I'll do the rest.

elharo:
[ERROR] Errors: 
[ERROR]   DevServersRunnerTest.unnecessary Mockito stubbings � UnnecessaryStubbing 
Unne...

loosebazooka:
Okay I can live with that.

PR #780, opened 2019-03-13
Remove all remaining devserver 2 support

loosebazooka:
Follow up to #778 

Removes remaining dev appserver 2 code.


src/main/java/com/google/cloud/tools/appengine/configuration/RunConfiguration.java
```diff
@@ -57,54 +39,18 @@ private RunConfiguration(
       List<Path> services,
       @Nullable String host,
       @Nullable Integer port,
-      @Nullable String adminHost,
```
etanshaul: all this stuff was devserver v2 related?

chanseokoh: Is there a doc or anything where we can quickly compare and verify which options are for v1 and v2?

loosebazooka: There is in the code, parameters we were intentionally ignoring: https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/780/files#diff-f3a8b789822cdfedaed11b529b5fef47L210

PR #785, opened 2019-04-03
check linkage

elharo:
Interestingly the check fails.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -300,7 +300,7 @@
           <dependency>
             <groupId>com.google.cloud.tools</groupId>
             <artifactId>linkage-checker-enforcer-rules</artifactId>
-            <version>0.1.1</version>
+            <version>0.1.2</version>
           </dependency>
         </dependencies>
         <executions>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -296,9 +296,16 @@
...
+          <dependency>
+            <groupId>com.google.cloud.tools</groupId>
+            <artifactId>linkage-checker-enforcer-rules</artifactId>
+            <version>0.1.1</version>
+          </dependency>
+        </dependencies>
         <executions>
           <execution>
-            <id>enforce</id>
+            <id>convergence</id>
```
loosebazooka: Is this execution necessarily only `convergence`?

elharo: This is just a name. It doesn't really mean anything, but they're can't be two rules named "enforce" or Maven complains.

loosebazooka: Right, but I don't know what the default behavior is. Shouldn't this remain just `enforce`. Why did we change it?

elharo: It should never have been enforce. That's just copy-pasta from the docs.

PR #790, opened 2019-05-08
add phishingprotection API

elharo:
@etanshaul @loosebazooka


src/main/resources/com/google/cloud/tools/libraries/libraries.json
```diff
@@ -749,6 +749,39 @@
       }
     ]
   },
+{
+    "name": "Phishing Protection=",
```
loosebazooka: is this `=` supposed to be in here?

elharo: nope, removed

PR #791, opened 2019-05-08
update linkage checker rule

elharo:
@loosebazooka @suztomo I need this for a demo at the API summit next week.


loosebazooka:
on appengine-plugins-core?

elharo:
Just demoing the output of the tool

PR #792, opened 2019-05-08
Add support for gcloud --flags-file

loosebazooka:
See: https://github.com/GoogleCloudPlatform/app-maven-plugin/issues/340


loosebazooka:
You can use it multiple times:
```
--flags-file file1.yaml --flags-file file2.yaml
```

PR #802, opened 2019-10-14
update dependencies

elharo:
@loosebazooka another security issue was found in commons compress

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
 sudo: false
 language: java
-dist: trusty
 matrix:
   include:
     - name: Oracle JDK 8
       # no openjdk8 yet on travis without special configuration
       jdk: oraclejdk8
+      dist: trusty
     - name: OpenJDK 11
       jdk: openjdk11
 cache:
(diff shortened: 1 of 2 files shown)


loosebazooka:
I think you need to also update the .travis.yaml for java8 builds to be something like: https://github.com/GoogleContainerTools/jib/blob/master/.travis.yml#L7

.travis.yml
```diff
@@ -1,5 +1,6 @@
 sudo: false
 language: java
+dist: trusty
```
loosebazooka: would probably prefer we only use try on jdk 8 instead of wholesale. (move to between line 8-9)

elharo: done

elharo:
Travis passed. For some reason Github didn't pick that up.

loosebazooka:
Weird, they're optional though, and usually so external contributors can see failures. You can probably go ahead and merge this. Or you can trigger a rebuild directly on travis.

loosebazooka:
I restarted one because of a cloud sdk installation failure. Lets see what happens.

PR #804, opened 2019-10-15
Support for java11 standard projects "Class-Path"

loosebazooka:
Automatically copies files referenced in the MANIFEST.MF to the
staging directory

_Context_: After talking to @saturnism and @ludoch this gives users default support for "thin" jars produced by springboot or quarkus with a well defined classpath in the MANIFEST. Similar to what the cloud sdk `gcloud app deploy some-thin.jar` would do if run from a `build` or `target` directory directly.


src/main/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStaging.java
```diff
@@ -31,6 +31,9 @@
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.List;
+import java.util.StringTokenizer;
```
elharo: StringTokenizer is a legacy class that is retained for compatibility reasons although its use is discouraged in new code.

src/main/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStaging.java
```diff
@@ -102,6 +105,7 @@ void stageStandardArchive(AppYamlProjectStageConfiguration config)
     copyExtraFiles(config, copyService);
     copyAppEngineContext(config, copyService);
     copyArtifact(config, copyService);
+    copyArtifactJarClasspath(config, copyService);
```
elharo: what's the effect of this on Java 8?

loosebazooka: It doesn't get called. This code path is only for app.yaml standard apps.

src/main/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStaging.java
loosebazooka: ah gotcha, will switch it to something else

src/test/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStagingTest.java
```diff
@@ -357,4 +358,54 @@ public void testFindRuntime_malformedAppYaml() throws IOException {
...
+
+  private static final Path complexLib = Paths.get("src/test/resources/jars/complexLib.jar");
+  private static final Path simpleLib = Paths.get("src/test/resources/jars/libs/simpleLib.jar");
+  private static final Path complexLibBadManifest =
+      Paths.get("src/test/resources/jars/complexLibBadManifest.jar");
+  private static final Path missingLib = Paths.get("src/test/resources/jars/libs/missing.jar");
+
+  @Test
+  public void testCopyArtifactJarClasspath_noClasspath() throws IOException, AppEngineException {
+    artifact = simpleLib;
```
elharo: This is strange. I don't think you need the artifact field or the static final Path fields here. You can move each path into the single method that uses that one path.

loosebazooka: yeah that was a refactor gone wrong. good catch. the static finals are still used in verification, but I definitely don't need artifact or config to use the pre-initialized values.

loosebazooka: I guess simpleLib is the only one that is repeated everywhere, I could potentially get rid of the static finals

src/main/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStaging.java
```diff
@@ -205,6 +209,37 @@ private static void copyArtifact(AppYamlProjectStageConfiguration config, CopySe
...
+      return;
+    }
+    StringTokenizer classpathEntries = new StringTokenizer(jarClassPath);
+    while (classpathEntries.hasMoreTokens()) {
+      String classpathEntry = classpathEntries.nextToken();
+      // classpath entries are relative to artifact's position and relativeness should be preserved
+      // in the target directory
+      Path jarSrc = artifact.getParent().resolve(classpathEntry);
+      if (!Files.isRegularFile(jarSrc)) {
+        throw new AppEngineException("Could not copy " + jarSrc + " referenced in MANIFEST.MF");
```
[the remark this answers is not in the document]
loosebazooka: makes sense, changed to warning.
also added warning when overwriting a file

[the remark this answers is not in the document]
ludoch: Hum, not sure we want a warning when overwriting.. We would have such warnings every time you restage, without doing a mvn clean....
the appengine-staging is additive and can be non empty before deploying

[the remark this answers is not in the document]
loosebazooka: switched it to "fine", shows up in debug logs

PR #807, opened 2019-10-16
no args constructors for builders

### LINKED ISSUE #765: Setters in builders
Noticed while working on Eclipse. A lot of our builders look like this:

```
    Builder(Path sourceDirectory, Path stagingDirectory) {
      Preconditions.checkNotNull(sourceDirectory);
      Preconditions.checkNotNull(stagingDirectory);

      this.sourceDirectory = sourceDirectory;
      this.stagingDirectory = stagingDirectory;
    }
```

The problem with this pattern is that it is very easy to flip the source and staging directories since they have the same type. Methods such as `setSourceDirectory` and `setStagingDirectory` make it less easy to unintentionally mix up the arguments.

elharo:
This should reduce the likelihood of swapped arguments of the same type (something that bedeviled me in CT4E a few months back though fortunately we caught it before it shipped to customers)

part of #765 
@loosebazooka


src/main/java/com/google/cloud/tools/appengine/configuration/AppEngineWebXmlProjectStageConfiguration.java
```diff
@@ -127,8 +127,20 @@ public String getRuntime() {
...
+   * @deprecated Use
+   *     newBuilder().sourceDirectory(sourceDirectory).stagingDirectory(stagingDirectory) instead.
+   */
+  @Deprecated
   public static Builder builder(Path sourceDirectory, Path stagingDirectory) {
-    return new Builder(sourceDirectory, stagingDirectory);
+    Builder builder = new Builder();
+    builder.sourceDirectory(sourceDirectory);
+    builder.stagingDirectory(stagingDirectory);
+    return builder;
```
chanseokoh: Chain them like `return new Builder().sourceDirectory(...).stagingDirectory(..)`?

src/main/java/com/google/cloud/tools/appengine/configuration/AppYamlProjectStageConfiguration.java
```diff
@@ -89,14 +102,21 @@ public static Builder builder(Path appEngineDirectory, Path artifact, Path stagi
...
     private Path stagingDirectory;
 
-    Builder(Path appEngineDirectory, Path artifact, Path stagingDirectory) {
-      Preconditions.checkNotNull(appEngineDirectory);
-      Preconditions.checkNotNull(artifact);
-      Preconditions.checkNotNull(stagingDirectory);
+    private Builder() {}
 
+    public Builder appEngineDirectory(Path appEngineDirectory) {
       this.appEngineDirectory = appEngineDirectory;
```
chanseokoh: I see that this method no longer does `Preconditions.checkNotNull`, but I this is fine, because `build()` will check these at the end of the day.

src/main/java/com/google/cloud/tools/appengine/configuration/AppEngineWebXmlProjectStageConfiguration.java
```diff
@@ -127,8 +127,20 @@ public String getRuntime() {
     return runtime;
   }
 
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  /**
+   * @deprecated Use
+   *     newBuilder().sourceDirectory(sourceDirectory).stagingDirectory(stagingDirectory) instead.
```
loosebazooka: I'm not convinced this really changes the error situation.

Couldn't someone accidentally do `newBuilder().sourceDirectory(target).stagingDirectory(source)`?

src/test/java/com/google/cloud/tools/appengine/operations/AppEngineWebXmlProjectStagingTest.java
```diff
@@ -75,6 +79,27 @@ public void setUp()
...
+      Assert.fail("allowed missing source directory");
+    } catch (IllegalStateException ex) {
+      Assert.assertEquals("No source directory supplied", ex.getMessage());
+    }
+  }
+  
+  @Test
+  public void testStagingDirectoryRequired() {
+    try {
+      AppEngineWebXmlProjectStageConfiguration.builder().sourceDirectory(destination).build();
```
loosebazooka: infact the test does it right here?

src/main/java/com/google/cloud/tools/appengine/configuration/AppEngineWebXmlProjectStageConfiguration.java
elharo: yes, but they are much less likely to do 

newBuilder().sourceDirectory(stagingDirectory).stagingDirectory(sourceDirectory)?

and if someone does it's easy to notice in code review. This is about making it harder to make a mistake, not making it impossible.

src/test/java/com/google/cloud/tools/appengine/operations/AppEngineWebXmlProjectStagingTest.java
elharo: and you noticed it because of the names :-)

src/main/java/com/google/cloud/tools/appengine/configuration/AppEngineWebXmlProjectStageConfiguration.java
briandealwis: FWIW I agree with @elharo's proposal.  The point of builders was to avoid multi-argument constructors… but we're now using multi-argument builders.  The downside is that missing-values are now a runtime error rather than a compile-time error.

src/main/java/com/google/cloud/tools/appengine/configuration/AppEngineWebXmlProjectStageConfiguration.java
briandealwis: Agreed.  Should be consistent (see `AppYamlProjectStageConfiguration`)

src/main/java/com/google/cloud/tools/appengine/configuration/DeployProjectConfigurationConfiguration.java
```diff
@@ -53,7 +53,7 @@ public String getServer() {
   }
 
   public static Builder builder(Path appEngineDirectory) {
-    return new Builder(appEngineDirectory);
+    return new Builder(Preconditions.checkNotNull(appEngineDirectory));
```
briandealwis: This can be removed.

src/main/java/com/google/cloud/tools/appengine/configuration/AppEngineWebXmlProjectStageConfiguration.java
elharo: done, although this style is my last favorite part of the builder pattern as "fluent" interfaces prioritize writeability over readability and debuggability.

src/main/java/com/google/cloud/tools/appengine/configuration/DeployProjectConfigurationConfiguration.java
elharo: done

loosebazooka:
Can this wait till after release, I'm trying to get something in for the appengine team?

loosebazooka:
Since we're changing this, should we just change all the builders to match this pattern (at least the configuration classes). Either here or in an immediate followup?

elharo:
Smaller PRs are better. No need to do it all in one big batch. This does cover the ones I know about though. If you know of more, please add them to bug #765

src/main/java/com/google/cloud/tools/appengine/configuration/AppEngineWebXmlProjectStageConfiguration.java
loosebazooka: Should we use intermediate interfaces here for required params. I still don't like the idea of going from compile time checks to runtime checks.

I feel like someone (maybe @briandealwis) suggested something like this in the past.

Uncertain about these names, but just to make the point
```
public RequiresSourceDirectory newBuilder()

interface RequiresSourceDirectory {
  RequiresStagingDirectory sourceDirectory()
}

interface RequiresStagingDirectory {
  Builder stagingDirectory()
}

public static final class Builder implements RequiresSourceDirectory, RequiresStagingDirectory {
...
}
```

so any call to would *HAVE* to look like:
```
new Builder().sourceDirectory(sourceDir).stagingDirectory(stagingDir)....build()
```

elharo: I don't think I follow. How does this require that those methods be invoked?

loosebazooka: Because you're only exposing the interface when you return the builder, you return the subset of functionality exposed by the interface.

newBuilder returns RequiresSourcedirectory 
sourceDirectory returns RequiresStagingDirectory
stagingDirectory returns Builder (for all optional things)

briandealwis: Unless this is [auto-generated](https://github.com/google/auto/blob/master/value/userguide/builders.md), it sounds painful.

elharo: @loosebazooka's proposal is clever. I'm still mulling it over in my head. I don't think autovalue will generate code like that though.

loosebazooka: Autovalue/builder actually creates code in the pattern shown in this PR.

loosebazooka: I think for simple cases like this it might work, but still looks ugly. Can punt on the idea for now.

PR #816, opened 2019-11-08
include stderr and stdout in CommandExecutionException

### LINKED ISSUE #814: Log error on command failures (non-zero exits)
Currently, when command fails (such as `ManagedCloudSDK.isUpToDate()`) fails, there is no insight into detailed error message, even though, error message is captured in `CommandExitException`.

https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/master/src/main/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdk.java#L162

Suggest that upon `CommandExitException`, log error with high priority and/or propagate to the wrapping Exception, or, use it as part of the message of `CommandExitException`

elharo:
@loosebazooka @saturnism 
See https://github.com/GoogleCloudPlatform/app-maven-plugin/issues/410


src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandExecutionException.java
```diff
@@ -21,4 +21,12 @@
   public CommandExecutionException(Throwable cause) {
     super(cause);
   }
+
+  /**
+   * @param message failure details; typically output of stderr
```
chanseokoh: output of stdout and stderr

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandCaller.java
```diff
@@ -69,7 +69,19 @@ public String call(
       }
```
chanseokoh: This still doesn't fix #814. `CommandExitException` above should also be modified to include the stdOut and stdErr as an exception message like `CommandExecutionException` does in this PR (if this is the way we want to surface the outputs). `CommandExitException` accepts stdOut and stdErr as an argument in the constructor, but they are not set as an exception message.

elharo: Correct. The information is in the exception, but the client needs to choose to surface that.

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandCaller.java
```diff
@@ -69,7 +69,19 @@ public String call(
...
+        stdOut = stdOutSaver.getResult().get();
+      } catch (InterruptedException | ExecutionException ignored) {
+        stdOut = "stdout collection interrupted";
+      }
+      try {
+        stdErr = stdErrSaver.getResult().get();
+      } catch (InterruptedException | ExecutionException ignored) {
+        stdErr = "stderr collection interrupted";
+      }
+      throw new CommandExecutionException(stdOut + "\n" + stdErr, ex);
```
saturnism: in CommandExitException, error is stored in a separate field instead of the actual `message` field of the exception. I feel that's cleaner. I'd also suggest that stdout and stderr be kept  in 2 different fields and let the caller render it.

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandCaller.java
chanseokoh: And this PR does not fix #814. The error in #814 is about `CommandExitException`, but this PR doesn't change anything about `CommandExitException`. I think we should let #814 open.

chanseokoh: Resolved. I've removed "fixes #814" from your initial PR comment.

loosebazooka:
I think it's fine... will review when I get a chance. @saturnism @chanseokoh any comments?

chanseokoh:
Nothing much from my side besides this minor Javadoc comment: https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/816#discussion_r344307810

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandExecutionException.java
elharo: done

src/main/java/com/google/cloud/tools/managedcloudsdk/command/CommandCaller.java
elharo: done

PR #818, opened 2020-01-27
Add minikube and skaffold to list of gcloud components

etanshaul:
gcloud now manages these extra components. Cloud Code is going to migrate to using the gcloud managed versions of these instead of using its own management system.

### DIFF
--- a/kokoro/continuous.bat
+++ b/kokoro/continuous.bat
@@ -2,7 +2,8 @@
 
 REM Java 9 does not work with our builds right now, force java 8
 set JAVA_HOME=c:\program files\java\jdk1.8.0_152
-set PATH=%JAVA_HOME%\bin;%PATH%
+set CLOUDSDK_PYTHON=c:\python27\python.exe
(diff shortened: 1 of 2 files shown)


etanshaul:
`gcloud components list`:
image

chanseokoh:
There's a Kokoro-windows build failure. Not sure if it's a flaky one.

etanshaul:
triggered a rebuild. Let's see

etanshaul:
> Very interesting to see minikube. Hopefully we are now finally reached the state where installing minikube is as easy as gcloud components install minikube?

Yes that is the goal. There could still be some friction involved with hypervisor setup and so on, but Minikube has been making strides in simplifying this process.

etanshaul:
hmm I think the test failure is because of the python changes to cloud sdk:

```
To use the Google Cloud SDK, you must have Python installed and on your PATH.
As an alternative, you may also set the CLOUDSDK_PYTHON environment variable
to the location of your Python executable.
c:\python27\python.exe: can't open file '!CLOUDSDK_PYTHON_ARGS_NO_S!': [Errno 2] No such file or directory
```

PR #819, opened 2020-01-27
this code is no longer relevant now that the Java 7 runtime is defunct

elharo:
@loosebazooka


chanseokoh:
@etanshaul does IntelliJ still use `AppEngineJreWhitelist` (Java 7 runtime thing) by any chance?

etanshaul:
it is. but since java 7 app engine is shutdown, I think we should remove usages of it. so this change lgtm.

thanks for checking.

PR #830, opened 2020-02-05
Update checkstyle

elharo:
@loosebazooka

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
 
   <groupId>com.google.cloud.tools</groupId>
   <artifactId>appengine-plugins-core</artifactId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -214,12 +214,12 @@
 
       <plugin>
         <artifactId>maven-checkstyle-plugin</artifactId>
-        <version>2.17</version>
+        <version>3.0.1</version>
```
chanseokoh: I think it's 3.1.0, not 3.0.1: https://search.maven.org/search?q=maven-checkstyle-plugin

elharo: done

PR #837, opened 2020-03-23
deps: update to Guava 29.0-jre

elharo:
@loosebazooka 28.2 marked several classes "Do Not Mock" which necessitated a surprisingly large number of changes.


src/main/java/com/google/cloud/tools/managedcloudsdk/command/AsyncByteConsumer.java
```diff
@@ -59,9 +59,10 @@
   @Override
   public void handleStream(final InputStream inputStream) {
     if (executorService.isShutdown()) {
-      throw new IllegalStateException("Cannot re-use " + this.getClass().getName());
+      throw new IllegalStateException("Cannot reuse " + this.getClass().getName());
     }
-    result.setFuture(executorService.submit(() -> consumeBytes(inputStream)));
+    ListenableFuture<String> submit = executorService.submit(() -> consumeBytes(inputStream));
```
loosebazooka: Any reason to expand this out?

elharo: No semantic difference, but it makes refactoring, debugging, and updating easier. The more operations get crammed into a single statement the harder it becomes to read. In this case, it was difficult to demock without introducing local variables.

loosebazooka:
I think you need to rebase to pull in #838 to fix the cloud sdk on windows downgrade issue

PR #823, opened 2020-01-27
Fix test failures: LibrariesTest and ManagedCloudSdkTest

chanseokoh:
Fixes #822.

Hopefully this also fixes https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/818#issuecomment-578885673.

PR #838, opened 2020-04-13
Remove ProcessExecutoryFactory

loosebazooka:
Refactoring out some old code in favor of reducing the number of
classes.


loosebazooka:
looks like we might have an issue on windows downgrading from 288 -> 276. Either that or it's a known issue and we should just update our [`FIXED_VERSION`](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/02abf192d9914dc95693af28b5293a68a53915dc/src/test/java/com/google/cloud/tools/managedcloudsdk/ManagedCloudSdkTest.java#L47).
internal cloud sdk bug: b/153909151

chanseokoh:
I'd try increasing 276. 276 was the highest version available at the time when I was trying to fix [gcloud failures related to Python on Windows](https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/823#issuecomment-578990209). Around that version, gcloud had a lot of changes adding support for both Python 2 and 3 and switching the default.

chanseokoh:
Worked!

PR #839, opened 2020-06-13
Add entrypoint extraction to AppYaml parser

loosebazooka:
part of #840 

also fixes some naming in tests.

NOTE: build is broken because of #841


loosebazooka:
rebased, this should be ready for review

Issue #840, opened 2020-06-13
Add support for binaries in the java11 standard buildpath


loosebazooka:
Currently, `artifact`s are presumed to be jars and processed according (inspected). This feature request is to allow non-jar `artifact`s as long as the user has defined a custom entrypoint in the `app.yaml`.

- [x] app entrypoint parsing (#839)
- [x] support binary buildpath in staging (#842)

loosebazooka:
@ludoch

loosebazooka:
Testing instructions

1. clone repo
2. ./mvnw clean install
3. modify pom config for app-maven-plugin in test app
```
<plugin>
    <groupId>com.google.cloud.tools</groupId>
    <artifactId>appengine-maven-plugin</artifactId>
    <version>2.2.0</version>
    <dependencies>
        <dependency>
            <groupId>com.google.cloud.tools</groupId>
            <artifactId>appengine-plugins-core</artifactId>
            <version>0.8.3-SNAPSHOT</version>
        </dependency>
    </dependencies>
</plugin>
```

ludoch:
Tested with a Quarkus native application. Ship it. Please update the plugin
doc so the feature is easy to discover with Graal-VM native image binaries.

PR #842, opened 2020-06-13
Add binary path to staging

loosebazooka:
part of #840
requires #839

Allows binaries in java11 build path (for example when building a quarkus native app)

Also changes the deprecated use of appYamlStaging builder in tests

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/configuration/AppYamlProjectStageConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/appengine/configuration/AppYamlProjectStageConfiguration.java
@@ -85,7 +85,7 @@ public class AppYamlProjectStageConfiguration {
   /**
    * Create a new builder for AppYamlProjectStageConfiguration.
    *
-   * @deprecated Use newBuilder().appEngineDirectory(appEngineDirectory).artifact(artifact)
+   * @deprecated Use builder().appEngineDirectory(appEngineDirectory).artifact(artifact)
    *     .stagingDirectory(stagingDirectory) instead.
    */
   @Deprecated
--- a/src/main/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStaging.java
+++ b/src/main/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStaging.java
@@ -88,10 +88,9 @@ public class AppYamlProjectStaging {
         }
         // I cannot deploy non-jars without custom entrypoints
         throw new AppEngineException(
-            "Cannot process application with runtime: java11,"
-                + " non-jar artifact: "
-                + config.getArtifact().toString()
-                + " must have a custom entrypoint defined in app.yaml");
+            "Cannot process application with runtime: java11."
+                + " A custom entrypoint must be defined in your app.yaml for non-jar artifact: "
+                + config.getArtifact().toString());
       }
       // I don't know how to deploy this
       throw new AppEngineException(
@@ -277,13 +276,13 @@ public class AppYamlProjectStaging {
   static boolean hasCustomEntrypoint(AppYamlProjectStageConfiguration config)
       throws IOException, AppEngineException {
     // verify that app.yaml that contains entrypoint:
-    Path appEngineDirectory = config.getAppEngineDirectory();
-    if (appEngineDirectory == null) {
+    if (config.getAppEngineDirectory() == null) {
       throw new AppEngineException("Invalid Staging Configuration: missing App Engine directory");
     }
     Path appYamlFile = config.getAppEngineDirectory().resolve(APP_YAML);
-    AppYaml appYaml = AppYaml.parse(Files.newInputStream(appYamlFile));
-    return (appYaml.getEntrypoint() != null);
+    try (InputStream input = Files.newInputStream(appYamlFile)) {
+      return AppYaml.parse(input).getEntrypoint() != null;
+    }
   }
 
   @VisibleForTesting
--- a/src/main/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStaging.java
+++ b/src/main/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStaging.java
@@ -221,7 +221,6 @@ public class AppYamlProjectStaging {
 
   private static void copyArtifact(AppYamlProjectStageConfiguration config, CopyService copyService)
       throws IOException, AppEngineException {
-    // Copy the JAR/WAR file to staging.
     Path artifact = config.getArtifact();
     if (Files.exists(artifact)) {
       Path stagingDirectory = config.getStagingDirectory();
--- a/src/main/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStaging.java
+++ b/src/main/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStaging.java
@@ -74,15 +74,30 @@ public class AppYamlProjectStaging {
       String runtime = findRuntime(config);
       if ("flex".equals(env)) {
         stageFlexibleArchive(config, runtime);
-      } else if ("java11".equals(runtime)) {
-        stageStandardArchive(config);
-      } else {
-        // I don't know how to deploy this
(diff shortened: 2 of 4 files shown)


src/main/java/com/google/cloud/tools/appengine/configuration/AppYamlProjectStageConfiguration.java
```diff
@@ -85,7 +85,7 @@ public static Builder builder() {
   /**
    * Create a new builder for AppYamlProjectStageConfiguration.
    *
-   * @deprecated Use newBuilder().appEngineDirectory(appEngineDirectory).artifact(artifact)
+   * @deprecated Use builder().appEngineDirectory(appEngineDirectory).artifact(artifact)
```
chanseokoh, line 88: I think this doc was right. This method is marked `@Deprecated`.

src/main/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStaging.java
```diff
@@ -74,15 +74,30 @@ public void stageArchive(AppYamlProjectStageConfiguration config) throws AppEngi
...
+        }
+        // I cannot deploy non-jars without custom entrypoints
         throw new AppEngineException(
-            "Cannot process application with runtime: "
-                + runtime
-                + (Strings.isNullOrEmpty(env) ? "" : " and env: " + env));
+            "Cannot process application with runtime: java11,"
+                + " non-jar artifact: "
+                + config.getArtifact().toString()
+                + " must have a custom entrypoint defined in app.yaml");
```
chanseokoh, line 94:
```suggestion
            "Cannot process application with runtime: java11. Non-jar artifact "
                + config.getArtifact().toString()
                + " must have a custom entrypoint defined in app.yaml");
```
And I find that it's desirable to put file path at the end of the message if possible, because file path can often be super long.

src/main/java/com/google/cloud/tools/appengine/configuration/AppYamlProjectStageConfiguration.java
loosebazooka: Hrmm, I think this change is required. There is no "newBuilder", just the non-deprecated "builder" at:  https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/842/files/55cd842928e906a30b3252ecbb026308753248af#diff-2f220793beb46c697d27366fe2a5a214L81

src/main/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStaging.java
```diff
@@ -248,6 +272,20 @@ static void copyArtifactJarClasspath(
...
+  // for non jar artifacts we want to ensure the entrypoint is custom
+  static boolean hasCustomEntrypoint(AppYamlProjectStageConfiguration config)
+      throws IOException, AppEngineException {
+    // verify that app.yaml that contains entrypoint:
+    Path appEngineDirectory = config.getAppEngineDirectory();
+    if (appEngineDirectory == null) {
+      throw new AppEngineException("Invalid Staging Configuration: missing App Engine directory");
+    }
+    Path appYamlFile = config.getAppEngineDirectory().resolve(APP_YAML);
+    AppYaml appYaml = AppYaml.parse(Files.newInputStream(appYamlFile));
```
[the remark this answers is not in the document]
loosebazooka: No, this is a good point. The other parts of the file use try with resources. Good catch.

src/main/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStaging.java
loosebazooka: rearranged this so file name is at end.

src/test/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStagingTest.java
```diff
@@ -118,6 +118,60 @@ public void testStageArchive_standardPath() throws IOException, AppEngineExcepti
...
+    Files.write(
+        appEngineDirectory.resolve("app.yaml"),
+        "runtime: java11\n".getBytes(StandardCharsets.UTF_8),
+        StandardOpenOption.CREATE_NEW);
+
+    AppYamlProjectStaging testStaging = new AppYamlProjectStaging();
+
+    try {
+      testStaging.stageArchive(config);
+      fail();
```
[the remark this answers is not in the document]
loosebazooka: use static import, since there was a lot of that in this file, although I dunno, I think our style has been pretty inconsistent in appengine-plugins-core

src/main/java/com/google/cloud/tools/appengine/configuration/AppYamlProjectStageConfiguration.java
chanseokoh: Ahhh, you're right.

chanseokoh: Ah, you're right.

src/main/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStaging.java
```diff
@@ -108,6 +122,15 @@ void stageStandardArchive(AppYamlProjectStageConfiguration config)
     copyArtifactJarClasspath(config, copyService);
   }
 
+  @VisibleForTesting
+  void stageStandardBinary(AppYamlProjectStageConfiguration config)
+      throws IOException, AppEngineException {
+    CopyService copyService = new CopyService();
+    copyExtraFiles(config, copyService);
+    copyAppEngineContext(config, copyService);
+    copyArtifact(config, copyService);
```
chanseokoh, line 131: Also update the comment on [this line](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/df426fd57c7568659d702992d9eaf50503bdd37e/src/main/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStaging.java#L201) that it's not only JAR or WAR?

src/test/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStagingTest.java
```diff
@@ -381,28 +438,48 @@ public void testFindRuntime_malformedAppYaml() throws IOException {
     }
   }
 
+  @Test
+  public void testHasCustomEntrypoint_true() throws IOException, AppEngineException {
+    Path file = appEngineDirectory.resolve("app.yaml");
+    Files.write(
+        file,
+        "entrypoint: custom custom".getBytes(StandardCharsets.UTF_8),
```
chanseokoh, line 446: Oh! Now I think probably `entrypoint:` should be a string list, not string? That is, `getEntrypoint()` should return `List<String>`?

Whatever it should be, I think it's worth testing things like `entrypoint:` (null) or `entrypoint: []` (`getEntryoint()` returning null and/or an empty list).

loosebazooka: So entrypoint in the app.yaml is a "space separated string", but I think @ludoch can verify this?

loosebazooka:
> Is this PR ready for review?

Yeah it should be now, not sure what's going on with codecov

src/test/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStagingTest.java
chanseokoh: Can you add another test for `entrypoint:`?

loosebazooka: To verify that using a list fails?

chanseokoh: No, I just want to confirm `entrypoint:` returns null instead of an empty string.

PR #843, opened 2020-06-15
Remove icon parameter from libraries

### LINKED ISSUE #841: library icons are returning 404
example: https://cloud.google.com/_static/images/cloud/products/logos/svg/bigquery.svg

loosebazooka:
fixes #841


loosebazooka:
My understanding is that no one is using it anymore.

chanseokoh:
Yeah, if we won't ever have `icon:` values in the JSON, why bother leaving code that will always return null? It also makes it easy for people using appengine-plugins-core to call `.getIcon()` while hopelessly expecting it may return an icon URL if lucky. I'd prefer preventing that from happening. It also makes people already using `.getIcon()` to stop using it and find an alternative.

PR #845, opened 2020-09-09
Add verbosity flag to gcloud commands

loosebazooka:


chanseokoh (approved):

src/main/java/com/google/cloud/tools/appengine/operations/Gcloud.java
```diff
@@ -247,6 +252,15 @@ public Builder setShowStructuredLogs(String showStructuredLogs) {
       return this;
     }
 
+    /**
+     * Sets the verbosity of output for gcloud. @see <a
+     * href="https://cloud.google.com/sdk/gcloud/reference#--verbosity">gcloud docs</a>
+     */
+    public Builder setVerbosity(String verbosity) {
+      this.verbosity = verbosity;
```
elharo, line 260: This needs an enum or a check on the allowed values.

loosebazooka: I thought about that, but gcloud will just spit out the error.

elharo: Failing earlier is better. Not failing at all because an enum prevents passing an incorrect value is better still.

PR #861, opened 2021-01-22
Update Javadoc as part of respectful code cleanup

mpeddada1:


chanseokoh (approved):
nice

elharo (changes requested):
This method doesn't matter much these days. The comment can be rewritten not to mention the list explicitly if you like, but as proposed this is not correct. There's no such thing as a "Java safe list".

elharo (approved):

PR #862, opened 2021-01-28
Remove Travis CI and migrate to Github Actions

chanseokoh:
About `actions/cache`, I consulted [their example](https://github.com/actions/cache/blob/main/examples.md#java---maven), which is the same config used in [spring-cloud-gcp](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/5eeedd5a6893909468bb7f6aceac9c2f9c6c8a23/.github/workflows/integrationTests.yaml#L42-L47).


elharo (approved):

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,45 @@
+name: Unit Tests
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  workflow_dispatch:
```
loosebazooka, line 7: so this `workflow dispatch` allows us to manually run it?

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+      - uses: google-github-actions/setup-gcloud@v0.2.0
+      - uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
```
loosebazooka, line 28: what does this block do?

loosebazooka (approved):
lgtm, I just am not sure I understand it all

.github/workflows/unit-tests.yaml
chanseokoh: Yes, and we can also manually run it on any branch.

.github/workflows/unit-tests.yaml
chanseokoh: `hashFiles` computes an aggregated SHA of all the given files, and the final hash is used as a key to save and retrieve the cache (~/.m2). That is, as long as those glob files remain untouched, this action will retrieve the same cache.

PR #877, opened 2021-09-02
Remove prepare_release script and migrate to GitHub Actions

chanseokoh:
This allows a release eng to prepare a release with a click. No need to run a script on a local machine that interferes with your local git repo.

This also automates filing a PR. So what comes with free is automated notifications: just click a button, and whether it succeeds or fails, you'll be notified by email.

image


.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,55 @@
...
+          fi
+          if [[ $( git ls-remote --tags origin refs/tags/v${{ github.event.inputs.release_version }} ) ]]; then
+            echo 'Tag "v${{ github.event.inputs.release_version }}" already exists. Double-check the version to release!'
+            exit 1
+          fi
+
+      - name: Create tag for release and prepare branch for next development version
+        run: |
+          git checkout -b release-v${{ github.event.inputs.release_version }}
+          git config user.email ${{ github.actor }}@users.noreply.github.com
```
suztomo, line 32: This email would make a bad commit history. If this is the only way, then it's okay.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,55 @@
...
+          git commit -am "next development version"
+
+          git push origin v${{ github.event.inputs.release_version }}
+          git push origin release-v${{ github.event.inputs.release_version }}
+
+      - name: Create pull request
+        uses: repo-sync/pull-request@v2.6
+        id: create-pr
+        with:
+          # Use a personal token to file a PR as a non-bot author to trigger other workflows (e.g., unit tests):
```
suztomo, line 49: This comment does not seem applicable here.

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,55 @@
...
+          # https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token;
+          github_token: ${{ secrets.GA_RELEASE_PR_PERSONAL_TOKEN }}
```
suztomo, line 51: Would you add source code comment on how this is set?

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Check out code
+        uses: actions/checkout@v2.3.4
+
+      - name: Check input
+        run: |
+          if [[ ! "${{ github.event.inputs.release_version }}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo 'version "${{ github.event.inputs.release_version }}" not in ###.###.### format'
+            exit 1
+          fi
+          if [[ $( git ls-remote --tags origin refs/tags/v${{ github.event.inputs.release_version }} ) ]]; then
```
suztomo, line 24: What about an existing branch with the same name?

.github/workflows/prepare-release.yml
chanseokoh: I believe this is customary. For example, dependabots:
```
commit 8c0de2dc7dfc3749d6ea920725c69ca9b20183fa
Author: dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com>
Date:   Mon Aug 2 14:59:06 2021 -0400

    Bump commons-compress from 1.20 to 1.21 (#875)
```

.github/workflows/prepare-release.yml
chanseokoh: It's possible and perfectly OK to have the same name for a branch and a tag, although anyone should really avoid it. The most critical thing is that we create a tag on the correct commit, which is what this workflow does, because we will build an artifact from a commit pointed to by the tag. So I'd say it is important to make sure that there is no existing tag with the same name that was likely created by a previous run of this workflow, e.g., due a build failure, or by that the person who triggered this workflow entered a wrong version. But for a branch name, it doesn't really matter. I think it isn't really worth checking branch names.

.github/workflows/prepare-release.yml
chanseokoh: The comment applies to the following line `github_token: ...`. Note that for other steps, it is a bot author, but for this one, we use a personal token (`secrets.GA_RELEASE_PR_PERSONAL_TOKEN`).

.github/workflows/prepare-release.yml
chanseokoh: The link in the comment above explains it with details.

.github/workflows/prepare-release.yml
suztomo: Good. Okay.

.github/workflows/prepare-release.yml
suztomo: That makes sense. I think I was confused "as a non-bot author" part.

```suggestion
          # Use a personal token to file a PR to trigger other workflows (e.g., unit tests):
```

.github/workflows/prepare-release.yml
suztomo: I didn't find GA_RELEASE_PR_PERSONAL_TOKEN in https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token; Where did I miss?

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,55 @@
...
+
+          git push origin v${{ github.event.inputs.release_version }}
+          git push origin release-v${{ github.event.inputs.release_version }}
+
+      - name: Create pull request
+        uses: repo-sync/pull-request@v2.6
+        id: create-pr
+        with:
+          # Use a personal token to file a PR as a non-bot author to trigger other workflows (e.g., unit tests):
+          # https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token;
```
suztomo, line 50:
```suggestion
          # https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token
```

.github/workflows/prepare-release.yml
suztomo: I think it's a matter of error message. If the branch with the same name exists, line 31 fails with an appropriate error message.

```
git checkout -b release-v${{ github.event.inputs.release_version }}
```

.github/workflows/prepare-release.yml
chanseokoh: The doc has the link to "[Creating and storing encrypted secrets](https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets)", where they explain how one can create and store a secret, and access it in GitHub Actions. `GA_RELEASE_PR_PERSONAL_TOKEN` is just a variable name; you can name it however you want.

.github/workflows/prepare-release.yml
chanseokoh: To me, "as a non-bot author" is correct (the PR will seen as filed by a user like [this](https://github.com/chanseokoh/app-gradle-plugin/pull/3)), but I have no problem just removing it.

chanseokoh:
@suztomo I accepted your code change suggestions, which made the bot require you write `@googlebot I fixed it.`.

.github/workflows/prepare-release.yml
suztomo: Thank you for explanation. Then a source code comment "Save your access token as GA_RELEASE_PR_PERSONAL_TOKEN following http://..." is I was looking for.

suztomo (approved):

chanseokoh:
@suztomo can you resolve the CLA bot complaint?

suztomo:
I didn't know it was complaining to me. Updated the labels. I remember that CLA bot doesn't work with accepting GitHub suggestions.

PR #879, opened 2022-01-25
Adding the gke-gcloud-auth-plugin component

peterlin741:
> The `gke-gcloud-auth-plugin` will be rolling out to all customers in Q2 2022 (with MSAs etc), so we will need to install this component for GKE cluster access. It is safe for us to start installing this for users now: no need to wait

context from @briandealwis 

Announcement: https://groups.google.com/a/google.com/g/containers-discuss/c/TorLPo6FOwc/m/4-fPMFP2AgAJ


etanshaul (approved):

chanseokoh (approved):

chanseokoh:
Do you want it now?

peterlin741:
> Do you want it now?

yeah, as discussed, this is needed for our next plugin release at the start of the next month, but it would be great to have this as soon as possible, so that our team can test our changes locally

PR #881, opened 2022-03-16
Support Java 17 runtime

colorizenl:
According to your [own documentation](https://cloud.google.com/appengine/docs/standard/java-gen2/runtime#java-11), Java 17 is now supported in the App Engine Standard Environment. 

However, the Gradle plugin did not accept the `runtime: java17` value in `app.yaml`.

### DIFF
--- a/src/main/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStaging.java
+++ b/src/main/java/com/google/cloud/tools/appengine/operations/AppYamlProjectStaging.java
@@ -76,7 +76,7 @@ public class AppYamlProjectStaging {
         stageFlexibleArchive(config, runtime);
         return;
       }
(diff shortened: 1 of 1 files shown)


clementdenis:
I have a slightly more future-proof version of the same:
https://github.com/GoogleCloudPlatform/appengine-plugins-core/compare/master...clementdenis:master

I also need this for the Maven plugin.

ludoch:
Thanks, both versions would work.
LGTM.

elefeint (changes requested):
Thank you for the contribution!
Could you fix the corresponding test `AppYamlProjectStagingTest.testStageArchive_java11BinaryWithoutEntrypoint()`?

colorizenl:
Sorry for the late reply, I updated the test.

PR #882, opened 2022-03-24
Update gcloud to a recent version

elefeint:
Can't be the latest version because `ManagedCloudSdkTest.testManagedCloudSdk_latest()` tries to "downgrade" to the fixed version, and then upgrades to latest, so they can't be the same.


mpeddada1 (approved):
Sorry for the delay! LGTM

PR #884, opened 2022-03-29
Compile-time Java 17 compatibility

elefeint:
Upgrades plugins and adds errorprone JPMS workarounds

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -97,13 +97,13 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <version>4.13.1</version>
+      <version>4.13.2</version>
       <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
-      <version>3.1.0</version>
+      <version>4.4.0</version>
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -334,6 +305,56 @@
...
+                  <version>0.7.8</version>
+                </path>
+              </annotationProcessorPaths>
+              <fork>true</fork>
+              <compilerArgs>
+                <arg>-XDcompilePolicy=simple</arg>
+                <arg>-Xplugin:ErrorProne -Xep:NullAway:ERROR -XepOpt:NullAway:AnnotatedPackages=com.google.cloud.tools -XepOpt:NullAway:KnownInitializers=com.google.gson.Gson.fromJson -XepOpt:NullAway:ExcludedFieldAnnotations=org.mockito.Mock,org.junit.runners.Parameterized -XepOpt:NullAway:ExcludedClasses=com.google.cloud.tools.appengine.operations.DevServerTest,com.google.cloud.tools.appengine.operations.cloudsdk.serialization.AppEngineDeployResult,com.google.cloud.tools.io.FilePermissionsTest,com.google.cloud.tools.managedcloudsdk.install.InstallerFactoryTest,com.google.cloud.tools.appengine.AppEngineDescriptorTest,com.google.cloud.tools.appengine.operations.cloudsdk.serialization.CloudSdkVersionTest,com.google.cloud.tools.appengine.operations.GenRepoInfoFileTest,com.google.cloud.tools.appengine.operations.DeploymentTest,com.google.cloud.tools.appengine.operations.AuthTest,com.google.cloud.tools.appengine.configuration.RunConfigurationTest</arg>
```
mpeddada1, line 340: could we have these arguments in separate `<arg></arg> ` blocks?

mpeddada1, line 340: Perhaps we could put the list of values for `XepOpt:NullAway:ExcludedClasses` in a variable?

mpeddada1, line 340: Would it be possible to document what these arguments do?

pom.xml
```diff
@@ -334,6 +305,56 @@
   </build>
 
   <profiles>
+    <profile>
+      <id>java9_and_up</id>
```
mpeddada1, line 309: What's the best way to try this out? And out of curiosity, will the usage be different for Java 8&below vs Java 9 &up?

elefeint: The reason I set Java 9 as the cutoff is that `--add-opens` option was added in Java 9. So if you pass `-add-opens` to `java` when running Java 8, you'll get an error saying that this option is unrecognized.

You can use SDKman to switch java distributions -- I've been testing with `java 8.0.322-tem` and `17.0.1-oracle`. 

For this codebase, use `mvn clean verify` to reproduce the issue -- comment out the customizations to see the problem.

For the example in https://github.com/elefeint/examples/blob/master/appengine17-helloworld/ , you can use `mvn clean package appengine:run` -- it's a lot less indirection than launching a test with failsafe that then runs devserver.

pom.xml
elefeint: Probably. These were there from before; they are errorprone exclusions.
I am would actually prefer to take out all the errorprone/findbugs plugins and replace them with Sonar github check. But out of scope for the current task.

elefeint: For the add-exports, I'll add a comment pointing at Errorprone documentation -- https://errorprone.info/docs/installation#maven

This is purely for errorprone; nothing to do with appengine plugins.

elefeint: For the errorprone optouts, I guess we can link to https://errorprone.info/docs/flags.

mpeddada1: That sounds like a good plan. We need some code coverage info in this repo as well so installing sonar will be helpful.

pom.xml
```diff
@@ -334,6 +307,59 @@
...
+                </path>
+              </annotationProcessorPaths>
+              <fork>true</fork>
+              <compilerArgs>
+                <arg>-XDcompilePolicy=simple</arg>
+                <arg>-Xplugin:ErrorProne -Xep:NullAway:ERROR -XepOpt:NullAway:AnnotatedPackages=com.google.cloud.tools -XepOpt:NullAway:KnownInitializers=com.google.gson.Gson.fromJson -XepOpt:NullAway:ExcludedFieldAnnotations=org.mockito.Mock,org.junit.runners.Parameterized -XepOpt:NullAway:ExcludedClasses=${errorprone_exclusions}</arg>
+                <!-- Errorprone requires opening up some JDK modules.
+                     See https://errorprone.info/docs/installation#maven for details
+                -->
+                <arg>-J--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED</arg>
```
mpeddada1, line 346: Ah looks like we run into the following issues if we don't include these options:
```
 com.sun.jna.Library is not found, referenced from net.bytebuddy.dynamic.loading.ClassInjector$UsingJna$Dispatcher$Jvm
  com.sun.jna.LastErrorException is not found, referenced from net.bytebuddy.dynamic.loading.ClassInjector$UsingJna$Dispatcher$Jvm
  com.sun.jna.Platform is not found, referenced from net.bytebuddy.dynamic.loading.ClassInjector$UsingJna$Dispatcher$CreationAction
  com.sun.jna.Native is not found, referenced from net.bytebuddy.dynamic.loading.ClassInjector$UsingJna$Dispatcher$CreationAction
  com.sun.jna.JNIEnv is not found, referenced from net.bytebuddy.dynamic.loading.ClassInjector$UsingJna$Dispatcher$Enabled
  com.sun.jna.FunctionMapper is not found, referenced from net.bytebuddy.dynamic.loading.ClassInjector$UsingJna$Dispatcher$Windows32BitFunctionMapper
  com.sun.jna.Platform.isWindows is not found, referenced from net.bytebuddy.dynamic.loading.ClassInjector$UsingJna$Dispatcher$CreationAction

```

mpeddada1 (approved):
This looks great! Will the next step be to include java 17 in the CI matrix?

PR #886, opened 2022-04-06
Compile to java 8

elefeint:
I introduced an unintented target Java version upgrade when updating to compile-time Java 17 compatibility.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -321,7 +321,8 @@
             <version>3.10.1</version>
             <configuration>
               <!-- For Java 8, the plugin is overridden below, without errorprone -->
-              <release>11</release>
+              <source>1.8</source>
+              <target>1.8</target>
               <showDeprecation>true</showDeprecation>
               <showWarnings>true</showWarnings>
               <annotationProcessorPaths>


meltsufin (approved):

chanseokoh:
```xml
      <id>java9_and_up</id>
      <activation>
        <jdk>[9,]</jdk>
      </activation>
            ...
            <configuration>
              <!-- For Java 8, the plugin is overridden below, without errorprone -->
              <source>1.8</source>
              <target>1.8</target>
```

I think this setup can be confusing. First of all, it currently generates the following warning:
```
Warning:  [options] bootstrap class path not set in conjunction with -source 8
```
See [this](https://stackoverflow.com/questions/51571537/bootstrap-class-path-not-set-in-conjunction-with-source-1-7-netbeans) and [this](https://stackoverflow.com/a/12050087) SO answers.

What the repo was doing with the Java 11 GitHub Action job in the past was to (1) use JDK 11 to compile the project source to see if it compiles well; and (2) use Java 11 to run tests to see if it runs well on Java 11. It was sort of an optional test for future Java 11 migration.

Specifying `--source 1.8` tends to convey the intention that "I want the compiler to consider only Java 8 language rules when compiling source", which is almost always unnecessary when using higher Java versions because practically it never happens that language rules become incompatible in newer versions. So usually the intention is to specify both `--source 1.8` and `--target 1.8`, hoping that you want to compile and generate Java 8-compatible binaries using, say, Java 11. However, as the warning says, it is not generating Java 8-compatible binaries; you need to specify the bootstrap class path of JDK 8. But at that point, you'd rather just use JDK 8 to compile. Right now, the setup is basically generating a Java 11 binary and testing it on Java 11.

elefeint:
That makes sense. I wanted to make sure target is 1.8, so we don't accidentally end up generating Java 11 bytecode if/when kokoro configuration changes to run on Java 11.

So it sounds like as long as we aim for Java 8 minimum runtime compatibility, we *should not* upgrade kokoro's version of Java.

elefeint:
(and then _java9_and_up_ profile will remain purely for testing)

PR #898, opened 2022-09-12
Update snakeyaml to 1.32

zhumin8:
upgrade to patched version for [vulnerability](https://github.com/advisories/GHSA-3mc7-4q67-w48m).

PR #900, opened 2022-09-14
Update source links in libraries.json

zhumin8:
This is a temporary fix to failed test in #898.


elefeint (changes requested):
Would it be easier long term to point at the monorepo?

zhumin8:
I am not sure if monorepo is ready at this point. For example, I don't see `java-bigquery` and `java-datastore` there.

elefeint (approved):

Issue #902, opened 2022-10-11
Fixing Flaky Test


bmk15897:
### Description
Flaky Test found using [NonDex](https://github.com/TestingResearchIllinois/NonDex) by running the command -
`mvn edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=com.google.cloud.tools.managedcloudsdk.install.InstallerTest#testCall_withOverrideComponents`

The logged failure - 
```[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   InstallerTest.testCall_withOverrideComponents:115 
Argument(s) are different! Wanted:
mockCommandRunner.run(
    [scriptexec, test-install.script, --path-update=false, --command-completion=false, --quiet, --usage-reporting=true, --override-components, myothercomponent, mycomponent],
    C:\Users\Bharati\AppData\Local\Temp\junit6297198120928949252,
    {"PROPERTY" = "value"},
    mockConsoleListener
);
-> at com.google.cloud.tools.managedcloudsdk.command.CommandRunner.run(CommandRunner.java:49)
Actual invocations have different arguments:
mockCommandRunner.run(
    [scriptexec, test-install.script, --path-update=false, --command-completion=false, --quiet, --usage-reporting=true, --override-components, mycomponent, myothercomponent],
    C:\Users\Bharati\AppData\Local\Temp\junit6297198120928949252,
    {"PROPERTY" = "value"},
    mockConsoleListener
);
-> at com.google.cloud.tools.managedcloudsdk.install.Installer.install(Installer.java:99)
```
### Investigation
The test fails at `InstallerTest.testCall_withOverrideComponents:115` with the error regarding a mismatch of arguments while using the `overrides` variable declared on line 102. The variable is declared using the HashSet constructor. According to the [documentation](https://docs.oracle.com/javase/8/docs/api/java/util/HashSet.html), the HashSet class makes no guarantees as to the iteration order of the set; in particular, it does not guarantee that the order will remain constant over time. This makes the test outcome non-deterministic. To fix this, HashSet needs to be replaced with a data structure that can guarantee the order of its elements. 
###  Fix
[...]

PR #903, opened 2022-10-12
Fix Flaky Test caused by HashSet Ordering permutations 

bmk15897:
Related to the Issue [#902](https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/902)

### Description
Flaky Test found using [NonDex](https://github.com/TestingResearchIllinois/NonDex) by running the command -
`mvn edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=com.google.cloud.tools.managedcloudsdk.install.InstallerTest#testCall_withOverrideComponents`

The logged failure for the test com.google.cloud.tools.managedcloudsdk.install.InstallerTest.testCall_withOverrideComponents -

```
[ERROR] Failures: 
[ERROR]   InstallerTest.testCall_withOverrideComponents:115 
Argument(s) are different! Wanted:
mockCommandRunner.run(
    [scriptexec, test-install.script, --path-update=false, --command-completion=false, --quiet, --usage-reporting=true, --override-components, myothercomponent, mycomponent],
    C:\Users\Bharati\AppData\Local\Temp\junit6297198120928949252,
    {"PROPERTY" = "value"},
    mockConsoleListener
);
-> at com.google.cloud.tools.managedcloudsdk.command.CommandRunner.run(CommandRunner.java:49)
Actual invocations have different arguments:
mockCommandRunner.run(
    [scriptexec, test-install.script, --path-update=false, --command-completion=false, --quiet, --usage-reporting=true, --override-components, mycomponent, myothercomponent],
    C:\Users\Bharati\AppData\Local\Temp\junit6297198120928949252,
    {"PROPERTY" = "value"},
    mockConsoleListener
);
-> at com.google.cloud.tools.managedcloudsdk.install.Installer.install(Installer.java:99)
```
### Investigation
The test fails at InstallerTest.testCall_withOverrideComponents:115 with the error regarding a mismatch of arguments while using the overrides variable declared on line 102. The variable is declared using the HashSet constructor. According to the [documentation](https://docs.oracle.com/javase/8/docs/api/java/util/HashSet.html), the HashSet class makes no guarantees as to the iteration order of the set; in particular, it does not guarantee that the order will remain constant over time. This makes the test outcome non-deterministic. To fix this, HashSet needs to be replace with a data structure that can guarantee the order of its elements.

### Fix
[LinkedHashSet](https://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashSet.html) preserves the iteration ordering, which is the order in which elements were inserted into the set. As this data structure is deterministic, replacing the HashSet with the LinkedHashSet removes the non-determinism without needing any further changes and ensures that the flakiness from the test is removed.
The PR does not introduce a breaking change.

### DIFF
--- a/src/test/java/com/google/cloud/tools/managedcloudsdk/install/InstallerTest.java
+++ b/src/test/java/com/google/cloud/tools/managedcloudsdk/install/InstallerTest.java
@@ -24,7 +24,7 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -99,7 +99,7 @@ public class InstallerTest {
 
   @Test
   public void testCall_withOverrideComponents() throws Exception {
-    Set<String> overrides = new HashSet<>(Arrays.asList("mycomponent", "myothercomponent"));
+    Set<String> overrides = new LinkedHashSet<>(Arrays.asList("mycomponent", "myothercomponent"));
 
     new Installer(
             fakeSdkRoot,


elefeint (approved):
Thank you!

PR #919, opened 2023-04-27
[do not merge] check adding -E flag in build script

emmileaf:
For troubleshooting release script / Kokoro environment compatibility

### DIFF
--- a/kokoro/continuous.sh
+++ b/kokoro/continuous.sh
@@ -5,8 +5,8 @@ set -e
 # Display commands to stderr.
 set -x
 
-sudo /opt/google-cloud-sdk/bin/gcloud components update
-sudo /opt/google-cloud-sdk/bin/gcloud components install app-engine-java
+sudo -E /opt/google-cloud-sdk/bin/gcloud components update
+sudo -E /opt/google-cloud-sdk/bin/gcloud components install app-engine-java
 
 cd github/appengine-plugins-core
 


emmileaf:
FYI @lqiu96 - let's see if the `kokoro-ubuntu` check gets stuck on the same issue as the release script?

lqiu96:
> FYI @lqiu96 - let's see if the `kokoro-ubuntu` check gets stuck on the same issue as the release script?

Thanks @emmileaf. I think `kokoro-ubuntu` runs the this job: https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/master/kokoro/continuous.sh, right? The PR I just submitted passed that CI job so I think it's probably the additional `-E` flags in the release script causing the issue.

lqiu96:
Not entirely sure why it was added in the first place or why it was only added to only the release job. I'll do a bit of digging before I just remove this flag.

lqiu96:
Ok, nvm. I misunderstood the PR. This still passes even with the `-E` flag added into the continuous script. That's odd...

emmileaf:
> This still passes even with the `-E` flag added into the continuous script. That's odd...

Yeah this is just a sanity check to see if the issue can be reproduced on the continuous script, but seems not...could it have been a one-time/transient thing?

lqiu96:
> > This still passes even with the `-E` flag added into the continuous script. That's odd...
> 
> Yeah this is just a sanity check to see if the issue can be reproduced on the continuous script, but seems not...could it have been a one-time/transient thing?  

Tried re-running the job and it's the same issue. Stuck waiting for user input unfortunately.

PR #978, opened 2024-01-16
chore: change release order

JoeWang1127:
In this PR:
- Change the release order to release app-gradle-plugin first, so that we can rerun the kokoro job if the release job failed. Otherwise, app-maven-plugin will release first and we can't rerun the job.

### DIFF
--- a/.kokoro/stage.sh
+++ b/.kokoro/stage.sh
@@ -23,10 +23,12 @@ source "${dir}"/common.sh
 
 pushd "${dir}"/../
 
-MAVEN_SETTINGS_FILE=$(realpath .)/settings.xml
-
+# Set up environment variables for release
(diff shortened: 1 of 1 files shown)


suztomo (approved):
This is fine for experiment. But after confirming Gradle release, we'll need to revert the order. We need to release appengine-plugins-core first because the Gradle plugin needs appengine-plugins-core. And, because of this dependency, I think the build fails.

JoeWang1127:
> This is fine for experiment. But after confirming Gradle release, we'll need to revert the order. We need to release appengine-plugins-core first because the Gradle plugin needs appengine-plugins-core. And, because of this dependency, I think the build fails.

Yeah, this makes sense. I'll hold this PR now and wait for the release result.

PR #1020, opened 2024-04-13
chore: update commons-compress to 1.26.1

JoeWang1127:


app-maven-plugin/pom.xml
```diff
@@ -112,12 +112,6 @@
       <scope>provided</scope>
     </dependency>
 
-    <dependency>
-      <groupId>commons-io</groupId>
-      <artifactId>commons-io</artifactId>
```
JoeWang1127, line 117: `commons-io` is a dependency of `commons-compress` so that we don't need to define this dependency anymore.

suztomo: This module uses FileUtils from commons-io. Please keep this declaration.

image

However, it's only used in `FileUtils.deleteDirectory(stagingDirectory.toFile())`. Can you see whether it can be replaced by Guava's method? https://guava.dev/releases/32.0.0-jre/api/docs/com/google/common/io/MoreFiles.html#deleteRecursively(java.nio.file.Path,com.google.common.io.RecursiveDeleteOption...) If it works, then this module does not need to declare the dependency.

suztomo (changes requested):
Please keep commons-io dependency or remove the reference to its class.

app-maven-plugin/pom.xml
JoeWang1127: I tested guava's `deleteRecursively` but the following error occurred:

```
Caused by: com.google.common.io.InsecureRecursiveDeleteException: /var/folders/2b/zvjk1hd11jx5kmjj2s8my6tr00zvqy/T/junit17596486141951131749: unable to guarantee security of recursive delete
        at com.google.common.io.MoreFiles.checkAllowsInsecure(MoreFiles.java:750)
        at com.google.common.io.MoreFiles.deleteRecursively(MoreFiles.java:542)
        at com.google.cloud.tools.maven.stage.AppEngineWebXmlStager.stage(AppEngineWebXmlStager.java:61)
```

The test passed by using `RecursiveDeleteOption.ALLOW_INSECURE`, but I'm not sure whether this is a good option.

I'll update the dependency version instead.

suztomo (approved):

